Issue #40, opened 2020-10-09
Frontend design work


teocomi:
- [x] update user profile
- [x] update server info (if role permits)
- [ ] add subscriptions
- [ ] manage stream permissions
  - [ ] change link sharing
  - [x] add/remove collaborators
- [ ] create /revoke tokens
- [x] display more detailed commit info
- [x] delete streams & branches
- [x] pagination on streams & commits
- [x] tooltips
- [ ] (optional) create commits 
- [ ] (optional) proper home / feed page 
- [ ] (REQUIRED) Easter egg of sorts

didimitrie:
Since this was marked as done in the alpha release, closing for now.

PR #48, opened 2020-10-17
Frontend 👓

teocomi:
Alpha frontend, addresses most of #40


didimitrie:
The frontend fails to build on a fresh clone: 
```
 ERROR  Failed to compile with 1 errors                 12:00:53 PM

This dependency was not found:

* v-tooltip in ./src/main-frontend.js

To install it, you can run: npm install --save v-tooltip
```

Also, as a quick question, why was a separate [tooltip library](https://github.com/Akryum/v-tooltip) needed vs the [vuetifyjs](https://vuetifyjs.com/en/components/tooltips/) one? I might be missing something!

frontend/src/AppFrontend.vue
```diff
@@ -1,33 +1,360 @@
...
-    </v-content>
+    </v-main>
   </v-app>
 </template>
 <script>
-import HelloWorld from './components/HelloWorld';
+import userQuery from "./graphql/user.gql"
+import gql from "graphql-tag"
 
 export default {
```
didimitrie, line 134: Search doesn't seem to be debounced? Not a blocker but to remember this is a self-inflicted DDoS route that should be patched.

teocomi: Not yet implemented, would you like me to add it to this PR or make an issue for it?

teocomi:
> Quick question, why was a separate [tooltip library](https://github.com/Akryum/v-tooltip) needed vs the [vuetifyjs](https://vuetifyjs.com/en/components/tooltips/) one? I might be missing something!

The default tooltip requires smth like 5-6 lines of code for each tooltip this lib adds a "v-tooltip" prop to any component and voila'!

didimitrie:
> The default tooltip requires smth like 5-6 lines of code for each tooltip this lib adds a "v-tooltip" prop to any component and voila'!

feel you dude, checked the docs for the other one and i can understand the pain... otherwise, we're good to merge pending issue on debouncing (i don't want this to slip into something like a beta!)

PR #76, opened 2021-01-14
Specle Server 2.0 dev deployment

gjedlicska:
Setting the vars in .env, running `docker-compose up` results in a fully functional server, with some niceties like postgres admin.

Remaining issues:
* .env file location and docker-compose.yml is not compatible. Should the compose file move to the server package folder?
* ~~where should the initdb folder reside?~~ No need for it, done via envvars
* ~~postgres admin is missing~~
* ~~documentation is not updated~~


docker-compose.yaml
```diff
@@ -0,0 +1,64 @@
...
+  # this could be enabled once the server container is added
+  server:
+    build:
+      .
+    depends_on:
+      - database
+      - redis
+    ports:
+      - "3000:3000"
+    environment:
```
peterjgrainger, line 13: The environment variables I used were 
```
        DEBUG: "speckle:*"
        POSTGRES_URL: postgres://speckle:password@database/speckle_database_name
        REDIS_URL: redis://redis:6379
        SESSION_SECRET: 'keyboard cat'
        STRATEGY_LOCAL: 'true'
        CANONICAL_URL: 'http://localhost:3000'
```

packages/server/knexfile.js
```diff
@@ -26,6 +26,13 @@ function walk( dir ) {
 
 let migrationDirs = walk( './modules' )
 
+let env = process.env
```
peterjgrainger, line 29: As above, no need to add this code you can use the `POSTGRES_URL`

.env-docker-compose
```diff
@@ -0,0 +1,70 @@
+#########################################################vvv
```
peterjgrainger, line 1: Instead of having a file you could put the environment variables in the `docker-compose` file. This might get confused with the `.env` file which is completely different

docker-compose.yaml
```diff
@@ -0,0 +1,64 @@
...
+    environment: 
+        PGADMIN_DEFAULT_EMAIL: "${PGADMIN_EMAIL}"
+        PGADMIN_DEFAULT_PASSWORD: "${PGADMIN_PASSWORD}"
+    ports: 
+        - "16543:80"
+    depends_on: 
+        - database
+
+  redis:
+    image: "redis"
```
peterjgrainger, line 42: same as postgres, could be the alpine image

peterjgrainger (commented):
Just a few notes. I know it's an initial go at it

.env-docker-compose
gjedlicska: Yes, I'm not satisfied with this yet.

docker-compose.yaml
gjedlicska: Yup.

docker-compose.yaml
```diff
@@ -0,0 +1,64 @@
...
+  database:
+    image: "postgres" # use latest official postgres version
+    env_file:
+      - .env # configure postgres
+    volumes:
+      # persist data even if container shuts down
+      - speckle-postgres-data:/var/lib/postgresql/data/
+      # this mounts the initdb folder to the db container
+      # .sql .sh files are executed from this folder in alphabetical order
+      - ./initdb:/docker-entrypoint-initdb.d
```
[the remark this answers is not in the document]
gjedlicska: The initialize script is there to auto create the speckle_dev database and an application specific user. Your solution works, but you have to add the DB manually if I'm not mistaken. If there is another way to get the required db up I'm all ears. :)
Also not having the db root user in the connection string is a lot more safer. These were my reasons for going this route. (I should put them in comments)

docker-compose.yaml
```diff
@@ -0,0 +1,64 @@
...
+      - redis
+    ports:
+      - "3000:3000"
+    environment:
+        DEBUG: "*"
+    volumes:
+      - ./packages/server:/app
+
+  database:
+    image: "postgres" # use latest official postgres version
```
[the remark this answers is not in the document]
gjedlicska: That sounds reasonable :)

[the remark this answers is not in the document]
gjedlicska: Any reason why stick to Pg@12 rather than going with v13?

packages/server/knexfile.js
gjedlicska: Yeah, I think I put down a note somewhere that I don't like the name of the env variable `POSTGRES_URL` while it is actually more of a connection string.

Issue #73, opened 2021-01-10
Speckle Server 2.0 dev deployment


gjedlicska:
To start a discussion on this, I've added my insigts to the question:

To make deployment and development significantly simpler, there should be a single command that with only providing environment variables starts up all the components and configures a fully functional SpeckleServer. 

Additionally automatic HTTPS & TLS with auto renewing certs makes things production ready. 

## Problem definition

This document details a plan to add a fully automated OCI container based runtime to the 2.0 Speckle server. Automatically running all the components together or only parts (for development) is the task. The server consist of 4 distinct parts, frontend, backend, persistent storage and cache.

### Frontend

The frontend is a static vue application. This means, that the artifacts of a production build are a set of html js and css files, that can directly be served to web clients without any server side processing.

Currently the backend nodejs server is responsible for serving these files (but it is not strictly required, the frontend may be served from anywhere as long as it is able to talk to the backend). 

With the current setup of the backend server serving the static files of the fronted, it is not required to create a separate container for the frontend. Only the build artifacts are needed to be copied into the backend server container.

### Backend

The backend is a nodejs express application. This application provides an auth application, a rest api for the connectors to talk to the server, a graphql api for frontend applications to querry data, and serves the static frontend application.

What else is needed here?

### Persistent storage

The data storage layer for the server is a Postgresql database. Currently there is a need for one SpeckleDev database and a user that has table creation and alteration rights in the DB (it should not be the root user in the db). 

Currently the tables and relationships are created by knex database migrations, which should stay that way.

The missing piece on the DB side is the automated creation of an application database and db user

with proper permissions. 

For development purposes a database admin tool is also helpful to get a full dev env up and running.

### Cache

Currently cache is used for auth session data storage. 

Only session secret is used to authenticate? Should there be more steps to make Redis production ready and safe?

### Reverse proxy

From the services above only 2 should be publicly accessible, the backend api and the frontend app. These public resources should also have automatic tls cert handling (its 2021).

This should be handled by a reverse proxy. 

### Overview

Today the industry is heading into a cloud native future, where everything is ran either serverless, managed or in an OCI container.

For development pruposes and some (if not all) production deployment cases a fully container based solution should be preffered. 

Running Postgres on windows is a pain, WSL2 is widely available. With a good guide, I think anyone can get a local environment with docker up and running. 

Possible exceptions from this is a managed database instance and a statically server frontend from some an edge/cdn hosting provider.

There is a solution that ties all requirements together with a stack:

[...]

PR #79, opened 2021-01-19
Docker image recipe containing the Frontend and the Server

peterjgrainger:
Provide a Dockerfile to create an image containing the Frontend and the Server.

Image has two parts. The first builds the frontend static pages. The second minimal part combines the static pages and the server and is stated using a non root user.

Part of #73 

An example `docker-compose` service would be 
```
  server:
    build:
      .
    depends_on:
      - database
      - redis
    ports:
      - "3000:3000"
    environment:
        DEBUG: "speckle:*"
        POSTGRES_URL: postgres://speckle:password@database/speckle_database_name
        REDIS_URL: redis://redis:6379
        SESSION_SECRET: 'keyboard cat'
        STRATEGY_LOCAL: 'true'
        CANONICAL_URL: 'http://localhost:3000'
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,35 +1,46 @@
-# Initial build stage. Will not be included in minimal production image
 FROM node:12.20.1-alpine3.12@sha256:42998ae4420998ff3255fc2d6884e882bd32f06d45b057f4b042e33bf48a1240 as build
+# Having multiple steps in builder doesn't increase the final image size
+# So having verbose steps for readability and caching should be the target 
 
 WORKDIR /opt
 
-COPY packages/frontend .
+# Copy package defs first they are the least likely to change
+# Keeping this order will least likely trigger full rebuild
+COPY packages/frontend/package*.json frontend/
+RUN npm --prefix frontend ci frontend
 
-# Create the static frontend files in the dist folder
-RUN npm ci && \
-    npm run build
-
-# Minimal production image
-FROM node:12.20.1-alpine3.12@sha256:42998ae4420998ff3255fc2d6884e882bd32f06d45b057f4b042e33bf48a1240
-
+COPY packages/server/package*.json server/
 ENV NODE_ENV production
+RUN npm --prefix server ci server
+
+# Copy remaining files across for frontend. Changes to these files 
+# will be more common than changes to the dependencies. This should 
+# speed up rebuilds.
+COPY packages/frontend frontend
+
+WORKDIR /opt/frontend
+RUN npm run build
+
+FROM node:12.20.1-alpine3.12@sha256:42998ae4420998ff3255fc2d6884e882bd32f06d45b057f4b042e33bf48a1240
 
 RUN apk add --no-cache tini=0.19.0-r0
 
-RUN mkdir frontend server
-
-COPY --from=build --chown=node /opt/dist /home/node/frontend/dist
-
-COPY --chown=node packages/server /home/node/server
-
-# Use a non-root user for security best practices
+# Use a non-root user for increased security.
 USER node
+
+ENV NODE_ENV production
+
+# Copy dependencies and static files from build layer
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:14.15.4-alpine3.10@sha256:fe215d05cdde4b7f2a0f546c88a8ddc4f5fa280a204acdfc2383afe901fd6d84 as build
+
+USER node
+
+WORKDIR /home/node
+
+COPY --chown=node packages/ .
+
+RUN npm --prefix frontend install frontend && \
```
gjedlicska, line 9: I think it is better to use `npm ci` instead of using `npm install` it makes the build reproducible, since it doesn't update the package-lock.json if it is not available.

See https://docs.npmjs.com/cli/v6/commands/npm-ci

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+
+COPY --from=build /home/node/frontend/dist frontend/dist
+COPY --chown=node packages/server server
+
+RUN npm --prefix server install server
+
+USER node
+
+WORKDIR /home/node/server
+
```
gjedlicska, line 26: In this form the app is not very graceful. It doesn't trap SIGTERM etc... Stopping it is not a pleasant experience. I would add tini as the entrypoint to fix this problem.
Like [this](https://github.com/krallin/tini).

packages/server/package.json
```diff
@@ -56,7 +56,8 @@
     "pg": "^7.18.2",
     "pg-query-stream": "^3.4.2",
     "redis": "^3.0.2",
-    "zxcvbn": "^4.4.2"
+    "zxcvbn": "^4.4.2",
+    "http-proxy-middleware": "^1.0.6"
```
gjedlicska, line 60: This is still a dev only dependency. It should not be part of a prod build.

I've solved this by moving the import into the if env == DEV in the app.js

Dockerfile
gjedlicska: I did not yet succeed with this solution, but it is not the end of the world I guess

Dockerfile
peterjgrainger: good idea, changed

Dockerfile
peterjgrainger: Yep, you are right it didn't work.... I've done the changes and all is well

peterjgrainger:
@gjedlicska great points about `tini` didn't know it existed. I've tested this container and all seems to be working

peterjgrainger:
@teocomi @didimitrie do you have any thoughts?

didimitrie:
Looks ok to me! I'm not the expert here though. Assume this unblocks #76?

gjedlicska:
@didimitrie sure does. We've been working with @peterjgrainger on this, and he also checked out my :hankey:  too. 

I added a bit of polish to the Dockerfile on my [version](https://github.com/specklesystems/speckle-server/blob/96425f4f4c6d3f5aeecf8f1396aa4a4575f770b0/Dockerfile)

We should merge our efforts and call it a day.

didimitrie:
Happy to merge away. Should i merge this one first, and follow with #76, or how should we go about it?

gjedlicska:
There is one last issue that requires your decision @didimitrie. The pg lib version is currently pinned to version 7, which means the node version has to be kept ~12. I upgraded to pg^8 on my fork and everything works fine with node 14. Both routes are valid, but I seam to recall you had a motive to have #73 done, to free up from dependency conflicts and be able to upgrade package versions.

Which way should this go?

didimitrie:
#73 is more of an issue with our internal staging deployment environment tbh, so can be ignored for now :)

peterjgrainger:
@didimitrie I did a live share with @gjedlicska and this is the best we can do I think :)

If @gjedlicska is happy to merge I am too

gjedlicska (approved):
Looks good to me.

PR #88, opened 2021-01-29
Specify supported node version

nicolas-chaulet:
Hi! I tried to launch the server today and faced the issue that it does not play nicely with node v14. So here is a small PR to help others avoid this trouble!


gjedlicska:
If #76 could be merged, the blocker for node 14 compat (pg lib) would be gone.

Issue #89, opened 2021-01-29
Enhancement: Viewer - Section Box


NGimbal:
Implement a section box as seen in this demo:
[Demo](http://daign.github.io/clipping-with-caps/)

Interaction should follow this pattern (repo for demo above):
[Clipping With Caps](https://github.com/daign/clipping-with-caps)

[Another useful demo](https://threejs.org/examples/?q=clipping#webgl_clipping_stencil)

Steps to implement (approximate):
- [x] Implement SectionBox class that displays geometry, add to Viewer scene
- [x] Implement face picking & hover state using SelectionHelper.js ideally
- [x] Implement dragging interaction
- [x] Align clip planes with box faces
- [x] Apply clip planes to materials
- [x] Limit planes so they can't cross each other while draggin
- [x] Set from bounding box
- [x] Translate, ~Rotate~ (not going to get into ui for this yet)
- [ ] ~Stencils~

I'm actively working on this :) please let me know if you have any thoughts or advice as to how this should work / be implemented.

didimitrie:
Thanks @NGimbal! I'm quite excited to see this happening!

PR #99, opened 2021-02-11
Section box

NGimbal:
All items listed on issue #89 have been completed except for stencils which as @didimitrie mentioned would require significant work. Architects really need their poché tho so it might be worth it.

Notes:
- Section plane related code has been removed from Viewer, SceneObjectHelper but hasn't been removed from repo in this PR
- SelectionHelper has been improved minimally, I'm sure there's further work to generalize it
- Added an additional ...args parameter to the emit function of the EventEmitter so that multiple parameters can be passed to the callback function. [Reference this MDN page.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)
- One question is whether section plane should be added to the renderer or to materials of all objects in the scene. Each has its minor annoyances, for the time being clipping planes are added to the materials as was the case with the section plane previously

Please let me know what the next steps are, if this PR needs to be amended in any way or if I need to know anything for the future!


didimitrie:
Wow, Nick - fabulous! I get the poché problems, but for now let's park it until the loading of objects gets stabilised (ie, merge same material objects into single buffers, or not).

Thanks a bunch for this, 🍺 are owed!

NGimbal:
Makes sense! I enjoyed doing this. I'm really rooting for Speckle!

PR #109, opened 2021-02-25
Viewer: Point + Curve conversion fixes

AlanRynne:
The `Point` structure changed in `speckle-sharp` but was never propagated to the viewer, causing any new entities to not show up.

I tested all the different types of curves using this stream https://staging.speckle.dev/streams/76ad4bd698/objects/f7893b7fdd44aafb2be70c19d33d5baf

- **Deleted** `demo.js` and `demo.js.map` from examples folder and added them to `.gitignore``
- Fixed `PointToBufferGeometry` as well as created a new `PointToVector3` function to consistently convert points.
- Curve entities
  - `Nurbs`: It appears to be some edge cases that are un-supported (see closed nurbs in test stream), so for now, I just skip the whole nurbs conversion and go for the `displayValue`
  - `bbox` props were not being removed, causing the base point to be displayed. This should be fixed in all curves now.
  - Fixed issue where some closed curves where not displaying correctly (where missing the last segment).


packages/viewer/src/modules/Converter.js
```diff
@@ -259,34 +297,50 @@ export default class Coverter {
...
+        let knots = [ obj.knots[0] ]
+        knots = knots.concat( obj.knots )
+        knots.push( knots[knots.length -1] )
+      }
+      else {
+        knots = obj.knots
+      }
+
       // Create the nurbs curve
       const curve = new NURBSCurve( obj.degree, knots, points, null, null )
```
didimitrie, line 345: I'll remove this try catch & remove the console logs, and just comment the code out. we'll get to it when we get to it :)

didimitrie: the main reason is that try catches on hot paths, when we hit the catch always, are mojo killers...

didimitrie (approved):

PR #115, opened 2021-03-05
Cristi/https proxy awareness

cristi8:
Fixes an issue with passport when behind a reverse proxy that handles https

see https://stackoverflow.com/questions/20739744/passportjs-callback-switch-between-http-and-https

This PR makes the express app trust the headers set by the proxy server


cristi8:
Hmm, this PR is mistakenly based on another PR ( https://github.com/specklesystems/speckle-server/pull/114 ).
Only the last commit is new

I can remake if needed

PR #204, opened 2021-04-30
feat(auth): Add Azure AD auth option

peterjgrainger:
Create a new strategy to enable authentication via Azure AD

You will need to create an App registration with the `id_token` permission added.

ad-auth

### DIFF
--- a/packages/server/.env-example
+++ b/packages/server/.env-example
@@ -52,6 +52,12 @@ STRATEGY_LOCAL=true
 # GOOGLE_CLIENT_ID="-> FILL IN <-"
 # GOOGLE_CLIENT_SECRET="-> FILL IN <-"
 
+# STRATEGY_GOOGLE=false
+# STRATEGY_AZURE_AD="-> FILL IN <-"
+# AZURE_AD_IDENTITY_METADATA="-> FILL IN <-"
+# AZURE_AD_CLIENT_ID="-> FILL IN <-"
+# AZURE_AD_ISSUER="-> FILL IN <-"
+# AZURE_AD_CLIENT_SECRET="-> FILL IN <-"
 
 ############################################################
 # Tracing & co.
--- a/packages/server/modules/auth/strategies/azure-ad.js
+++ b/packages/server/modules/auth/strategies/azure-ad.js
@@ -1,0 +1,99 @@
+/* istanbul ignore file */
+'use strict'
+
+const passport = require('passport')
+const OIDCStrategy = require('passport-azure-ad').OIDCStrategy
+const URL = require('url').URL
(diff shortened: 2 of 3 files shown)


packages/server/modules/auth/strategies/azure-ad.js
```diff
@@ -0,0 +1,58 @@
...
+    async (req, res, next) => {
+      const { upn, displayName, _raw } = req.user;
+      let user = {
+        email: upn,
+        name: displayName
+      }
+      if (req.session.suuid) {
+        user.suuid = req.session.suuid
+      }
+      const myUser = await findOrCreateUser({ user: user, rawProfile: _raw })
```
didimitrie, line 43: This bypasses the invite-only behaviour (which is not documented anywhere properly 😅), but check the google strategy for an example correct flow. 

Currently, if the server is invite-only, the expected behaviour is that unless invited, nobody can create an account. That's why there's a bunch of extra checks that need to go in here, but the google strategy should offer a quick copy paste: https://github.com/specklesystems/speckle-server/blob/63340c78b7f02142fdbc3db2459f5b39dba58a6d/packages/server/modules/auth/strategies/google.js#L30-L59

As I'm looking over the flow in there, i think i've spotted a bug of my very own in there. I'll detail in PR next steps for this.

didimitrie (changes requested):
Just some minor changes!

didimitrie:
Ok @peterjgrainger, this looks good! (and also prompted a moment of introspection 😅 ; see #205). For now, if you'll just hack those changes in I'll merge, and will investigate #205 next week / over the weekend. 

Thanks!

didimitrie:
Hi @peterjgrainger, following up re #205: it's been "neutralised".  Please rebase off main, and the correct flow should now be implemented as in the latest version of the google strategy: https://github.com/specklesystems/speckle-server/blob/934be7c5a25ee2acff36b9a66eb40d28d786ae1c/packages/server/modules/auth/strategies/google.js#L28-L82

🙇🏻‍♂️

peterjgrainger:
@didimitrie I think that is all sorted now. Let me know if I need to change anything else.

Tested and works for me :)

packages/server/.env-example
```diff
@@ -52,6 +52,12 @@ STRATEGY_LOCAL=true
 # GOOGLE_CLIENT_ID="-> FILL IN <-"
 # GOOGLE_CLIENT_SECRET="-> FILL IN <-"
 
+# STRATEGY_GOOGLE=false
```
didimitrie, line 55: this looks doubled up 😬 I'll merge, and remove it afterwards

PR #217, opened 2021-05-04
Dim/ad fixes

didimitrie:
Fixes some bugs that escaped me from the azure ad pr 😬 
- missing `package.json` ref for passport lib
- taking user email from `_json` ad response field, rather than the `upn`

@peterjgrainger, quick ping re main change, so you're up to date with it: I've had to rely on `user._json.email` rather than the `upn` - we would not get this in via our org (it would stay undefined). I've cross-checked with 1.0, and it seems I did the same there (https://github.com/speckleworks/SpeckleServer/blob/1690abaaaf3fc3e6a21b70a199e3cfa401997466/app/auth/azure-ad/index.js#L50), so this should be good to go re integrating with Arup's AD - hopefully 🤞 Of course, if not please do let us know.


peterjgrainger:
@didimitrie woops, sorry about that. I think it happened when I did a rebase. Everything works well with Arup AD :)

PR #228, opened 2021-05-08
Adds stream previews

didimitrie:
First pass at closing #210! 

image

and in dark mode: 

image

Also the renderer displays a preview image before loading the model:

image


didimitrie:
ok i just wanna see this on latest

PR #255, opened 2021-05-21
Fix: Embed view improvements for small iframes/screens

AlanRynne:
Fixes specklesystems/admin#71

- Adds responsiveness to top buttons in EmbedView.vue
- Removes interaction events from detail tags
- Hides details by default when screen is small.
- Stream name and branch name are truncated to prevent overflow.
- In `Renderer.vue`, added a condition to remove "Selection Details" from the selection button, and swap it for an icon, when the screen is small


didimitrie (approved):
legit!

PR #271, opened 2021-06-02
delete account feature

cristi8:


didimitrie (approved):
Legit! Didn't look again in detail, but it should be ready to go to latest.

PR #331, opened 2021-07-09
feat(webhooks): initial gql queries & mutations

izzylys:
started implementing a `webhooks` query (extension on `stream`) and the basic create, update, delete mutations. opening a pr for the purpose of discussion!

some questions:
1. should we change the name of the `events` column in `webhooks_config` to something else (triggers, actions, etc)? 

imo it's a bit confusing bc of the `webhooks_events` table. my intuition is that `events` on a webhook should return the history, not the actions that the webhook will be notified for. the services are also named as `dispatchStreamEvent` and `getLastWebhookEvents` for getting the history

2.  is there a reason for the `events` column beaing a json object? 

it seems simpler for it to be a list of strings eg `["branch_create", "commit_create", "stream_update"]`. this would also make it consistent with other conceptually similar mutations which require lists of strings for roles and scopes

### DIFF
--- a/packages/server/modules/webhooks/graph/resolvers/webhooks.js
+++ b/packages/server/modules/webhooks/graph/resolvers/webhooks.js
@@ -28,6 +28,7 @@ module.exports = {
     },
     async webhookDelete( parent, args, context, info ) {
       let deleted = await deleteWebhook( { id: args.id } )
+
       return !!deleted
     }
   }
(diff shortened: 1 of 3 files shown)


cristi8 (approved):

izzylys:
merging so we can test this together. will create a new pr w more updates and tests!

PR #343, opened 2021-07-19
Webhooks

didimitrie:
Adds preliminary webhook support.

@izzylys + @cristi8: what would any leftover TODOs be for this one, or are we good to go?


packages/server/modules/webhooks/graph/schemas/webhooks.graphql
```diff
@@ -0,0 +1,76 @@
+extend type Stream {
+  webhooks(id: String): WebhookCollection
+    @hasRole(role: "stream:owner")
```
didimitrie, line 3: This directive can only be used for checking against the `server_acl`, it doesn't work for named resources. Hence, it should be `@hasRole(role: "server:user")` and resource access should be checked inside the resolver (like we do everywhere else). 

Currently, this prevents anyone but server admins to create webhooks on a stream, which probably explains why it passes tests (which probably will need some updating?).

packages/server/modules/webhooks/graph/resolvers/webhooks.js
```diff
@@ -0,0 +1,44 @@
...
+      let items = await getLastWebhookEvents( { webhookId: parent.id, limit: args.limit } )
+      let totalCount = await getWebhookEventsCount( { webhookId: parent.id } )
+
+      return { items, totalCount }
+    }
+  },
+
+  Mutation: {
+    async webhookCreate( parent, args, context, info ) {
+      let id = await createWebhook( { streamId: args.webhook.streamId, url: args.webhook.url, description: args.webhook.description, secret: args.webhook.secret, enabled: args.webhook.enabled !== false, triggers: args.webhook.triggers } )
```
didimitrie, line 29: We need to add the `await authorizeResolver( context.userId, args.stream.id, 'stream:owner' )` guard in here and everywhere else if it makes sense. 

We can discuss what's viewable by whom, but my £0.02 would be: 

- Stream contributors can see webhooks, but not create them
- Stream owners can see everything

If it's like this already, we're good (i didn't check!)

packages/frontend/src/router/index.js
```diff
@@ -141,6 +141,45 @@ const routes = [
           }
         ]
       },
+      {
+        path: 'settings/:streamId/',
```
didimitrie, line 145: This smells a bit, keep the stream's settings under the stream prefix. So basically: `/stream/${id}/settings` + children routes.

Why? Basically namespacing for the future, eg. when/if we'll have projects, things will get confusing or inconsistent...

didimitrie (commented):
Last remarks: should **globals** be moved under stream settings? Then we'll have less conflicting buttons in the sidebar, and things will be grouped nicely. Otherwise happy to leave as is the two edit buttons (a la github, agree 100 w/h @izzylys that it's fine, pending delete making its way in the main settings as well).

_I know docs update will be a PITA 😭_

packages/server/modules/webhooks/graph/schemas/webhooks.graphql
izzylys: totally my bad! soz for the fuckup 😅
✔ fixed

packages/frontend/src/router/index.js
izzylys: changed it to your second suggestion of `settings/streams/:streamId` !
did not use child route of `stream/:streamId` due to not wanting to have the stream sidebar 
✔ fixed

packages/server/modules/webhooks/graph/resolvers/webhooks.js
izzylys: re this, we previously decided that only `stream:owner`s could see / edit webhooks so that's how it is at the moment

if we want to change this, i can easily do it! note that right now, the settings page as a whole is not visible unless you are a `stream:owner`. i chose to do this bc editing stream details (the other settings page) is only available to owners, so the settings stuff would be entirely read only to a contributor

izzylys: conclusion: leave as is until poss more stuff is added to stream settings that would be relevant to contributors

cristi8 (approved):
looking stable enough to merge

didimitrie:
@cristi8, I saw now your comment! Feel free to do so 😎

PR #363, opened 2021-07-29
Frontend: activity stream & navigation updates

teocomi:
Various frontend changes across the board, mostly around navigation, UI uniformization, and the addition of the activity stream.


teocomi:
Merging this now, and we'll get to Dim's laundry list later.

PR #374, opened 2021-08-10
Add timeline events for user update and create

Mycobee:


packages/server/modules/activitystream/services/index.js
```diff
@@ -91,14 +91,19 @@ module.exports = {
...
-      FROM stream_acl acl
-      INNER JOIN stream_activity act ON acl."resourceId" = act."streamId"
-      WHERE acl."userId" = ? AND time < ?
+        FROM stream_acl acl
+        INNER JOIN stream_activity act ON acl."resourceId" = act."streamId"
+        WHERE acl."userId" = ? AND time < ?
+      UNION
+      SELECT act.*
+        FROM stream_acl acl
+        INNER JOIN stream_activity act ON acl."userId" = act."userId"
```
cristi8, line 100: Do we need a join here? In my tests, where I have multiple streams and multiple rows in `stream_acl` for my user, i get duplicate results

packages/server/modules/activitystream/services/index.js
```diff
@@ -118,10 +123,16 @@ module.exports = {
...
 
   async getTimelineCount( { userId } ) {
-    let [ res ] = await StreamAcl().count()
+    let [ streamActivityRes ] = await StreamAcl().count()
       .innerJoin( 'stream_activity', { 'stream_acl.resourceId': 'stream_activity.streamId' } )
       .where( { 'stream_acl.userId': userId } )
 
-    return parseInt( res.count )
+    let [ userActivityRes ] = await StreamAcl().count()
+      .innerJoin( 'stream_activity', { 'stream_acl.userId': 'stream_activity.userId' } )
```
cristi8, line 131: same join

packages/server/modules/core/services/users.js
```diff
@@ -106,7 +106,19 @@ module.exports = {
     delete user.passwordDigest
     delete user.password
     delete user.email
-    await Users( ).where( { id: id } ).update( user )
+
```
cristi8, line 109: The user_update activity is already saved from the graphql entrypoint:  https://github.com/specklesystems/speckle-server/blob/main/packages/server/modules/core/graph/resolvers/users.js#L89

Mycobee: I see now i was calling the wrong method in my test which is why I added this.  Removed and changed the spec

Mycobee:
@cristi8 I addressed your comments and pushed the new changes. Let me know what you think when you get some time. Thank you 🙌🏼

packages/server/modules/activitystream/services/index.js
```diff
@@ -145,13 +153,15 @@ module.exports = {
     return parseInt( res.count )
   },
 
-  async getTimelineCount( { userId, after, before } ) {
-    let query = StreamAcl().count()
+  async getTimelineCount( { userId } ) {
```
cristi8, line 156: The `before` and `after` parameters were introduced recently and you probably missed the merging of those :)

Mycobee: good catch, how about now ? 😄

Mycobee:
Changes made. Requesting another one @cristi8 . Thanks  !

packages/server/modules/activitystream/services/index.js
```diff
@@ -146,12 +154,14 @@ module.exports = {
...
   async getTimelineCount( { userId, after, before } ) {
-    let query = StreamAcl().count()
+    let [ streamActivityRes ] = await StreamAcl().count()
       .innerJoin( 'stream_activity', { 'stream_acl.resourceId': 'stream_activity.streamId' } )
       .where( { 'stream_acl.userId': userId } )
-    if ( after ) query.andWhere( 'stream_activity.time', '>', after )
-    if ( before ) query.andWhere( 'stream_activity.time', '<', before )
-    let [ res ] = await query
-    return parseInt( res.count )
+
```
cristi8, line 160: The `after` and `before` parameters are not used anymore in your patch.

PR #382, opened 2021-08-16
fix(server): uniquely name new random streams

### LINKED ISSUE #373: Uniquely name new random streams
New streams are all named "Random Stream", if you create multiple it gets hard to identify what is what, without looking at their id.
I'd like to randomly assign them unique names with some sort of library a bit as google does to anonymous docs viewers.

Eg: https://www.npmjs.com/package/unique-names-generator

nemoload:
Fixes #373

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,16905 @@
 {
   "name": "root",
+  "lockfileVersion": 2,
   "requires": true,
-  "lockfileVersion": 1,
+  "packages": {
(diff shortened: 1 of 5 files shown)


teocomi:
Hey @nemoload thanks for this! As mentioned via email, the implementation looks correct, but we might want to have it behave slightly differently (eg with AEC related names/words) and it needs discussing, so I'll close it for now.

PR #395, opened 2021-09-08
frontend fix(registration and password reset)

### LINKED ISSUE #383: Server registration form not playing nicely with password managers
### Expected vs. Actual Behavior

I tried to register a new user on speckle.xzy, and my password manager (1password) was not able to pick up the correct fields.
The username is switched with the organization name.

### Reproduction Steps & System Config (win, osx, web, etc.)

I'm using 1password for this, fill the forms, and this is what i get:
image
 
### Proposed Solution (if any)

This is a minor papercut issue, but I think the fix should be relatively easy, with providing the input tag type properly.

gjedlicska:
add html tags to user and pass fields

Fix password managers autocomplete based on this: https://web.dev/sign-in-form-best-practices/

fix #383


didimitrie (approved):
LGTM!

PR #401, opened 2021-09-10
unique stream names

gjedlicska:
- chore(commitizen config): git cz droped me to a text editor without starting cz session
- feat(stream naming): add unique naming of streams ala docker


packages/server/modules/shared/architects.js
```diff
@@ -0,0 +1,122 @@
+'use strict'
```
didimitrie, line 1: General comment (discussed live with @gjedlicska), but putting it up here for posterity's sake: 

Main problem: I'd not have `Corbusier's Random Stream` as a name because, frankly, it's not his; it's the author's. 

We could restrict names to super famous and long-dead architects, or we fall back to using more generic names; reason being that if i suddenly see `Hector Guimard's Plunky Stream` in my feed I might get totally confused about it - is he my colleague, or actually a famous french architect that lent his name to a random stream? 

---

To the point comment: since this is not used anywhere else, I'll move it into the service rather than in `shared`.

didimitrie (changes requested):
I'll poke at it tonight to shape it in the way I see it without indirect directions! (so i'm requesting changes of myself)

packages/server/modules/shared/architects.js
teocomi: I agree with this, and if we can't find any AEC related names, we can just do ` random adjective` + stream

didimitrie:
superseded by #404 (what an ominous pr number). it's a super basic implementation, but we didn't need more!

Issue #405, opened 2021-09-13
Investigate queries with select clauses containing "@" in variable name


didimitrie:
The server seems to not like "@" in select fields: `syntax error, unexpected '@' at or near \"@\" of jsonpath input`

How to reproduce: try selecting something with an `@` in its name via the table view example app: https://speckle-table-view.netlify.app/

To check as well the app itself, as we might wrongly encode things (but i doubt it).

didimitrie:
Okay, actually this is totally do-able without any changes on our end. The way to be able to query, select and sort by fields with less standard names is to simply escape their key names, like so: `["\"speckle_type\"", "\"@height\"", "\"spaces in names\""]` for the table demo app.

PR #406, opened 2021-09-13
fix(server): surrounds field names with qoutes in get obj children query

didimitrie:
so as to allow for the selection of fields with names like "@height" or "super test". should close #405. @cristi8, theoretically these arguments pass through the knex sanitisation filter, but please have a look!


didimitrie:
closing as not needed; we're good as we were!

PR #415, opened 2021-09-28
Dim/parser

didimitrie:


didimitrie:
@cristi8, the parsing part is more or less done (for now). We can go ahead with the rest of the backend todos; we can do a walkthrough whenever you're up for it to scope next steps.

cristi8 (approved):

Issue #362, opened 2021-07-28
Server Admin Features


cristi8:
User management:
- [x] view server users P1 
- [x] archive server users P2
- [x] delete server users P1 
- [x] add other admin users (see [this](https://speckle.community/t/speckle-v2-multiple-admin-users/1545)) P1 
- [x] batch invite server users P1 

Stream management:
- [ ] ability to delete individual streams P1 
- [ ] view stream sizes for all streams on the server, grouped by user and sorted by total stream size P2

PR #421, opened 2021-10-04
batch invites

gjedlicska:
- feat((frontend) admin invites): initial invites admin page
- feat((frontend) batch invites): add batch invite functionality to admin invites page


gjedlicska:
completes part of #362

PR #466, opened 2021-11-12
Only use lowercase names for local auth strategy

peterjgrainger:
As discussed in the speckle community https://speckle.community/t/users-email-stored-in-lowercase-in-the-server-but-lookup-uses-mixed-case/1996

When using federated access email's don't need to be stored in lowercase and currently new users can't log in to the system from Azure AD.


gjedlicska:
Actually properly handling case insensitive `findOrCreate` user makes you usecase work.

While reviewing your code, I've realized, that our previous implementation was not backwards compatible, since the previously upper case stored users, were inaccessible with the `toLowerCase` comparison. Now it handles it gracefully. Also added some tests for the case above. 

Sry but it was easier to redoit on a new branch. 
Check out my solution at #467

PR #467, opened 2021-11-12
gergo/properEmailCasing

gjedlicska:
- Only use lowercase names for local auth strategy
- fix(core users): handle case insensitive email comparison in a backwards compatible way

### DIFF
--- a/packages/server/modules/core/services/users.js
+++ b/packages/server/modules/core/services/users.js
@@ -27,6 +27,8 @@ const _ensureAtleastOneAdminRemains = async ( userId ) => {
   }
 }
 
+const userByEmailQuery = email => Users( ).where( 'email', 'ilike', `%${email}` )
+
 
 module.exports = {
 
@@ -38,6 +40,7 @@ module.exports = {
 
   async createUser( user ) {
     user.id = crs( { length: 10 } )
+    user.email = user.email.toLowerCase()
 
     if ( user.password ) {
       if ( user.password.length < 8 ) throw new Error( 'Password to short; needs to be 8 characters or longer.' )
@@ -45,7 +48,7 @@ module.exports = {
     }
     delete user.password
 
-    let usr = await Users( ).select( 'id' ).where( { email: user.email } ).first( )
+    let usr = await userByEmailQuery( user.email ).select( 'id' ).first( )
     if ( usr ) throw new Error( 'Email taken. Try logging in?' )
 
     let res = await Users( ).returning( 'id' ).insert( user )
@@ -70,7 +73,7 @@ module.exports = {
   },
 
   async findOrCreateUser( { user, rawProfile } ) {
-    let existingUser = await Users( ).select( 'id' ).where( { email: user.email } ).first( )
+    let existingUser = await userByEmailQuery( user.email ).select( 'id' ).first( )
 
     if ( existingUser ) {
       if ( user.suuid ) {
@@ -102,7 +105,7 @@ module.exports = {
   },
 
   async getUserByEmail( { email } ) {
-    let user = await Users( ).where( { email: email.toLowerCase() } ).select( '*' ).first( )
+    let user = await userByEmailQuery( email ).select( '*' ).first( )
     if ( !user ) return null
     delete user.passwordDigest
     return user
@@ -147,7 +150,7 @@ module.exports = {
   },
 
   async validatePasssword( { email, password } ) {
-    let { passwordDigest } = await Users( ).where( { email: email.toLowerCase() } ).select( 'passwordDigest' ).first( )
+    let { passwordDigest } = await userByEmailQuery( email ).select( 'passwordDigest' ).first( )
     return bcrypt.compare( password, passwordDigest )
   },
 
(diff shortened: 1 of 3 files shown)


gjedlicska:
achieves the same as #466

packages/server/modules/core/services/users.js
```diff
@@ -27,6 +27,8 @@ const _ensureAtleastOneAdminRemains = async ( userId ) => {
   }
 }
 
+const userByEmailQuery = email => Users( ).where( 'email', 'ilike', `%${email}` )
```
peterjgrainger, line 30: You might want to consider using the `lower` function rather than `like` here. It's much more performant and this query gets called a lot. https://www.postgresql.org/docs/current/indexes-expressional.html

peterjgrainger: @gjedlicska

peterjgrainger: Might be worth creating an index on the column as well `CREATE UNIQUE INDEX lower_email_idx ON email ((lower(email)));`

gjedlicska: yes, I know, the reason why it is there, is not using case insensitive compare, breaks older server installations.

So previously we would save the user email as entered like `eXamPle@eMaIL.AddreSs`.
Given this, or the previous implementation, every time we would try to lookup the user with `example@email.address` making it so, that an existing user would be locked out of their valid account.

peterjgrainger: Then in the lookup you use `SELECT username FROM users WHERE lower(email) = 'MiXeD.Case@gmail.com';` with no performance hit

didimitrie:
@peterjgrainger, i'm afraid of creating an index right now, as we'd need to think about it (i suspect this needs a new migration file, which is a hassle 😅 ) - so let's skip that for this bug fix. 

Agreed on the use of lower in the queries. Knex might pose some difficulties here, but I remember doing this for branches - where were also plagued by casing problems... life is a pain - here's the code: 

https://github.com/specklesystems/speckle-server/blob/fb3feec7c6df288a5f830bac3d686aa44acb32f0/packages/server/modules/core/services/branches.js#L62

So I think for that specific line, it would be something like:

```
const userByEmailQuery = email => Users( ).where( knex.raw( 'LOWER(email) = ?', [ email.toLowerCase() ] ) )
```

peterjgrainger:
> @peterjgrainger, i'm afraid of creating an index right now, as we'd need to think about it (i suspect this needs a new migration file, which is a hassle 😅 ) - so let's skip that for this bug fix.
> 
> Agreed on the use of lower in the queries. Knex might pose some difficulties here, but I remember doing this for branches - where were also plagued by casing problems... life is a pain - here's the code:
> 
> https://github.com/specklesystems/speckle-server/blob/fb3feec7c6df288a5f830bac3d686aa44acb32f0/packages/server/modules/core/services/branches.js#L62
> 
> So I think for that specific line, it would be something like:
> 
> ```
> const userByEmailQuery = email => Users( ).where( knex.raw( 'LOWER(email) = ?', [ email.toLowerCase() ] ) )
> ```

Does the email column currently have an index? If not then it won't be affected. If it does then the index would be pointless. So yeah, if it works, looks good :)

peterjgrainger:
@gjedlicska, what happens if there are two emails now stored for a user? Hi@email.com (before your change) and hi@email.com. Would this code be OK with that?

gjedlicska:
@didimitrie where has a `whereRaw` equivalent, gonna give that a go.
@peterjgrainger with our current implementation an email stored previously as uppercase, and the same email in lowercase is not the same user. Storing 2 emails for a user doesn't seam like a good idea... Also, as Dimitrie said, adding the extra migration to the DB seams like an overkill, if the whereRaw with SQL lower works.

peterjgrainger:
> @didimitrie where has a `whereRaw` equivalent, gonna give that a go. @peterjgrainger with our current implementation an email stored previously as uppercase, and the same email in lowercase is not the same user. Storing 2 emails for a user doesn't seam like a good idea... Also, as Dimitrie said, adding the extra migration to the DB seams like an overkill, if the whereRaw with SQL lower works.
> 
> Good thing i've refactored the base query, so now we only need to change it in one place.

My meaning is if you have those two users and you run the query `select * from users where LOWER(email) = 'hi@email.com';` you will get two results as they are now now unique

peterjgrainger:
I think the best option, if you want to keep backwards compatibility is to run a migration to change all the emails to lowercase and add the index then merge that code. Then you won't get an issue.

Either that, or communicate to everyone that they should do this on their own database manually, or make the email case insensitive again

gjedlicska:
Naturally we don't want to break anyone's deployment. If you check the test added in this PR, it ensures, that if someone registered with an email that has uppercase characters it still works with the new toLowerCase comparison. 
The only edge case of this edge case, which is not covered, if someone registered with an upper email, lost access to their account, and reregistered with the same email. This would result two separate users.
So far, this case hasn't come up on any of our deployments.

Thus I don't see there is a need to create any migrations fortunately.

If this highly unlikely case exists for you let us know, and will figure it out.

peterjgrainger:
Yes this is the case for many users of our system because we use federated access with ACtive directory. Those that logged into the system in the last month now have two accounts. One with the same case as in active directory and one with lowercase. I'd be surprised if others using federated access don't have the same issue

PR #475, opened 2021-11-19
gergo/userMigration

gjedlicska:
- chore(server): add duplicate user migration script
- feat(user migration): update with proper stream and server acl, user cleanup
- refactor(server user migration): remove dummy data creation
- feat(server core migrations): add user deduplicating migration

### DIFF
--- a/packages/server/modules/core/migrations/20211119105730_de_duplicate_users.js
+++ b/packages/server/modules/core/migrations/20211119105730_de_duplicate_users.js
@@ -1,0 +1,108 @@
+/*
+This migration is fixing the duplicate user problem reported 
+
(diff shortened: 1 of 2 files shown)


gjedlicska:
Hey @peterjgrainger,
we've added a migration to de-duplicate the uppercase users, and consolidate them to 1.
Hopefully this fixes your issue too.

peterjgrainger:
That looks great @gjedlicska! I would have just created a migration that fails if there are duplicates and let the owner of the DB change it. This must have been a decent amount of effort!

Now it's possible to put in an index on so making sure it's unique isn't enforced just in the code. There is a bit of a risk in 6 months / a year someone adds some code and doesn't realise that you should use the new function you made to add users in lowercase and look up in lowercase.

Thanks for this!

PR #476, opened 2021-11-23
gergo/streamAdmin

gjedlicska:
- feat(frontend): add WIP stream admin page
- feat(frontend, server): add stream admin page with backend
- test(server): add tests for stream admin stuff
- refactor(frontend, server): backport stream admin route handling solution to user admin


didimitrie (changes requested):
Looks good, two problems I've encountered and a nitpick: 

**Sorting by size:** 
When sorting by stream size, either `asc` or `desc`, returned size is null, and i can't tell if it's doing the right thing. 

The frontend simply displays `0.00MB` as size. As this is headline functionality re this PR, it got me a bit flummoxed.

I'm testing against a dump of the staging db. 

**Opening streams**
A quick button to open the respective stream in a new tab would be really welcome. Currently the list item is clickable but doesn't do anything. 

(same observation for users admin, could we add a link there to open up their profile page?) 

**Minor nitpick on the way collaborators are displayed:** 
The call received icon currently is confusing me in the collabs display. I'd get rid of it.

gjedlicska:
Stream size just needed some extra coalesceing.
For opening streams, if the admin doesn't have access to the stream, it gets a u dont have access page, if thats ok, i can add the new tab link.

gjedlicska:
ok, fixed stuff.

Getting the streams ordered by size is going to be a loooong operation on bigger servers, so we'll need something more performant down the line.

didimitrie (approved):

PR #477, opened 2021-11-23
fix(frontend): replace streamid with streamId in frontend gql query

### LINKED ISSUE #232: Bug: inconsistent query parameter names in GQL
Noticed some query parameter names are slightly inconsistent, should go through the repo and uniform them.
Note: this is a breaking change, but we should do it before the end of the beta!

image

gjedlicska:
fixes #232

### DIFF
--- a/packages/frontend/src/components/SearchBar.vue
+++ b/packages/frontend/src/components/SearchBar.vue
@@ -26,7 +26,7 @@
             <v-row class="pa-0 ma-0">
               {{ item.name }}
               <v-spacer></v-spacer>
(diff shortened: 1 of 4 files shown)


gjedlicska:
This was a super simple fix @teocomi only the frontend query parameter name was mistyped.
Anyhow I've checked sharp and py too, no occurence of `streamid` found.

didimitrie:
@gjedlicska, i'm going to merge this one on faith, as @teocomi's busy anyway with other stuff. The tests you put it through seem legit, so 🤞

PR #485, opened 2021-11-30
edit app and delete commits functions added

KatKatKateryna:


didimitrie:
Hey @KatKatKateryna, this looks super good! As self appointed fashion police, i'm swooping in with some nitpicks:

1) I know the mobile version of the top bar is lacking from before, but this looks even more broken now 😅  

image

To resolve this, I'd potentially move the delete button inside the edit dialog of the commit. To resolve the nesting, i'd simply make the `<commit-edit-dialog ref="commitDialog" @show-delete="showDeleteDialog=true"></commit-edit-dialog>` emit a "show delete" event of some sorts, and then you don't need to move logic around the components. 

2) The tooltip for the delete button still shows 'edit commit':

image

3) On editing an app: 

The dialog shows fine and does what it's supposed to, but I have some little to-dos: 

3a) when editing, populate the scopes select with the ones from the existing app. If i want to just edit the name, currently one still needs to add the scopes back again, and without feedback, it's cumbersome. 

I'd also do a little internal diff to check if old scopes === new scopes, and then omit the fully from the payload. 

3b) add a disclaimer in the edit app dialog that says: 

```
After editing an app, all users will need to authorise it again (existing tokens will be invalidated). 
```

This is to prevent scope escalation and be transparent towards end users (scenario: app with little permissions should not be able to upgrade itself without the user's consent). 

Sorry if I jumped on reviewing this too soon, but i figured i'd do it now since i have some time and energy left over (and it's not marked as WIP either...)

KatKatKateryna:
Great, thanks @didimitrie ! 
All the points addressed, plus: 
- after deleting commit, it redirects to Branch page and refreshes
- App edit "Success dialog" resets back to "Edit dialog" if we close and open it again
- SAVE button disabled if no changes detected in App details

didimitrie (approved):
@KatKatKateryna, feel free to merge this one!

Issue #444, opened 2021-10-27
Viewer: n-gon support 


didimitrie:
P0: 
- [x] support `3` and for `4` for face indexes

P2: 
- [ ] properly support n-gons (includes some triangulation logic that I fear because never looked at this. see https://github.com/specklesystems/speckle-sharp/issues/780 discussion)

PR #486, opened 2021-11-30
Updated viewer conversion to support 3,4 cardinality faces, and not throw exception when n-gon faces are present

JR-Morgan:
Fixes the P0 part of #444:
 - Mesh faces with 3/4 cardinality indicators will now be converted as triangles/quads respecfully, previosuly they were ignored and only faces with 0/1 indicators would be converted.

- N-gon faces on a mesh will be ignored, previosuly, the entire mesh object would be ignored.

The seccond part of #444, Full N-gon conversion, is still not complete, and will likley require a triangulation function.

Has been tested in the example.html viewer and behaves expectedly.

### DIFF
--- a/packages/viewer/package.json
+++ b/packages/viewer/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@speckle/viewer",
-  "version": "2.2.2",
+  "version": "2.2.3",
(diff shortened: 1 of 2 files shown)


didimitrie (approved):
looking forward to full n-gon support 😎

PR #509, opened 2022-01-04
Kate/presentation constructor as globals

KatKatKateryna:
## Presentation viewer features:

- Shows distinct renderer for branches called "presentations/"+ something
- Each saved or published presentation is saved as a new commit (PUBLISH function saves 2 commits: editable and non-editable)
- Uploading all existing branches into viewer as Layers (when Published, uploading only the branches that exist in the presentation)
- Creating and deleting slides (message + branch visibility + animation + camera position)
- Hiding editing options if a presentation status is converted to "published"
- Animation toggle for branch iterates through the objects in that layer
- Animation can be added to the layer, and while presenting, animation will continue uninterrupted if it is also activated on the subsequent slides

## Limitations:

- **Works properly only with branches with "traditional" Speckle objects** - where the structure was not significantly changed. For example, if you modify Speckle object in Grasshopper, expanding the object properties will work fine, but the branch will be not displayed properly if the geometry itself has been assigned a custom name.


didimitrie:
Moved into our R[&D table](https://www.notion.so/speckle/R-D-0bf08c8198644ab88e013a8b05da3411?p=43ddf204dae843aba9820d034735ae66)

PR #527, opened 2022-01-18
viewer parents is now an array, added includes filters

cristi8:


cristi8:
will be integrated in dim's following changes / branch

PR #549, opened 2022-01-25
Create CodeSpaces Configuration

peterjgrainger:
Hi 👋 .

I was making a codespaces https://github.com/features/codespaces for my org and thought you might want this as well. It's in beta at the minute for individual access but GA for organisations. Not sure how many people using speckle have this configured but if they do it will greatly speed up their onboarding.

If you don't want to add it no worries, you might change your mind in the future when it's more mainstream.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,1 @@
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:16-bullseye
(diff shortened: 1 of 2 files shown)


gjedlicska:
Hey @peterjgrainger,

thanks for the initiative, this is a neat concept, that we'd be happy to add to the server project. 
Ur current solution targets Codespaces and works fine on linux desktop too, but unfortunately it did not work on either windows or mac desktop. Since VSCode pops a big banner in your face, if it sees a devcontainer config, i think we need a solution that would work on all desktop OS-es too.

Btw i like the socker mounting magic, but that's what is causing permission errors.

If you can come up with a solution I'd be happy to proceed with this, otherwise I will look into this at some point.

peterjgrainger:
I developed this on a Mac, what errors did you get? Maybe I can figure out what is going wrong. FYI first load takes ages

izzylys:
heya, this was the error I got on windows. the same issue occurred when Alan tried it on mac

image

PR #555, opened 2022-01-27
Kate/viewer maps

KatKatKateryna:
Added 2 modules:
- SceneSurroundings, with functions to add/remove base maps and 3d buildings, change styles and adjust map location, scale and rotation angle
- 2nd module adapted from https://github.com/skewart/osmthree to send overpass API requests for 3d osm buildings

Added dependencies: geo-three


didimitrie:
I know this is a draft, but i'd like to point out a big current itch:

Dependencies: 
- `osmthree.js` seems to come from [this repo](https://github.com/skewart/osmthree) originally, which doesn't have a license block (or npm presence) - we need to get to the bottom of the source & license, otherwise it's risky to use it.

And some style itches: 
- OSM3 is imported in the SceneSurroundings class but used from the window global space
- Linting and indentation does not respect package norms (`var`s, etc.) 

There's a set of bigger todo-s to expose this feature in the mainline, namely:
- extend the server settings table with a mabbox api key that can be provided by server admins in the server settings section of admin
- provide that via the serverInfoQuery and use it by default
- store lat/long + other settings in globals (per commit?)
- add frontend button for displaying/setting/hiding map 
- ...and a host of other todos - i'll leave those for later

PR #608, opened 2022-02-28
Feature/isomorphic object loader

messismore:
Hi! 🙂

This pull request adds the option to provide `ObjectLoader` with a custom `fetch` for server side use. 

@didimitrie @teocomi Do you think this is a viable approach? If so, I'll add tests.

Went with [undici](https://github.com/nodejs/undici) as it will become Node's fetch implementation in v18, even though it is still [considered experimental](https://undici.nodejs.org/#/?id=undicifetchinput-init-promise) at this time.

### DIFF
--- a/packages/objectloader/examples/node/script.js
+++ b/packages/objectloader/examples/node/script.js
@@ -1,1 +1,24 @@
-// NOTE: This lib is not working in node, because node-fetch returns node-native readable streams - we need a workaround first. 
+// Since Node v<18 does not provide fetch, we need to pass it in the options object. Note that fetch must return a WHATWG compliant stream, so cross-fetch won't work, but node/undici's implementation will.
+
+import { fetch } from 'undici'
+import ObjectLoader from '../../index.js'
+
+let loader = new ObjectLoader({
+  serverUrl:"https://latest.speckle.dev",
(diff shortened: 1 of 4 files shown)


didimitrie:
Hey @messismore, thanks for starting this - it's something I wanted to happens since a long time! Let us know if you get stuck and how we can help 🙇

didimitrie:
Nice! I thought it's WIP as the PR was still draft. 

One thing that comes to mind right now - i'm "typing out loud" so we're sure of this - currently there was no build step per se as the lib didn't have any dependencies, and we could publish on npm straight away. I like the approach you took, where by the objectloader actually **still doesn't have a strong dependency**, and in a node environment it's the **developer's responsibility** to provide an adequate fetch (undici) until node finally takes it out from under the feature flag 💯

Basically, with adding a note in the readme about this, we're good to go!

messismore:
> and in a node environment it's the developer's responsibility to provide an adequate fetch

"BYO `fetch`" should make it easy to implement bespoke cashing logic, too, which should also come handy on the server.

I updated the readme, so this should be all set!

didimitrie:
Hurray! I'll merge this now - apologies for the delay, week's been a bit crazy!

PR #618, opened 2022-03-14
fix: linting & formatting config fixes #617

### LINKED ISSUE #617: Formalizing & standardizing linting & formatting across repo
The monorepo has different linting configs in each package, different ESLint versions even and no standard approach when it comes to formatting.

I'm going to try to refactor this area and arrive at consistent & DRY configs, the same ESlint installation used across the repo & standard formatting.

fabis94:
closes #617


fabis94:
- Moved all ESLint installations to the same version (latest)
- Created a monorepo base config that is used by all packages and extended when necessary, resulting in a much more clean and DRY config
- Added linting commands to each package, so you can easily run linter through `npm run lint`
- Got rid of all formatting related ESLint rules and just switched to using prettier for formatting, achieving a better separation of concerns, cleaner configs and ability to format not just JS, but also CSS, HTML using the same tool. For the prettier config I used the config that was already there for the frontend package.
- Added VSCode "suggested extensions" config file so that anyone developing here always has ESLint, Prettier & Vetur set up & enabled
- Added jsconfig.json in all repos for better interop with VSCode & somewhat improved intellisense
- See this file for example of how prettier will format JS files from now on: https://github.com/specklesystems/speckle-server/pull/618/files#diff-047796c07a2f325f1796eba4322ae8730d58ec543a991be634ffb8e975a8f94c
- Removed preview-service/public from repo since its a build artifact

gjedlicska:
I dont have enough XP with eslint + prettier to pinpoint anything that would break for us. Based on my testing it works for us, and is a good baseline.

.gitignore
```diff
@@ -6,7 +6,6 @@ packages/viewer/dist
 *.crt
 .nyc_output
 coverage/
-.vscode
```
gjedlicska, line 9: removing this would make commiting the launch.json files possible. I think we should not allow that.

fabis94: Why not? Would result in better DX imo, considering that each dev wouldn't have to come up with his own launch config, but could use one of the standardized ones

fabis94: if possible i'd lock this folder to be only editable by the speckle team, however

gjedlicska: for specific testing purposes i usually change some variables, ie bake in the regex, for the test runner, or override some environment variables just for a specific startup config. i kinda like doing that, since it allows me to run different scenarios. These are usually not that applicable to others and sometimes might be os specific too.

The npm scripts part of vscode is clever enough to pick up all the subpackage scripts, which works perfectly if you don't want to customize stuff. I'd guide newcommers towards that and leave the launch.json as a per machine option open.
image

fabis94: @gjedlicska I think we should discuss this further in chat, I think that in most scenarios you should be able to get away with using standard debug configs.

PR #624, opened 2022-03-17
feat(server): configurable .env.test & auto-creation of the test db

fabis94:
- .env.test should be created from .env.test-example, which enables it to be configured per user. If it isn't created, tests will run in the main DB (relying on main env vars)
- Auto-creating both DBs during docker DB volume setup


gjedlicska (approved):
it works on a new setup for me too

Issue #620, opened 2022-03-16
Add feature to favorite streams


fabis94:
Favoriting streams and being able to browse favorited streams in the web

Partially completes #483

PR #633, opened 2022-03-23
feat: favoriting streams #620

fabis94:
Lots of changes, but honestly most are due to the auto-format & changes in lock files.

I've tried to leave comments to explain the changes in the PR so you know what to focus on


packages/frontend/src/graphql/server.js
```diff
@@ -0,0 +1,60 @@
+import gql from 'graphql-tag'
+
```
fabis94, line 2: Most of the graphql query definition changes can be summarized as follows
1. Using a .js file for defining queries is better, because you can 1) have multiple queries in the same file 2) compose queries from various fragments, increasing DRYness
2. Defining standard queries for optimized cache usage. Instead of executing 3 different user/serverInfo queries causing unnecessary requests, have all components rely on the same query so that cache can be re-used. (I was able to reduce GQL request amount on Streams page from 9 to 3)

packages/frontend/src/main/components/admin/ActivityCard.vue
```diff
@@ -7,7 +7,7 @@
           <v-icon x-small color="primary" class="mr-1">{{ icons[value.name] }}</v-icon>
           {{ capitalize(value.name.split('History')[0]) }} history
         </p>
-        <apexchart class="primary--text" type="bar" :options="options" :series="[value]" />
+        <apex-chart class="primary--text" type="bar" :options="options" :series="[value]" />
```
fabis94, line 10: Vue linter asks for multi-word component names

packages/frontend/src/main/components/common/StreamPreviewCard.vue
```diff
@@ -63,14 +67,25 @@
 <script>
 export default {
   components: {
-    PreviewImage: () => import('@/main/components/common/PreviewImage'),
-    CollaboratorsDisplay: () => import('@/main/components/stream/CollaboratorsDisplay')
+    PreviewImage: () => import('@/main/components/common/PreviewImage.vue'),
```
fabis94, line 70: VSCode expects me to add the .vue extension for it to be able to understand what file this points to

packages/frontend/src/main/components/stream/CollaboratorsDisplay.vue
```diff
@@ -20,7 +20,7 @@
         <span class="caption">+{{ collaborators.length - 4 }}</span>
       </v-btn>
       <v-btn
-        v-if="collaborators.length <= 5"
+        v-if="stream.role === Roles.Stream.Owner && collaborators.length <= 5"
```
fabis94, line 23: Hiding "Manage collaborators" button from streams toolbar if you're not the owner

packages/frontend/src/main/components/stream/favorites/StreamFavoriteBtn.vue
```diff
@@ -0,0 +1,121 @@
...
+          __typename: 'Mutation',
+          streamFavorite: {
+            __typename: 'Stream',
+            id,
+            favoritedDate: newFavoritedDate,
+            favoritesCount: newFavoritesCount
+          }
+        },
+        update: (cache, { data: { streamFavorite } }) => {
+          const { id, favoritedDate } = streamFavorite || {}
```
fabis94, line 65: This part's cool - using the update function I'm updating the results of the "favorite streams" query, so that favorite streams are added/removed without needing to refetch from server

packages/frontend/src/plugins/authHelpers.js
```diff
@@ -24,37 +27,33 @@ export async function checkAccessCodeAndGetTokens() {
...
-      localStorage.setItem('distinct_id', distinct_id)
-      localStorage.setItem('uuid', data.user.id)
-      localStorage.setItem('stcount', data.user.streams.totalCount)
-      return data
-    } else {
-      await signOut()
-      throw new Error('Failed to set user')
-    }
+  // Pull user info (& remember it in the Apollo cache)
+  const { data } = await apolloClient.query({
```
fabis94, line 39: Moved from `fetch`  to an apollo client query so that we get to keep the results in the Apollo cache

packages/frontend/src/vue-apollo.js
```diff
@@ -2,117 +2,116 @@
 import Vue from 'vue'
 import VueApollo from 'vue-apollo'
 import { createApolloClient, restartWebsockets } from 'vue-cli-plugin-apollo/graphql-client'
-import { SubscriptionClient } from 'subscriptions-transport-ws';
+import { SubscriptionClient } from 'subscriptions-transport-ws'
+import { LocalStorageKeys } from '@/helpers/mainConstants'
```
fabis94, line 6: Moved everything to functions here (instead of triggering object creation as a side-effect when importing htis module) cause originally I planned to use different ApolloProvider instances. I ended up not needing to do it, but getting rid of side-effects is good anyway, so kept these changes (everything works the same as before, it's just refactored)

packages/server/app.js
```diff
@@ -71,7 +68,7 @@ exports.init = async () => {
   })
   graphqlServer = new ApolloServer({
     ...graph(),
-    context: contextApiTokenHelper,
+    context: buildContext,
```
fabis94, line 71: `contextApiTokenHelper` is very unintuitive as a name for a context builder function

packages/server/bin/cli
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env node
+'use strict'
+const path = require('path')
+const yargs = require('yargs')
+require('../bootstrap')
+
+const execution = yargs
```
fabis94, line 7: I had an issue where the `knex` CLI couldn't generate new migrations in the proper `module` directory, it just doesn't support choosing a target directory. Additionally I also remember Gergo talking about occasionally creating new scripts for random testing/debugging and needing to build separate debugging launch profiles for each of them.

To solve both issues I added `yargs` which allows for easy/fast CLI building. Currently I've defined 3 commands for it: migrating up, migrating down and adding a new migration into the correct module, but we can now add all kinds of CLI commands using `yargs`

packages/server/db/knex.js
```diff
@@ -18,4 +19,11 @@ const debug = require('debug')
...
 
-module.exports = require('knex')(conf)
+/**
+ * Need to override type because type def file incorrectly uses ES6
+ * @type {import('knex').default}
+ */
+const knex = require('knex')
+const knexInstance = knex(config)
+
+module.exports = knexInstance
```
fabis94, line 29: Most of these changes are to get better types for the exports here

packages/server/knexfile.js
```diff
@@ -19,7 +19,7 @@ function walk(dir) {
   return results
 }
 
-let migrationDirs = walk('./modules')
+let migrationDirs = walk(path.resolve(__dirname, './modules'))
```
fabis94, line 22: Had to fix this otherwise knex couldn't find these directories if you invoked it from somewhere other than the `server` package root

packages/server/modules/core/dbSchema.js
```diff
@@ -0,0 +1,40 @@
+const knex = require('@/db/knex')
+
+/**
+ * Single source of truth for DB schema in the codebase
```
fabis94, line 4: Context: https://discord.com/channels/726756379083145269/743849923803349022/954402440269140008

Basically, instead of duplicating magic strings (table, column names) everywhere and having no visibility into whether or not the thing you're referencing actually exists (is a valid table/column name), I created this single source of truth for knex instances, table names and columns names.

packages/server/modules/core/graph/resolvers/users.js
```diff
@@ -1,154 +1,163 @@
...
-
     async _() {
       return `Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn.`
     },
 
-    async user( parent, args, context, info ) {
-      await validateServerRole( context, 'server:user' )
+    async user(parent, args, context, info) {
+      // User wants info about himself and he's not authenticated - just return null
+      if (!context.auth && !args.id) return null
```
fabis94, line 30: From now on the query will not fail if user is unauthenticated. `null` is a valid value representing that the user isn't logged in

packages/server/modules/core/graph/schemas/streams.graphql
```diff
@@ -32,13 +32,24 @@ type Stream {
...
 extend type User {
   """
   All the streams that a user has access to.
   """
   streams(limit: Int! = 25, cursor: String): StreamCollection
+
+  """
+  All the streams that a user has favorited
+  """
+  favoriteStreams(limit: Int! = 25, cursor: String): StreamCollection
```
fabis94, line 52: Makes more sense for this to be a User field not a top level query.

(It's kinda weird how you can do `User.streams` and also just `streams` and both queries nearly do the same thing.)

packages/server/modules/core/loaders.js
```diff
@@ -0,0 +1,58 @@
+const _ = require('lodash')
+const { StreamFavorites, knex } = require('@/modules/core/dbSchema')
+const DataLoader = require('dataloader')
+
```
fabis94, line 4: Introducing request-scoped dataloaders. Why? Cause with the addition of favorites, the main `getStreams` function would have had 2 extra joins (cant get favoritedDate and favoritesCount with the same join). So, just to get streams you'd have to join on ACL & Favorites (twice). Imagine you're only looking for stream ids - there's gonna be a lot of overfetching!

So it's better to only load the data that is being asked - the favorites table should only be queried if the request asks for the `Stream.favoritedDate` field. Moving all "joins" to their own resolvers introduces the [N+1 problem](https://dev.to/kenwanjohi/solving-graphql-n-1-problem-in-fastify-with-loaders-and-conditional-queries-380o), however. Basically, if you'd query for 100 streams and for each of them ask for the `Stream.favoritedDate` field, GraphQL would execute the `Stream.favoritedDate` resolver 100 times each time invoking an SQL query - that's a ton of SQL operations slowing everything down!

And the solution to that is [dataloaders](https://github.com/graphql/dataloader). What it does is batches calls to retrieve some data from somewhere (e.g. execute an SQL select) and then processes all of those calls only once. In the `Stream.favoritedDate` example this would mean that using DataLoaders I'd first batch all requests to get the favorited date for all 100 streams, and then once the DataLoader batch loader function would execute I'd only do a single SQL query asking for favorites info for all 100 streams.

packages/server/modules/core/services/objects.js
```diff
@@ -1,283 +1,323 @@
...
-const bcrypt = require( 'bcrypt' )
-const crs = require( 'crypto-random-string' )
-const { performance } = require( 'perf_hooks' )
-const crypto = require( 'crypto' )
-const set = require( 'lodash.set' )
-const get = require( 'lodash.get' )
-const chunk = require( 'lodash.chunk' )
+const { performance } = require('perf_hooks')
+const crypto = require('crypto')
+const { set, get, chunk } = require('lodash')
```
fabis94, line 4: Replaced lodash and this triggered all the freformatting.

packages/server/modules/core/services/streams.js
```diff
@@ -1,13 +1,85 @@
...
-const debug = require('debug')
-const { createBranch } = require('./branches')
+/**
+ * Get base query for finding or counting user streams
+ * @param {object} p
+ * @param {string} [p.searchQuery] Filter by name/description/id
+ * @param {boolean} [p.publicOnly] Whether to only look for public streams
+ * @param {string} p.userId The user's ID
+ */
+function getUserStreamsQueryBase({ userId, publicOnly, searchQuery }) {
```
fabis94, line 30: Moved out this setup to its own function cause it was duplicated in `getUserStreams` and `getUserStreamsCount`

packages/server/modules/core/tests/generic.spec.js
```diff
@@ -1,85 +1,86 @@
 /* istanbul ignore file */
-const expect = require( 'chai' ).expect
-const assert = require( 'assert' )
+const expect = require('chai').expect
+const assert = require('assert')
```
fabis94, line 3: Only changes: contextApiTokenHelper rename + reformat

packages/server/modules/index.js
```diff
@@ -1,33 +1,40 @@
 'use strict'
-const fs = require( 'fs' )
-const path = require( 'path' )
-const appRoot = require( 'app-root-path' )
-const autoload = require( 'auto-load' )
-const values = require( 'lodash.values' )
-const merge = require( 'lodash.merge' )
-const debug = require( 'debug' )( 'speckle:modules' )
-const { scalarResolvers, scalarSchemas } = require( './core/graph/scalars' )
+const fs = require('fs')
```
fabis94, line 2: Updated lodash import + reformat

packages/server/modules/shared/index.js
```diff
@@ -1,46 +1,58 @@
...
+const { RedisPubSub } = require('graphql-redis-subscriptions')
+const { buildRequestLoaders } = require('@/modules/core/loaders')
+const { validateToken } = require(`${appRoot}/modules/core/services/tokens`)
+
+let pubsub = new RedisPubSub({
+  publisher: new Redis(process.env.REDIS_URL),
+  subscriber: new Redis(process.env.REDIS_URL)
+})
+
+async function buildContext({ req, connection }) {
```
fabis94, line 16: 1. better name for a function that builds the GQL context
2. added request-scoped dataloaders onto the ctx object, so that they can be accessed in any resolver

packages/server/modules/stats/tests/stats.spec.js
```diff
@@ -27,8 +25,7 @@ const params = { numUsers: 25, numStreams: 30, numObjects: 100, numCommits: 100
 
 describe('Server stats services @stats-services', function () {
   before(async function () {
-    this.timeout(10000)
-
+    this.timeout(15000)
```
fabis94, line 28: Had to increase the timeout since these tests easily went over 10s with most of that time coming from the setup & seeding.

I didn't actually change any tests, tho.

packages/server/package.json
```diff
@@ -45,13 +46,9 @@
     "graphql-tools": "^4.0.7",
     "ioredis": "^4.19.4",
     "knex": "^1.0.3",
-    "lodash.chunk": "^4.2.0",
-    "lodash.debounce": "^4.0.8",
-    "lodash.get": "^4.4.2",
-    "lodash.merge": "^4.6.2",
-    "lodash.set": "^4.3.2",
-    "lodash.values": "^4.3.0",
+    "lodash": "^4.17.21",
```
fabis94, line 49: For the server-side it's not necessary to install each lodash utility separately, even if the main `lodash` package wouldn't be optimized. We don't have the same kinds of memory/CPU constraints that we do on the frontend, where each user has a different device with different specs.

packages/server/test/hooks.js
```diff
@@ -58,20 +56,18 @@ exports.mochaHooks = {
     await unlock()
     await knex.migrate.rollback()
     await knex.migrate.latest()
-    console.log( 'running before all' )
+    console.log('running before all')
   },
   afterAll: async () => {
     await unlock()
-    await knex.migrate.rollback()
```
fabis94, line 65: Removed rollback since it's already done on `beforeAll`

packages/viewer/example/example.html
```diff
@@ -1,74 +1,87 @@
 <!DOCTYPE html>
 <html>
+  <head>
```
fabis94, line 3: Honestly not sure why I touched this one 🤔 I think this only has formatting changes

fabis94:
I know I'm missing tests, that's what I'm going to be working on next

packages/frontend/src/main/components/stream/favorites/StreamFavoriteBtn.vue
fabis94: I'm also being very careful to not mutate existing cache objects, in case you're wondering why I don't replace property values on old objects directly

packages/server/modules/core/loaders.js
fabis94: This could be done more modularly, but that would've broken typehinting and I really hate working without typehints :D With the introduction of TypeScript it would be easier to modularize this

gjedlicska:
> Vue linter asks for multi-word component names 

we prob should disable that, or its going to be a massive rename show

gjedlicska:
why did u mark the PR draft?

packages/viewer/example/example.html
```diff
@@ -1,74 +1,87 @@
...
-    integrity="sha512-EZLkOqwILORob+p0BXZc+Vm3RgJBOe1Iq/0fiI7r/wJgzOFZMlsqTa29UEl6v6U6gsV4uIpsNZoV32YZqrCRCQ=="
-    crossorigin="anonymous" />
-  <style type="text/css">
-    #renderer {
-      height: 700px;
-      width: 100%;
-    }
-  </style>
-<script defer src="demo.js"></script></head>
+  <body>
```
gjedlicska, line 28: the bg is missing from here

packages/viewer/example/example.html
```diff
@@ -1,74 +1,87 @@
...
+            Click an object to select it. Double click it to focus on it. Press `esc` to clear the
+            selection. Press `shift-s` to toggle a section plane. Press `s` while the section plane
+            is active to toggle its control mode. Double click anywhere outside an object to zoom
+            extents to the entire scene.
+          </p>
+        </div>
+        <div class="twelve columns">
+          <button onclick="v.postprocessing = !v.postprocessing">Postprocessing Toggle</button>
+          <button onclick="v.interactions.zoomExtents()">Zoom Extents</button>
+          <button onclick="v.interactions.toggleSectionBox()">Toggle Section Box</button>
```
gjedlicska, line 44: and this interactions is added

packages/viewer/example/example.html
gjedlicska: this looks like, it has some more changes, not sure where they are from

packages/frontend/src/vue-apollo.js
```diff
@@ -2,117 +2,116 @@
...
+    // Use websockets for everything (no HTTP)
+    // You need to pass a `wsEndpoint` for this to work
+    websocketsOnly: false,
+    // Is being rendered on the server?
+    ssr: false,
+    // Subscription Client
+    networkInterface: createSubscriptionClient(),
+    // Extra cache settings
+    inMemoryCacheOptions: {
+      cacheRedirects: {
```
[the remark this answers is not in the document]
gjedlicska: we should do some tests, re my PR on the search users query cache issue. I do remember having some nasty caching issues, but Dim know what they were

packages/frontend/src/plugins/authHelpers.js
```diff
@@ -103,13 +102,13 @@ export async function signOut(mixpanelInstance) {
   window.location = '/'
 
   if (mixpanelInstance) {
-    mixpanelInstance.track('Log Out', { type: 'action'  })
+    mixpanelInstance.track('Log Out', { type: 'action' })
```
gjedlicska, line 105: we should start a mixpanel constants like you did for the local storage. I like that.

packages/server/.vscode/settings.json
```diff
@@ -1,3 +1,7 @@
 {
-  "mochaExplorer.env": { "NODE_ENV": "test" }
+  "mochaExplorer.env": { "NODE_ENV": "test" },
+  "javascript.suggest.autoImports": true,
```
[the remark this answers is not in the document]
fabis94: That shouldn't be the case, it should just prefer importing with the `@/` prefix with absolute paths. Haven't had any issues with this in the past, but of course if it breaks or doesn't work we can switch back.

fabis94:
> why did u mark the PR draft?

@gjedlicska no special reason, just didn't want anyone to merge it in until the review is done :)

fabis94:
> > Vue linter asks for multi-word component names
> 
> we prob should disable that, or its going to be a massive rename show

@gjedlicska I think it's a good idea to follow the Vue component naming guidelines. It's way easier to understand the relation & hierarchy of components when they're named accordingly, because you can just look at file names. Currently you actually have to read each component to get the kind of metadata that good naming provides.

packages/frontend/src/plugins/authHelpers.js
fabis94: yea, good idea

packages/viewer/example/example.html
fabis94: honestly, i don't know what happened here, I just reverted it back to the main version

cristi8 (approved):
Looked at the server mostly. Hard to review in detail with all the formatting changes, but got the main points, looking good

packages/server/modules/core/loaders.js
```diff
@@ -0,0 +1,58 @@
...
+   */
+  buildRequestLoaders(ctx) {
+    const userId = ctx.userId
+
+    return {
+      streams: {
+        /**
+         * Get favorite metadata for a specific stream and user
+         */
+        getUserFavoriteData: new DataLoader(async (keys) => {
```
gjedlicska, line 28: this is neat :)
how can u run tests on this withour relying on the graphql server?
U'd have to extract the StreamsFavorites by streamId query part into a separate function, and test that right?

packages/server/modules/core/migrations/20220318121405_add_stream_favorites.js
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async function (knex) {
+  await knex.schema.createTable(TABLE_NAME, (table) => {
+    table.string('streamId', 10).references('id').inTable('streams').onDelete('cascade')
+    table.string('userId', 10).references('id').inTable('users').onDelete('cascade')
+    table.timestamp('createdAt').defaultTo(knex.fn.now())
```
gjedlicska, line 13: pls switch to the same specific type implementation, that we use in comments (20220222173000_comments.js) for datetime types.

It solves the javascript vs postgres date storage precisison incompatibility

packages/server/modules/core/tests/graphSubs.spec.js
```diff
@@ -1,593 +1,742 @@
...
 
   // set up app & two basic users to ping pong permissions around
-  before( async function ( ) {
-    this.timeout( 10000 ) // we need to wait for the server to start in the child process!
+  before(async function () {
+    this.timeout(10000) // we need to wait for the server to start in the child process!
 
-    await knex.migrate.rollback( )
-    await knex.migrate.latest( )
+    await beforeEachContext()
```
fabis94, line 52: Main thing I did here - replaced rollback/latest w/ truncate, cause these were really slow (on top of being E2E tests)

packages/server/modules/core/tests/favoriteStreams.spec.js
```diff
@@ -0,0 +1,294 @@
+/* instanbul ignore file */
```
fabis94, line 1: Event though tests for this functionality could've been put inside `streams.spec.js`, I put everything in a separate file for 2 reasons:
1) Tests in streams.spec.js aren't very deterministic, because there are tests depending on the results of other tests. This makes them flaky depending on the order they're ran. So I wanted to create a new one with the test state only being prepared in `before` calls and tested in a way that its not not mutated in a serious way accross tests.
2) I wanted to organize tests in such a way that there's more re-use of test states set up in `before` calls in nested `describe` blocks, so that there's less duplication of test state/data creation code. Jest suggests devs to write the titles of `describe` and `it` calls in such a way that they make a grammatically correct sentence at the end, because organizing `describe`/`it` blocks in this way is more conducive to well organized & nested tests and it's easier to understand what's going on that way. So instead of a test being named something like 'Objects > Repository > error when object deleted' they would end up being named 'Favorite streams > when authenticated > and favorited > can be unfavorited'.

packages/server/modules/core/tests/favoriteStreams.spec.js
```diff
@@ -0,0 +1,294 @@
...
+/*
+ * TODO: Extra tests
+ * - Test for new error handler, checking if exceptions are reported correctly in responses
+ */
+
+/**
+ * Cleaning up relevant tables
+ */
+async function cleanup() {
+  await truncateTables([StreamFavorites.name, Streams.name, Users.name])
```
fabis94, line 22: Quicker than beforeEachContext()

packages/server/modules/core/tests/favoriteStreams.spec.js
```diff
@@ -0,0 +1,294 @@
...
+    name: 'Some Other DUde',
+    email: 'otherguy@gmail.com',
+    password: 'sn3aky-1337-b1m'
+  }
+
+  before(async function () {
+    await cleanup()
+
+    // Seeding
+    await Promise.all([
```
fabis94, line 82: Executing promises in parallel for speed increase

packages/server/modules/core/tests/favoriteStreams.spec.js
```diff
@@ -0,0 +1,294 @@
...
+  after(async () => {
+    await cleanup()
+  })
+
+  describe('when authenticated', () => {
+    /** @type {import('apollo-server-express').ApolloServer} */
+    let apollo
+
+    const favoriteStream = async (sid, favorited) =>
+      await apollo.executeOperation({
```
fabis94, line 106: This is the way Apollo suggests writing integration tests
https://www.apollographql.com/docs/apollo-server/testing/testing/

packages/server/modules/core/tests/favoriteStreams.spec.js
```diff
@@ -0,0 +1,294 @@
...
+      await StreamFavorites.knex().truncate()
+    })
+
+    const accessibleStreamIds = [
+      [() => myPubStream.id, 'owned and public'],
+      [() => myStream.id, 'owned and not public'],
+      [() => notMyPubStream.id, 'not owened, but public']
+    ]
+
+    accessibleStreamIds.forEach(([id, msgSuffix]) => {
```
fabis94, line 133: apparently this is how dataProviders are implemented in mocha

packages/server/app.js
```diff
@@ -20,44 +20,25 @@ const { ApolloServer, ForbiddenError } = require('apollo-server-express')
...
+ * Create Apollo Server instance
+ * @param {Partial<import('apollo-server-express').ApolloServerExpressConfig>} optionOverrides Optionally override ctor options
+ * @returns {import('apollo-server-express').ApolloServer}
  */
-exports.init = async () => {
-  const app = express()
-
-  Logging(app)
-  MatStartup()
+exports.buildApolloServer = (optionOverrides) => {
```
fabis94, line 33: Moved out ApolloServer creation to a function so that tests can create their own instances

packages/server/modules/core/errors/base.js
```diff
@@ -0,0 +1,80 @@
+const VError = require('verror')
+
+/**
+ * Base application error (don't use directly, treat it as abstract). Built on top of `verror` so that you can
+ * chain errors (e.cause is the previous error) and also add arbitrary metadata using the `info` option.
+ *
+ * This allows for much nicer error handling & monitoring
+ */
+class BaseError extends VError {
```
fabis94, line 9: Added VError as an abstract base error class (adjusted apollo server response error formatting to output these kinds of errors as well).

VError is very nice, because it allows you to add specify a `cause` Error in the Error constructor. So you can chain errors this way (e.g. when you catch an error and then throw a new one) and get full stack traces for all errors in the chain and a concatenated error message - very useful for debugging!

Additionally you can add any kind of arbitrary metadata when invoking the constructor through the `info` options property. Monitoring tools like Sentry rely on grouping errors by their name, and with this feature you no longer have to build dynamic error names w/ relevant metadata (e.g. UserID), you can just add it to the metadata that can easily be sent to Sentry as well.

```js
// The message would be: "Couldn't create user: Lost connection to database"
try {
  await createUser(userInfo);
} catch (e) {
  throw new VError("Couldn't create user, {
     cause: e,
     info: {userId: userInfo.id}
  }
})
}
```

packages/server/modules/core/repositories/streams.js
```diff
@@ -0,0 +1,216 @@
+const _ = require('lodash')
+const { Streams, StreamAcl, StreamFavorites, knex } = require('@/modules/core/dbSchema')
+const { InvalidArgumentError } = require('@/modules/core/errors/base')
+
+/**
```
fabis94, line 5: Split all data access code to the repository layer for a better separation of concerns. Similarly moved all logic away from the resolvers to the service layer.

So layers are utilized like this: Resolver > Service > (DataLoader) > Repository

packages/server/nyc.config.js
```diff
@@ -0,0 +1,20 @@
+const testFileExtensions = ['ts', 'js']
+
+module.exports = {
+  exclude: [
+    `**/migrations/*.{${testFileExtensions}}`,
```
fabis94, line 5: Removed migrations from coverage calculations permanently

packages/server/modules/core/tests/favoriteStreams.spec.js
gjedlicska: ok, i can get behind this. Even if the syntaxt is a bit weird to me atleast.

packages/server/test/hooks.js
```diff
@@ -53,21 +55,22 @@ const initializeTestServer = async (app) => {
 
 exports.mochaHooks = {
   beforeAll: async () => {
+    console.log('running before all')
     await unlock()
     await knex.migrate.rollback()
     await knex.migrate.latest()
-    console.log('running before all')
+    await init()
```
gjedlicska, line 65: why do we need to initialize here, if we run the init in the beforeEachContext?

gjedlicska (approved):

packages/server/modules/core/tests/favoriteStreams.spec.js
fabis94: yea in jest there are built-in helpers for this, like `it.each(data)(message, (a,b,c) => {..});`, but in mocha there's nothing like that and you just use a basic array forEach for this

packages/server/test/hooks.js
fabis94: IIRC it's because I don't use beforeNewContext() in my test module, because of performance issues - manually truncating a couple of relevant tables works quicker for me.

PR #649, opened 2022-03-29
Matteo/onboarding

teocomi:
Improves the onboarding wizard.


packages/frontend/src/main/components/onboarding/GettingStartedSteps.vue
```diff
@@ -14,242 +14,162 @@
...
-                        Software engineer, computational designer, hacker
-                      </v-card-subtitle>
-                    </v-card>
-                  </v-hover>
-                </v-col>
-              </v-row>
-            </v-container>
-            <p class="my-7"><strong>Let us give you a quick tour! 🙋‍♂️</strong></p>
+          <v-card class="transparent elevation-0 text-center" color="transparent" align="start">
+            <v-img class="align-start mb-3" src="@/assets/onboarding-1.png" :aspect-ratio="16 / 9">
```
fabis94, line 18: Would be nice to optimize these images further:
1. Compress as much as possible (maybe this is already done)
2. Generate a .webp version of these images and fallback to PNG if browsers don't support them https://usefulangle.com/post/114/webp-image-in-html-with-fallback

(In the future we should make this process automatic so that webpack handles it on build)

fabis94 (approved):

PR #650, opened 2022-03-29
gergo/CIRewrite

gjedlicska:
- initial ci config restructure
- disable main branch trigger
- its YML not YAML
- update
- split workflow
- move filters to workflow def
- unify workflow
- install packages for test
- changing directory
- add server appdir for tests
- packages!
- add trying to add test runner
- appdir test runner
- add test reporting
- remove node/test entry
- add test parallelism
- its not worth it to do parallelism yet...
- store coverage lcov
- build viewer
- build-viewer
- install objectloader localy
- run the actual build
- add server build
- add get version
- get-version job
- python image version
- build script update
- disable global version get script
- fix filter ref
- fix docker image tag for build
- add build for all containers
- fix job names
- fix job names
- fix job names
- add helm publish
- add npm publish
- rename build step
- add docker hub context to server build
- use generalized build job
- remove ci context from job def
- yaml anchor base build
- add base build anchoring
- test multiple <<-s
- fix context anchor
- fix anchors
- comment cleanup
- checkout for helm publish too...
- publish to docker hub
- ls images docker
- add docker image tag filtering
- re-enable tests in CI
- final polish


.circleci/build.sh
```diff
@@ -6,18 +6,17 @@ set -e
...
 
-echo "$DOCKER_REG_PASS" | docker login -u "$DOCKER_REG_USER" --password-stdin $DOCKER_REG_URL
-docker push $DOCKER_IMAGE_TAG:latest
-docker push $DOCKER_IMAGE_TAG:$IMAGE_VERSION_TAG
+docker tag $DOCKER_IMAGE_TAG:latest $DOCKER_IMAGE_TAG:$IMAGE_VERSION_TAG
 
 if [[ "$IMAGE_VERSION_TAG" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-  docker tag $DOCKER_IMAGE_TAG:latest $DOCKER_IMAGE_TAG:2
-  docker push $DOCKER_IMAGE_TAG:2
+  docker tag $DOCKER_IMAGE_TAG:2
```
cristi8, line 18: missing parameter to docker tag

cristi8 (changes requested):
looking good, except the missing `docker tag` parameter when tagging `:2`

.circleci/build.sh
gjedlicska: its actually tagget on line 15. if we're pushing an alpha tag or a release tag, its value is stored on the $IMAGE_VERSION_TAG and applied above. The only tag, that we're adding only when we're releasing is the `:2` tag.

Below with the `docker push -a` pushes all tags that is attached to the given image

gjedlicska: fixed it

PR #653, opened 2022-03-31
eslintAllTheThings

gjedlicska:
- feat: added prettierignore file and fixed some issues prettier found in some files
- added script for fixing all formatting issues
- some extra configuration
- chore: reformatting everything with prettier
- style(all): reformat with print width 88
- style(webhook-service): fix eslint errors on webhooks
- style(objectloader): fix eslint errors
- style(viewer): fix eslint errors
- style(fileimport-service): fix eslint errors
- style(previews): fix most of eslint errors
- style(server): fix all eslint errors and break the tests


packages/frontend/.eslintrc.js
```diff
@@ -27,10 +27,7 @@ const config = {
       }
     }
   ],
-  plugins: ['vue'],
-  rules: {
-    'no-console': 1
```
fabis94, line 32: I don't see a point to this rule if there's no other way to gain insight into the application. Sure, outputting debugging information where everyone can see it isn't ideal, but not getting _any_ information about what's going on is even worse. Additionally, I don't really see the big issue with dumping debugging info in the console as long as 1) there's not too much of it 2) it doesn't leak info that we'd like to keep private.

Another way to resolve this would be to add monitoring (sentry?) to the frontend as well and then send debugging info to sentry directly (or some other target).

packages/frontend/src/embed/EmbedApp.vue
```diff
@@ -7,6 +7,8 @@ export default {
   mounted() {
     let mixpanelId = this.$mixpanelId()
     this.$mixpanel.register({
+      // Unfortunately we can't replace this to camelCase, because that will break metrics
+      // eslint-disable-next-line camelcase
```
fabis94, line 11: Not great as we can't really change the property. We could create a helper/mixin that wraps this call, however, so that there's only 1 line that needs to be ignored.

packages/frontend/src/main/components/activity/ListItemActivity.vue
```diff
@@ -540,6 +539,10 @@ export default {
     }
   },
   methods: {
+    /**
+     * TODO: no bueno, needs refactoring. As eslint warns us - using v-html can lead to XSS attacks.
+     * Additionally I don't see why this can't be done with Vue templates
```
fabis94, line 544: I'm guessing that the description of streams & branches can be freely edited by the user, right? In that case this is a serious security issue

packages/frontend/src/main/components/viewer/CommentAddOverlay.vue
```diff
@@ -149,6 +149,8 @@
   </div>
 </template>
 <script>
+// TODO: Need to fix the viewer package build process to be able to properly reference THREE.js
+/* global THREE */
```
fabis94, line 153: Read this for more context: https://discord.com/channels/726756379083145269/743849923803349022/959432401275215872

packages/frontend/src/main/dialogs/ShareStreamDialog.vue
```diff
@@ -103,12 +103,13 @@
         </v-toolbar-title>
         <v-spacer></v-spacer>
         <v-switch
-          v-model="stream.isPublic"
```
fabis94, line 106: Mutating props is a big no no in Vue, it can lead to hard to debug bugs. The correct approach is to emit changes out with an event or in this case since the data comes from Apollo cache - to update the apollo cache directly

packages/frontend/src/main/navigation/StreamNav.vue
```diff
@@ -319,6 +324,7 @@ export default {
       return this.branchQuery.branches.items.filter((b) => b.name !== 'globals').length
     },
     parsedDescription() {
+      // TODO: Pretty sketchy, isn't the descripton user editable? A user could inject XSS this way
```
fabis94, line 327: Another opening for an XSS attack

workspace.code-workspace
```diff
@@ -43,7 +43,6 @@
     "editor.formatOnPaste": true,
     "editor.multiCursorModifier": "ctrlCmd",
     "editor.snippetSuggestions": "top",
-    "eslint.format.enable": true,
```
fabis94, line 46: maybe this resolves Cristi's issue with multiple formatters running

PR #655, opened 2022-04-04
fix(frontend): removed all v-html usages in the frontend protecting us against XSS

fabis94:


packages/frontend/src/main/pages/admin/ServerSettings.vue
```diff
@@ -106,7 +107,7 @@ export default {
           }
         `,
         variables: {
-          info: this.serverModifications
+          info: pick(this.serverModifications, Object.keys(this.serverDetails))
```
fabis94, line 110: Fixed a bug that had broken the server settings form so that you couldn't submit it. It sends too many fields to the backend essentially.

packages/frontend/src/main/pages/user/ProfileSelf.vue
```diff
@@ -1,15 +1,6 @@
 <template>
   <v-container class="pa-0" fluid>
     <portal to="toolbar"><b>Your Profile</b></portal>
-    <portal to="subnav-profile">
```
fabis94, line 4: Removed confusing "sign out" menu item

PR #677, opened 2022-04-07
Fixing viewer & objectloader build config

fabis94:
- Fixing viewer build config so that it builds a proper library build instead of a web application build with all of the dependencies bundled in
- Adding build & transpilation step to objectloader, because the source code syntax isn't going to work without it in various environments
- Fixed examples in viewer and objectloader packages so that they actually work as expected without CORS issues, also updated readmes to be more clear about how to use and test these packages


packages/frontend/.browserslistrc
```diff
@@ -1,3 +1 @@
-> 1%
-last 2 versions
-not dead
+since 2019
```
fabis94, line 1: After discussion w/ dim, moving all packages to rely on a "since 2019" browserslist query

packages/frontend/Dockerfile
```diff
@@ -3,10 +3,14 @@
 # build stage
 FROM node:16.13-bullseye-slim as build-stage
 
+WORKDIR /opt/objectloader
```
fabis94, line 6: Are the changes in both dockerfiles correct? Also, are these really the only ones that need updating? The root Dockerfile also seems to install the viewer (and objectloader by proxy)

gjedlicska (approved):

PR #679, opened 2022-04-08
feat(server): adds `users:invite` scope to manager and connector defa…

teocomi:
Adds `users:invite` scope to manager and connector default apps

### DIFF
--- a/packages/server/modules/auth/defaultApps.js
+++ b/packages/server/modules/auth/defaultApps.js
@@ -103,7 +103,8 @@ const SpeckleDesktopApp = {
     'streams:write',
     'profile:read',
     'profile:email',
-    'users:read'
+    'users:read',
+    'users:invite'
   ]
 }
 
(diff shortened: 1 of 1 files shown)


teocomi:
Blocked by https://github.com/specklesystems/speckle-server/issues/680

packages/server/modules/auth/defaultApps.js
```diff
@@ -103,7 +103,8 @@ const SpeckleDesktopApp = {
     'streams:write',
     'profile:read',
     'profile:email',
-    'users:read'
+    'users:read',
+    'users:invite'
```
fabis94, line 107: Would be great if you'd add the new scopes to `/packages/server/modules/core/helpers/mainConstants.js` and use those constants here, instead of duplicating the strings

gjedlicska:
superseeded by #718

Issue #689, opened 2022-04-12
Introduce proper workspaces support to get around duplicated dependencies in builds


fabis94:
**What package are you referring to?**
This relates to the entire speckle-server repo. 

**Describe the bug**
Basically, when we do `npx lerna bootstrap` to symlink local packages as deps into eachother's node_modules, there's no de-duplication happening afterwards, that would clean up duplicate dependencies in symlinked package `node_modules` folders.

Because of this if we're not careful, it's easy to bundle duplicates of any dependencies that are dependencies of symlinked packages. So core-js, three.js - both of these libs are going to be bundled twice if you aren't careful.

**To Reproduce**
Add `import * as THREE from 'three'` anywhere, and you should get a warning in console about multiple three.js libs being loaded. This is because one version comes from 'viewer/node_modules/' and the other from 'frontend/node_modules'.

**Expected behavior**
After linking in packages, there should be appropriate de-duplication of dependencies. I've tried various things like upgrading webpack to 5 or using lerna boostrap with the --hoist flag, but none of these resolve the issue. The real solution is to move to a proper & configurable workspaces/monorepo solution like node 16+ workspaces or better yet `yarn` workspaces, because those are more configurable.

**Note: ** This doesn't affect speckle-server in production, because our CI/CD doesn't rely on symlinking to install local packages and the way it does the building & installing makes the deps be de-duplicated correctly.

gjedlicska:
closed by #753

fabis94:
awesome!

Issue #658, opened 2022-04-04
Investigate webpack build issues for frontend


fabis94:
- Attempting to profile build
- Looking into how code splitting is configured
- ESBuild? Parallel webpack?
- Investigate current webpack config to see if parallelization & caching is utilized the way we want it to be utilized

Issue #340, opened 2021-07-16
Support queries on the get object REST endpoints.


didimitrie:

didimitrie:
@AlanRynne + @clairekuang: would solve powerbi integrations problems too.

teocomi:
For reference, currently blocking: https://github.com/specklesystems/speckle-powerbi/issues/2

PR #690, opened 2022-04-12
feat(frontend): build speed & chunk structure optimization

fabis94:
closes [#340](https://github.com/specklesystems/speckle-server/issues/658)

Improvements & comments:
- Switched to using esbuild-loader instead of babel-loader for local/dev builds since that will speed things up a little bit. This is kind of an experiment, so if there are any issues we can disable this fairly easily.
- Added possibility to `serve` the built `frontend` build (from /dist) locally using `npm run serve`. This wasn't possible before.
- Added a `--profile` flag (& relevant scripts) to build commands that will output useful info about the build process and also open up the WebpackBundleAnalyzer tree-view in the browser. This is useful for analyzing how optimized our bundles & build processes are.
- Optimized chunks with the help of WebpackBundleAnalyzer. Asynchronized ApexCharts, VueHistogram & removed all `crypto` related dependencies, and this has greatly reduced the amount of JS on all pages. For example, on the login page the amount of requested JS has been reduced from 2.3MB (pre-compression) to 700kb (pre-compression), so a nearly 70% reduction in size.
- Discovered issue where webpack bundles in duplicate dependencies in local builds in some cases. [The solution isn't obvious](https://github.com/specklesystems/speckle-server/issues/689), so as a temporary measure I added a webpack plugin that returns duplicate dependencies on builds.
- Fixed spacing issue on login page that caused unnecessary spaces in the main text there. Basically prettier broke out a couple of `<span>` elements into multiple lines, which causes those extra whitespaces. By adding instructing comments I configured prettier to not break those `<span>` elements into multiple lines.

image

- Fixed incorrect `engines` field in package.jsons, the way we set it up didn't allow for node16. It's not really strictly followed, only generates a warning, so this isn't a big issue right now, but I've fixed it anyway.


packages/frontend/src/helpers/md5.js
```diff
@@ -0,0 +1,179 @@
+/**
```
didimitrie, line 1: Tbh, the only check i'd want you to make is to see if we get the same hash as before. If we do, all is well and this can be merged :)

didimitrie: perfect, then imho 👍

PR #701, opened 2022-04-15
Fix various typos

luzpaz:
Found via `codespell -q 3 -S CHANGELOG.md -L doesnt,optio,promisses,tread,usera`

### DIFF
--- a/packages/viewer/src/modules/SceneObjects.js
+++ b/packages/viewer/src/modules/SceneObjects.js
@@ -295,7 +295,7 @@ export default class SceneObjects {
 
         let materialId = `${m.type}/${m.vertexColors}/${m.color.toJSON()}/${m.side}/${
           m.transparent
-        }/${m.opactiy}/${m.emissive}/${m.metalness}/${m.roughness}/${m.wireframe}`
+        }/${m.opacity}/${m.emissive}/${m.metalness}/${m.roughness}/${m.wireframe}`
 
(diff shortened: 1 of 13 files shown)


packages/viewer/src/modules/SceneObjects.js
```diff
@@ -295,7 +295,7 @@ export default class SceneObjects {
 
         let materialId = `${m.type}/${m.vertexColors}/${m.color.toJSON()}/${m.side}/${
           m.transparent
-        }/${m.opactiy}/${m.emissive}/${m.metalness}/${m.roughness}/${m.wireframe}`
+        }/${m.opacity}/${m.emissive}/${m.metalness}/${m.roughness}/${m.wireframe}`
```
luzpaz, line 298: Please review closely

didimitrie: It's a fun catch. I think it should be fine as these ids are transient/local session only relevant, but I'll ping @cristi8 to double check as it was his brain at work here :)

cristi8: oh, yes, nice catch. this is actually a bug.. we would now group meshes with materials that only differ in opacity value, but have the same other attributes -> resulting in wrong materials for some objects in the final rendered scene

cristi8 (approved):

didimitrie:
thanks @luzpaz, merged!

PR #702, opened 2022-04-17
Hotfix globals

didimitrie:
As reported on the forum: https://speckle.community/t/globals-not-appearing-in-web-ui/2746/4


packages/frontend/src/main/components/stream/globals/GlobalsBuilder.vue
```diff
@@ -119,7 +119,7 @@ export default {
         return data.stream.object
       },
       skip() {
-        return !!this.objectId
```
didimitrie, line 122: cc @gjedlicska @fabis94: somehow linting made a `!` into a `!!` :D

Issue #680, opened 2022-04-08
Pre-registered apps scope changes


teocomi:
Currently, any scope change to a registered app would invalidate all existing tokens, for default pre-registered apps this is undesirable especially as it would affect Manager and Connectors.

Therefore we should allow for scope changes to default apps without requiring the user to re-authorize.
Should be tacked together with https://github.com/specklesystems/speckle-server/issues/227

Issue #227, opened 2021-05-08
Speckle App should not ask for access every time


teocomi:
Speckle Apps should only ask for access the first time they are used, afterwards should automatically redirect if the user has not revoked them.

![image](https://user-images.githubusercontent.com/2679513/117536211-a713bd80-aff1-11eb-98bc-ab53e24c4c63.png)

PR #718, opened 2022-04-25
defaultAppScopeHandling

gjedlicska:
- feat(server default apps): remove revoke default app credentials after updating it
- test(server apps tests): test default app permissions not being revoked
- refactor(scopes constants): add users:invite to Scopes constant
- feat(default apps): update manager and connectors with user invite scopes


packages/server/modules/auth/tests/apps.spec.js
```diff
@@ -214,6 +216,61 @@ describe('Services @apps-services', () => {
       .catch((err) => expect(err.message).to.equal('Access code not found.'))
   })
 
+  const defaultApps = ['spklwebapp', 'explorer', 'sdm', 'sca', 'spklexcel']
+  defaultApps.forEach((speckleAppId) => {
+    it(`Should not invalidate tokens, refresh tokens and access codes for default app: ${speckleAppId}, if updated`, async () => {
+      const { updateDefaultApp } = require('@/modules/auth/defaultApps')
```
fabis94, line 222: Better to keep all require calls near the top for clarity

fabis94 (approved):

packages/server/modules/auth/tests/apps.spec.js
gjedlicska: this one is a deliberate placement, the updateDefaultApp function used to be private. Since its operation on the default server apps it works with sensitive stuff, I would preferred to keep it private.

Do you have better options to testing not exported functions?

packages/server/modules/auth/defaultApps.js
```diff
@@ -1,11 +1,11 @@
...
-const appRoot = require('app-root-path')
-const knex = require(`${appRoot}/db/knex`)
+const knex = require('@/db/knex')
 const Scopes = () => knex('scopes')
 const Apps = () => knex('server_apps')
 const AppScopes = () => knex('server_apps_scopes')
 
-const { getApp, revokeExistingAppCredentials } = require('./services/apps')
+const { getApp } = require('@/modules/auth/services/apps')
+const { Scopes: ScopesConst } = require('@/modules/core/helpers/mainConstants')
```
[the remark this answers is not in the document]
gjedlicska: AllScopes is another exported member of the mainConstants module :D

packages/server/modules/auth/tests/apps.spec.js
fabis94: The placement doesn't fully resolve the encapsulation issue, because devs can still import the newly public function anywhere. I get what you meant, you put the `require` call in a more hidden location so that devs don't see it, but if they just open up `defaultApps` directly they'll find the export there.

In this case you aren't testing the private function (which would be a code smell as well, since module/class internals should just be implementation details that you shouldn't need to directly test/validate), you're using a previously private function to prepare the test data state. Considering that you need it outside of the private module context, it can no longer be a private/non-exported function and it makes sense that you've exported it. It doesn't matter whether the new requirement comes from a test file or some business-oriented service - if the function needs to be public, it becomes public.

So the correct thing to do is still to move the `require` call to the top of the test file, the `updateDefaultApp` is a new public member of the `defaultApps` module.

I do get your concern regarding the semantics however - we want `updateDefaultApp` to be an internal function, not something to be used directly anywhere. For that problem proper organization of code and further separation of concerns is a good solution. You'll always have seemingly internal functions that have to be public/exported, just so you can invoke them from other seemingly internal areas of the app. And since that is the case, the only thing you can really do is organize internal things separately from public things. So, for example, instead of having default app definitions, service functions and module init code all defined in the same file (defaultApps.js), the service functions could be split out to a `defaultAppService` as exported members, and it should be somewhat clear to devs that this deals with the setup of default apps and is an "internal service".

packages/server/modules/auth/defaultApps.js
```diff
@@ -57,7 +57,6 @@ async function updateDefaultApp(app, existingApp) {
   )
 
   if (scopeDiffA.length !== 0 || scopeDiffB.length !== 0) {
-    await revokeExistingAppCredentials({ appId: app.id })
```
didimitrie, line 60: Okay - sorry for being late on this; we can do tomorrow am a meeting!  

This actually indeed prevents token revoking of default apps, but it leaves us in an invalid state: existing tokens will only have old scopes attached to them - when the app will try to use flex its new permissions, it won't actually be able to as its token does not hold the relevant scope (reverse is a problem too, token remains more powerful than it should be if we remove scopes). 

There's an extra tedious step in which we need to update all tokens that have been granted for this specific app to reflect the new scopes. 

`user_server_app_tokens` holds the info on what app what tokens, and `token_scopes` holds the token <> scopes relationship. Off the top of my head, I'd split this in two cases, addition and removal, and do some batch inserts/deletions if possible (or within one transaction) to prevent db chugging.

didimitrie (approved):

didimitrie:
TL;DR post convo with @gjedlicska: 
- This PR closes #680 
- Does not close #227, this will be treated in a separate PR

Issue #723, opened 2022-04-26
Viewer: parse `displayValue` for objects with no direct convertor


AlexandruPopovici:
An example for this is `spirals` from this stream https://speckle.xyz/streams/99abc74dd4/commits/b32fdcf171

~Related to (if not duplicates) https://github.com/specklesystems/speckle-server/issues/708~

Issue #711, opened 2022-04-21
Investigate why some curves are badly constructed inside the viewer


AlexandruPopovici:
Here is an example: https://latest.speckle.dev/streams/0c6ad366c4/commits/912d83412e

AlexandruPopovici:
There was a problem with how arcs are constructed inside the viewer. Handled this issue in 
- https://github.com/specklesystems/speckle-server/commit/4164e12d56eacfa14dda9d37047e6c8dba864e79
- https://github.com/specklesystems/speckle-server/commit/f7ff0c6ebf1d7176d7d76936741eb7553a92491d

Would like some more arcs to test with tho

Issue #708, opened 2022-04-21
Display `displayValue` from random objects


didimitrie:
If any object has a display value, show it.

Issue #282, opened 2021-06-08
Viewer: Confusing handling of root commit object property `data`


izzylys:
### Expected vs. Actual Behavior

It seems that a property called `data` on the root commit object causes some unexpected behaviour for streams not coming from Grasshopper.

In @AlanRynne's ifc test stream, none of the objects within a single base under the `data` field on the root commit showed up. Making this a list of the one base rather than just the base itself caused the objects to show up (eg `data: [obj]` rather than `data: obj`).

In this blender stream, only the object within the root `data` field show up - other objects on the root are skipped.

- [here](https://latest.speckle.dev/streams/0c6ad366c4/commits/c52cfd5a54) is the borked commit
- [here](https://latest.speckle.dev/streams/0c6ad366c4/commits/8141f58152) is a fixed commit (changed `data` field to `objects`)

didimitrie:
@izzylys, is this still happening? I haven't heard anything re it in a while, but this doesn't mean it's not active still 😬

izzylys:
yep we never fixed this! as mentioned in the issue, i got around it in Blender by changing the `data` field to be called `objects` so it's not a problem for blender. but it is still a problem that there is some special handling of the `data` field on a root commit obj

izzylys:
bumping this just bc i got stuck again debugging something only to realise it was this issue 🤡

so here's another one - geo ignored bc it is within a `data` prop

https://latest.speckle.dev/streams/73539e248d/objects/2e2ae95239035884b8d612030f598918

PR #733, opened 2022-04-28
Alex/#723#711

AlexandruPopovici:
- closes https://github.com/specklesystems/speckle-server/issues/723 -> Example stream https://speckle.xyz/streams/99abc74dd4/commits/b32fdcf171 
- closes https://github.com/specklesystems/speckle-server/issues/711 -> Example streams https://latest.speckle.dev/streams/0c6ad366c4/commits/912d83412e ; https://speckle.xyz/streams/99abc74dd4/commits/b32fdcf171
- closes https://github.com/specklesystems/speckle-server/issues/708 -. Example stream https://latest.speckle.dev/streams/0c6ad366c4/commits/462dabba78 (this one is a bit strange because there are two display values with unique ids inside the hierarchy, so they are both converted. But since they have identical geometry, they double up and cause z-fighting. Not sure if this is indented behaviour)
- closes https://github.com/specklesystems/speckle-server/issues/282 too :)


didimitrie:
> this one is a bit strange because there are two display values with unique ids inside the hierarchy, so they are both converted. But since they have identical geometry, they double up and cause z-fighting. Not sure if this is indented behaviour)

In that specific case, it is intended behaviour. @tsvilans / @izzylys doubled up some data without needing to (ie the brep value already contains a mesh, but we also have the mesh as direct display value. It might be that the data structure is a direct result of how we were (not) handling things so far. 

We could return from the conversion if we encounter a `displayValue`, but this opens up another can of worms.

packages/viewer/src/modules/converter/Converter.js
```diff
@@ -186,7 +202,19 @@ export default class Coverter {
       const type = this.getSpeckleType(obj)
       if (this[`${type}ToBufferGeometry`]) {
         return await this[`${type}ToBufferGeometry`](obj.data || obj, scale)
-      } else return null
+      }
+      /**
+       * Regarding #723. This would be more generic and possibly handle other
+       * types with missing direct convertor, however I don't feel it is the
+       * 'convert' fuction's place to handle this...
+       */
```
didimitrie, line 210: Agreed. The whole traversal & conversion logic needs to be cleaned up massively anyway...

packages/viewer/src/modules/converter/Converter.js
```diff
@@ -620,6 +725,9 @@ export default class Coverter {
       this.PointToVector3(plane.normal).normalize()
     )
     m.setPosition(this.PointToVector3(plane.origin))
+    /**
+     * I think scaling should be done first.
+     */
```
didimitrie, line 730: You know you're the boss, so next pass over this one, feel free :)

didimitrie (approved):

packages/viewer/src/modules/converter/Converter.js
```diff
@@ -173,6 +176,19 @@ export default class Coverter {
     this.activePromises -= childrenConversionPromisses.length
   }
 
+  directConverterExists(obj) {
+    return this[`${this.getSpeckleType(obj)}ToBufferGeometry`] !== undefined
+  }
+
+  getDisplayValue(obj) {
```
clairekuang, line 183: So `displayMesh` was our really old version of fallback geometry values: I think the viewer ought to render only what's inside `displayValue` if it exists, and only to check for `displayMesh` if not (to maintain backwards compatibility)

PR #734, opened 2022-04-29
fix: throwing error if running tests without .env.test

### LINKED ISSUE #731: Running `npm run test` will wipe the local DB
We shouldn't risk wiping the database when testing env is not correctly configured.

In my case, i have `.env` file, but not configured the `.env.test` file

CC @fabis94 : i am thinking either commit the `.env.test` in the repo, or fail the test if there is no such file. But running tests on the DB configured in .env should never happen. What do you think?

fabis94:
closes #731

### DIFF
--- a/packages/server/bootstrap.js
+++ b/packages/server/bootstrap.js
@@ -14,7 +14,14 @@ const appRoot = require('app-root-path')
 // If running in test env, load .env.test first
 // (appRoot necessary, cause env files aren't loaded through require() calls)
 if (isTestEnv()) {
-  dotenv.config({ path: `${appRoot}/.env.test` })
+  const { error } = dotenv.config({ path: `${appRoot}/.env.test` })
+  if (error) {
+    const e = new Error(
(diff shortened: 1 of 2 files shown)


packages/server/bootstrap.js
```diff
@@ -14,7 +14,14 @@ const appRoot = require('app-root-path')
...
 // (appRoot necessary, cause env files aren't loaded through require() calls)
 if (isTestEnv()) {
-  dotenv.config({ path: `${appRoot}/.env.test` })
+  const { error } = dotenv.config({ path: `${appRoot}/.env.test` })
+  if (error) {
+    const e = new Error(
+      'Attempting to run tests without an .env.test file properly set up! Check readme!'
+    )
+    console.error(e)
+    process.exit(1)
```
fabis94, line 23: Exiting directly not throwing, so that the error doesn't get caught by yargs and reported as something else (happened to me with an old node_modules)

cristi8 (approved):

gjedlicska (approved):

PR #739, opened 2022-05-03
Server no app-root-path

gjedlicska:


packages/server/bootstrap.js
```diff
@@ -5,11 +5,11 @@
...
 require('module-alias')({ base: __dirname })
+const appRoot = __dirname
 
 // Initializing env vars
 const dotenv = require('dotenv')
 const { isTestEnv } = require('./modules/core/helpers/envHelper')
-const appRoot = require('app-root-path')
 
 // If running in test env, load .env.test first
 // (appRoot necessary, cause env files aren't loaded through require() calls)
```
fabis94, line 15: Maybe we can get rid of appRoot here and just use `require.resolve` to resolve the path to the file, like you've done elsewhere 🤔

packages/server/scripts/duplicateUserMigration.js
```diff
@@ -1,6 +1,5 @@
-const appRoot = require('app-root-path')
-const knex = require(`${appRoot}/db/knex`)
-const roles = require(`${appRoot}/modules/core/roles.js`)
+const knex = require('@/db/knex')
+const roles = require('@/modules/core/roles.js')
```
fabis94, line 2: These server scripts could be converted to yargs commands, but up to you

fabis94 (approved):
Added a few comments about potential improvements, but looks good overall 👍

packages/server/scripts/duplicateUserMigration.js
gjedlicska: yeah i was too lazy to do it :)

packages/server/bootstrap.js
gjedlicska: yeah i've pused an, we'll need the appRoot in `server/modules/index.js` line 34 anyhow

Issue #740, opened 2022-05-04
Add Webhook events for Comments


jsdbroughton:
**What package are you referring to?**
Speckle Server, Speckle Viewer

**Is your feature request related to a problem? Please describe.**
Once the realtime commenting feature has settled, could a webhook event be added to allow synchronicity with other systems?

**Describe the solution you'd like**
webhook events: comment_created, comment_edited, and comment_deleted

**Describe alternatives you've considered**
I have tested as I half-wondered if listening to stream_update and commit_update would fire off on comments (revealing secret sauce) but it is not and seemingly a separate data source

PR #745, opened 2022-05-07
Adds webhook events for comment actions

jsdbroughton:
- Adopting the pattern for the graphql resolvers from core services, saveActivity() is called on: create, reply, edit and archive of comments.
- The webhook payload is pretty straightforward for each. 
- Edit comment doesn't return anything, so this webhook will fire immediately after edit is called.

Addresses #740

### DIFF
--- a/packages/frontend/src/main/components/stream/WebhookForm.vue
+++ b/packages/frontend/src/main/components/stream/WebhookForm.vue
@@ -135,7 +135,11 @@ export default {
       'commit_receive',
       'commit_delete',
       'stream_permissions_add',
-      'stream_permissions_remove'
+      'stream_permissions_remove',
+      'comment_created',
+      'comment_updated',
+      'comment_archived',
+      'comment_replied',
     ],
     validation: {
       urlRules: [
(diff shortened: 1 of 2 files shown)


jsdbroughton:
No tests were broken, and as the webhook stream activity is effectively a side effect on the Comments code, I looked for how this is being addressed in the equivalent GraphQL resolvers in Core services. I couldn't see anything. 

I suppose `when comment is added the activity stream should contain a task with resourceId = commentID` but the ActivityStream would probably need to be mocked to catch that? Sorry still feeling my way into the code a bit.

I'm happy to take advice on what you want added to cover that.

didimitrie:
I'm planning to look at this today/tomorrow. Also scaffolding "public comments" at the same time so this is a timely draft PR!

didimitrie:
@jsdbroughton, for reference - I've copy-pasted your PR into https://github.com/specklesystems/speckle-server/pull/756, as it would be easier for us to merge. 

I'll close this one now, but you've been credited (i'm proud i figured out how to coerce the commit message to do so!). We're going to merge that in tomorrow, and we should be... only a few days late for the hackaton 😓  - still, it's in!

PR #747, opened 2022-05-10
WIP: Alex/viewer redux

fabis94:


packages/viewer-sandbox/src/main.ts
```diff
@@ -1,42 +1,32 @@
-import { Pane } from 'tweakpane'
-import { Viewer } from '@speckle/viewer'
+import { Viewer, IViewer } from '@speckle/viewer'
 import './style.css'
+import Sandbox from './Sandbox'
 
-const container = document.querySelector<HTMLDivElement>('#renderer')
+const container = document.querySelector<HTMLDivElement>('#renderer') as HTMLElement
```
fabis94, line 5: Don't forcefully recast types with `as` unless there's no other way, cause then 1) you leave the underlying typing problem unresolved 2) introduce the possibility of bugs where the TS compiler will think it's 1 type, but it's actually another.

In this case you can just change the generic type

packages/viewer-sandbox/src/main.ts
```diff
@@ -1,42 +1,32 @@
...
 if (!container) {
   throw new Error("Couldn't find #app container!")
 }
 
 // Viewer setup
-const viewer = new Viewer({
-  container,
-  showStats: true
-})
+const viewer = new Viewer(container) as IViewer
```
fabis94, line 11: Ditto, don't recast with `as`. If the @speckle/viewer package has real TS types now, then just delete the type augmentations I created for it in the sandbox package.

lerna-debug.log
```diff
@@ -0,0 +1,14 @@
+0 silly argv { _: [ 'clean' ], lernaVersion: '4.0.0', '$0': 'lerna' }
```
fabis94, line 1: Add this to gitignore. Also it shows that you've added a nonexistant/invalid package.

packages/viewer-sandbox/tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "esModuleInterop": true,
-    "noEmit": true,
+    // "noEmit": true,
```
fabis94, line 13: why?

packages/viewer/.prettierrc
```diff
@@ -0,0 +1,11 @@
+{
+    "trailingComma": "none",
+    "useTabs": false,
+    "tabWidth": 4,
+    "semi": true,
```
fabis94, line 5: not necessary, we already have global prettier settings in the monorepo root

packages/viewer/package.json
```diff
@@ -1,77 +1,82 @@
...
-    "jest": "27.2.5",
-    "mocha": "^9.1.2",
-    "prettier": "^2.5.1",
-    "rollup": "^2.70.1",
-    "rollup-plugin-delete": "^2.0.0",
-    "rollup-plugin-terser": "^7.0.2",
-    "yargs": "^17.2.1"
-  },
-  "gitHead": "5627e490f9a3ecadf19cc4686ad15f344d9ad2d3"
+    "name": "@speckle/viewer",
```
fabis94, line 2: Because of the formatting config changes you added (the prettier config file) it's now reformatted the entire file, please undo the changes and re-do the ones you did without the reformatting

packages/viewer/src/modules/SceneObjectManager.js
```diff
@@ -1,378 +1,404 @@
-import * as THREE from 'three'
-import debounce from 'lodash.debounce'
-import SceneObjects from './SceneObjects'
+import * as THREE from 'three';
+import debounce from 'lodash.debounce';
+import SceneObjects from './SceneObjects';
```
fabis94, line 3: Make sure you reformat all files back to their original formatting once you remove the prettier config

packages/viewer-sandbox/src/main.ts
AlexandruPopovici: Here the viewer instance is cast to the type which holds the public contract. This was intentional

packages/viewer-sandbox/tsconfig.json
AlexandruPopovici: Because I don't know how rollup works, and thought that since it's going to bundle up everything in the end, there would be no need for any standalone compiled js

packages/viewer-sandbox/src/main.ts
AlexandruPopovici: They are all HTMLElement

fabis94: Take a look at this example:

```ts
function fooBar(): any {
   // ...
   return someNumber;
}

// (in some other file)

const result = fooBar() as number;
```

For whatever reason the fooBar() function has an invalid return type and so you use `as` to cast it to number - because the value definitely is a number. Then later on someone rewrites the fooBar() function to this:

```ts
function fooBar(): bool {
   // ...
   return true;
}

// (in some other file)

const result = fooBar() as number;
```

The fooBar() function now has a completely different signature and yet TypeScript still thinks that `result` holds a number, because with `as` TypeScript was told to ignore whatever it think it is and to just use that forced type.

So forcefully casting a type to something other than what TypeScript thinks it is opens you up for more bugs in the future.
In this case it's unlikely that querySelector() would ever return something other than an HTMLElement, but being vary of using `as` is a good approach still and in this case you can easily specify the expected type in the generic argument of querySelector() without having to use `as`.

packages/viewer-sandbox/src/main.ts
fabis94: That's not necessary, because Viewer already has a TS type from the class (now that I've fixed it in the local version of the branch, I'm about to push it)

packages/viewer-sandbox/src/main.ts
AlexandruPopovici: I agree with the above. I never said that you can use `as` wheneve/wherever you want. However in this particular case, the use of `as` was very much harmless and also correct. `querySelector` does take a type argument here. My bad for not seeing that. However, in general, I don't rely much on typescript generics since they are just compile-time

packages/viewer-sandbox/src/main.ts
AlexandruPopovici: It wasn't about necessity, more for verbosity. You want to work with the public contract of the library, not with it's implementation

PR #760, opened 2022-05-23
feat(frontend): animated preview images

didimitrie:
changes:
- preview service generates full 360 images now
- preview image component loads on hover
- preview image component backwards compat with non-360 image angles
- adds optional angle arguments in all preview routes & removes superflous route

Quick demo:

https://user-images.githubusercontent.com/7696515/169804357-4a4ac3a9-3de2-475a-b257-e4adbfbb526e.mov

### DIFF
--- a/packages/frontend/src/main/components/common/PreviewImage.vue
+++ b/packages/frontend/src/main/components/common/PreviewImage.vue
@@ -1,18 +1,29 @@
 <template>
-  <div @mouseenter="hovered = true" @mouseleave="hovered = false" @mousemove="setIndex">
+  <div
+    style="position: relative"
+    @mouseenter="hovered = true"
+    @mouseleave="hovered = false"
+    @mousemove="setIndex"
+  >
     <v-img
       ref="cover"
       :height="height"
       cover
       :class="`${color ? '' : 'grasycale-img'} preview-img`"
-      :src="previewImages[imageIndex]"
+      :src="revImg[imageIndex]"
       :gradient="`to top right, ${
         $vuetify.theme.dark
           ? 'rgba(100,115,201,.13), rgba(25,32,72,.2)'
           : 'rgba(100,115,231,.075), rgba(25,32,72,.02)'
       }`"
     />
-    <v-progress-linear v-show="loading" indeterminate height="2" />
+    <v-progress-linear
+      v-show="loading"
+      indeterminate
+      height="4"
+      style="position: absolute; bottom: 0"
+    />
+    <!-- <span class="caption">{{ imageIndex }}</span> -->
   </div>
 </template>
 <script>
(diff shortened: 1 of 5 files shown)


packages/server/modules/previews/index.js
```diff
@@ -252,33 +235,33 @@ exports.init = (app) => {
         id: req.params.commitId
       })
       if (!commit) return res.sendFile(noPreviewImage)
-
+      console.log(req.params.angle || DEFAULT_ANGLE)
```
fabis94, line 238: Are all of the console.log calls meant to be there or is that leftover debugging?

packages/frontend/src/main/components/common/PreviewImage.vue
```diff
@@ -27,33 +35,119 @@ export default {
...
+      const blob = await res.blob()
+      return URL.createObjectURL(blob)
+    },
+    async getOtherAngles() {
+      // Note: previously, previews were generated for only 5 angles (-30deg, -15, 0, 15, 30deg), corresponding to
+      // labelled angles (-2, -1, 0, 1, 2). We have now switched to generating full 360deg previews in increments
+      // of 15 deg, going clockwise straight. Eg., 0 = 0deg, 1 = 15deg, 2 = 30 deg, etc.
+      this.loading = true
+      // starting in reverse to quickly figure out if we're on the legacy track!
+      for (let i = this.angles.length - 1; i > 0; i--) {
```
fabis94, line 132: It'll def be faster to load all of the images concurrently, instead of awaiting each `fetch` call sequentially.

e.g.
```js
const images = await Promise.all([
  this.getPreviewImage(0),
this.getPreviewImage(1),
this.getPreviewImage(2),
...
]);
```

packages/server/modules/previews/index.js
didimitrie: nope 😅

packages/frontend/src/main/components/common/PreviewImage.vue
fabis94: ah I see, yea you can do it that way. since the amount of angles (images) is pretty big, even if you do the first 2 sequentially, making all of the other ones concurrent should help a lot

fabis94: btw it's worth testing out how this will act on a slow network (in dev tools you can usually throttle your network connection to 3G, for example), another concern I have is that `loading` is set to `false` only after all images are loaded and there are quite a lot. might be more performant to show the images you have, while the other ones are loading

didimitrie: yea, i tested on slow 3g - not going work magic there. we display the first one by default anyway; the rest of the angles start loading after a hover/touchstart event.

PR #762, opened 2022-05-23
fix(frontend): fixed toolbars/navbars/action bars showing the wrong content randomly

fabis94:
It could be done cleaner without requiring state management for portals, but that would require re-architecting how all of the toolbars & nav bars are done. So for now we've at least got a working fix - portals only send things through if they've got approval from the portal state manager.

I went through all of the usages for the main portal targets (toolbar, nav, actions) and tested them out as well, seemed to work OK.


fabis94:
@didimitrie by the way, the portal targets can still flash on page load before they figure out which portal source should be used, but to resolve this we'd need to ensure all targets can only ever have a single source and that would require re-architecting this quite heavily

at least with this you get the right source

Issue #755, opened 2022-05-15
Deleted branches persist.


jsdbroughton:
**What package are you referring to?**
This manifests itself both on the server and has visible artefacts in the viewer

**Describe the bug**
After a branch has been deleted, the ability to create a new branch with the same name (within an untested period of time)  is blocked. The deleted branch persists in the branch listing (for a time). Eventually, this resolves itself.

**To Reproduce**
Delete a branch, either in the frontend UI or via Client API. Shortly afterwards attempts to create a branch with the same name are blocked (unexpected), and so too is the ability to commit to the deleted branch (expected)

**Expected behaviour**
Either, a) branch is created, or b) documentation or error message clarifies that the action is eventual and not immediate.

**Screenshots**
![Screenshot 2022-05-15 at 14 17 38](https://user-images.githubusercontent.com/760691/168475322-0c816d0a-9605-465d-92ec-d2718cc0c30a.png)

UI artefact - the branch is visible in the listing (unexpected) but the query to retrieve it fails (expected). This mirrors the API call behaviour.

**Additional context**
This may be expected behaviour, the edge case of rapidly creating a deleted branch of the same name is not a frequent activity I assume. e.g. Deleting an email/id from Google, for instance, warns this may be a 24-hour wait for completion. Removing references to a branch in db may be similar.

gjedlicska:
@didimitrie how were you able to repro this? I did not manage to do it on latest

jsdbroughton:
I can extract minimal reproducible if it helps.

didimitrie:
I reproduced... manually:

![image](https://user-images.githubusercontent.com/7696515/168791727-112884c1-efa2-4e02-92a9-be77c2055bf9.png)

Call it ~~fuzz~~ furious testing 😅

gjedlicska:
Ok, i still cannot repro the branch remaining, but i ran into another, possibly related issue.

Steps:

* create branch
* delete branch ( it gets removed from the branch list for me ever time)
* create branch with the same name
* delete branch fails, thinks branch doesn't exist, while the gql api returns the undeletable branch. 

I'm thinking both of these could be an apollo / graphql cache thing.

![image](https://user-images.githubusercontent.com/57442769/168795176-d7ad5567-7947-44f4-8a42-485e8d7315bf.png)

jsdbroughton:
It's intermittent behaviour feels like a race condition.

fabis94:
[...]

PR #767, opened 2022-05-25
fix(frontend): branch stale cache bug when creating/deleting them

fabis94:
Fixed 3 bugs actually:

1. The branch entity remaining in the cache after it's deleted, causing issues later on (e.g. when creating a new one with the same name and trying to delete it). At first I tried to resolve this without refetching from server & just updating the Apollo cache, but the ApolloClientv2 cache eviction utilities are nonexistant (there are better ones in v3) so it was easier to refetch

2. The branch edit dialog component mutated its inputs which shouldn't be done. In this case it wasn't a mutation of the props, but the data that comes directly from the Apollo cache. So previously while you were changing the name of the branch, you were making direct changes to the apollo cache.

3. Fix for the issue Jedd mentioned once where the User Avatar hover menu closes once you select text inside of it. From now on the hover menu won't close on any clicks inside its content.

closes https://github.com/specklesystems/speckle-server/issues/755

### DIFF
--- a/packages/frontend/src/main/components/common/UserAvatar.vue
+++ b/packages/frontend/src/main/components/common/UserAvatar.vue
@@ -1,6 +1,6 @@
 <template>
   <div style="display: inline-block">
-    <v-menu v-if="$loggedIn()" offset-x open-on-hover>
+    <v-menu v-if="$loggedIn()" offset-x open-on-hover :close-on-content-click="false">
       <template #activator="{ on, attrs }">
         <div v-bind="attrs" v-on="on">
           <user-avatar-icon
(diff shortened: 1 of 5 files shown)


didimitrie (approved):

didimitrie:
Approved, looks nice! I'd want a small master class actually around point (2) when we have the time :)

fabis94:
> Approved, looks nice! I'd want a small master class actually around point (2) when we have the time :)

Sure thing, we can have a chat about this at any point. [Here's a quick article](https://michaelnthiessen.com/avoid-mutating-prop-directly/) about why you shouldn't mutate props in Vue, and even though no props were mutated here I think it's the same problem generally speaking - the component is mutating data that doesn't belong to it, making it more difficult to track how data changes in the app possibly introducing weird and hard to debug bugs. The only way devs will expect the Apollo Client Cache to change is when some kind of GQL operation is executed or when the cache is mutated through the `cache.writeQuery` and `cache.writeFragment` utilities. They definitely won't expect some Vue component mutating the cache objects directly.

PR #775, opened 2022-05-25
gergo/testCommentsGQL

gjedlicska:
- yarn first go
- fix frontend build cache loader
- yarn workspaces built server Docker
- build(yarn): add workspaces plugin config
- chore(package defs): clean package*.json -s
- chore(gitignore): ignore yarn error log
- build(yarn): update yarn lock
- build(preview-service webpack): add extra resolved path to preview service webpack config
- build(docker): update dockerignore with yarn specific configs
- build(docker): update Dockerfiles for yarn workspaces utilization
- ci(circleci): update server test job to yarn
- ci(circle): disable cache restore
- ci(circleci): trying the node orb yarn-run
- ci(circleci): yarn-run again
- ci(circleci): disable node orb
- ci(circleci): change base node image for tests
- ci(circleci): add yarn cache
- ci(circleci): remove node install step
- ci(circleci): add server specific cache archives
- ci(circleci): test build and publish
- ci(circleci): change npm auth method to suit yarn
- ci(circleci): trying new builder image
- ci(circleci): another base image, maybe this works
- ci(circleci): force a specific docker engine version
- ci(circleci): add yarn version plugin and its changes
- ci(circleci): cleanup and remove temp branch config
- chore(package defs): moving from npm run to yarn
- explicitly specifying webpack4 as a frontend dep
- chore(package defs): replace npm with yarn everywhere
- docs(root readme): update with some yarn specific docs
- test(server comments gql): add wip server comments gql tests
- test(server comments graphql): add missing test operations and generate a bunch of testcases
- test(server comments graphql api): fix all authz test cases for comments


packages/server/modules/comments/graph/resolvers/comments.js
```diff
@@ -67,41 +84,54 @@ module.exports = {
     }
   },
   Stream: {
-    async commentCount(parent) {
+    async commentCount(parent, args, context) {
+      if (context.role === Roles.Server.ArchivedUser)
```
gjedlicska, line 88: this is where this gets hairy

packages/server/modules/comments/graph/resolvers/comments.js
```diff
@@ -67,41 +84,54 @@ module.exports = {
...
+        throw new ForbiddenError('You are not authorized.')
       return await getStreamCommentCount({ streamId: parent.id })
     }
   },
   Commit: {
-    async commentCount(parent) {
+    async commentCount(parent, args, context) {
+      if (context.role === Roles.Server.ArchivedUser)
```
gjedlicska, line 95: and these are a bit scary

packages/server/modules/comments/graph/resolvers/comments.js
```diff
@@ -1,6 +1,8 @@
-const { authorizeResolver, pubsub } = require('@/modules/shared')
+const { pubsub } = require('@/modules/shared')
 const { ForbiddenError, ApolloError, withFilter } = require('apollo-server-express')
+const { Forbidden } = require('@/modules/shared/errors')
```
fabis94, line 3: The common convention is to always suffix all error subclasses with `Error`. I realize that there already is a `ForbiddenError` in this scope, but then you should just rely on import aliases (`import {ForbiddenError as OtherForbiddenError} from 'whatever'`)

packages/server/modules/comments/graph/resolvers/comments.js
```diff
@@ -173,21 +203,32 @@ module.exports = {
...
+        userId: context.userId,
+        serverRole: context.role,
+        auth: context.auth,
+        //public, but not public comments needs a stream role to do this
+        requireRole: true
+      })
+      const matchUser = !stream.role
+      try {
+        await editComment({ userId: context.userId, input: args.input, matchUser })
+        return true
```
fabis94, line 217: The function only has a return in one of its execution branches. You should either omit returns altogether (if it's a void return) or make sure all execution branches have one.

You could argue that it's impossible for anything other than `true` to be returned, but that would be more clear if the `return true` statement would be moved to the end of the function.

packages/server/modules/comments/services/index.js
```diff
@@ -1,12 +1,55 @@
...
+  switch (res.resourceType) {
+    case 'stream':
+      // Stream validity is already checked, so we can just go ahead.
+      break
+    case 'commit': {
+      const linkage = await knex('stream_commits')
+        .select()
+        .where({ commitId: res.resourceId, streamId })
+        .first()
+      if (!linkage) throw new Error('Commit not found')
```
fabis94, line 22: Ideally we'd want to use custom error subclasses, so that we can ensure our `catch` clauses only catch what we intend to catch and that we don't have to match errors by their messages which is easy to break

packages/server/modules/comments/tests/comments.graph.spec.js
```diff
@@ -1,4 +1,853 @@
-describe('Graphqql @comments', () => {
+const expect = require('chai').expect
```
fabis94, line 1: Didn't go through this very thoroughly, but it's cool that you're running operations against `apollo`, not the service layer directly, and that it's all defined so abstractly

gjedlicska: The point of this is, that after all this boilerplate, it was and will be really easy to add user, stream, action permutations to the test data object and all the rest is done automatically.

PR #782, opened 2022-05-30
fix(preview service): there is no need to double promise a preview

gjedlicska:
fixed the linting error https://results.pre-commit.ci/run/github/255673761/1653894937.ppQxFVvgThqePL35tQsgqA


didimitrie (approved):
if it twerks locally, should twerk remotely :)

PR #789, opened 2022-06-05
Silencing the noisy Embedded Viewer console log

jsdbroughton:
The embedded viewer either doesn't implement the event-hub or doesn't add it at mount time for the Viewer Filter component. Consequently, this is noisy with associated errors of accessing the $on event triggers for $eventHub.

This edit makes the addition of the $on event listener conditional on the $eventHub being instantiated.


didimitrie:
I'm thinking to rethink the whole embedding thing. In the meantime, thanks - this looks good!

PR #794, opened 2022-06-07
refactor(object-loader): disable debug console logs

vwnd:
When working with the object-loader on a server, several logs which I believe were there for debug purposes were "polluting" the logging. For example, if am downloading a large object, or if I am instantiating several ObjectLoaders.

E.g.:

image

image

### DIFF
--- a/packages/objectloader/src/index.js
+++ b/packages/objectloader/src/index.js
@@ -30,7 +30,7 @@ export default class ObjectLoader {
     this.serverUrl = serverUrl || window.location.origin
     this.streamId = streamId
     this.objectId = objectId
-    console.log('Object loader constructor called!')
+    // console.log('Object loader constructor called!')
     try {
       this.token = token || localStorage.getItem('AuthToken')
     } catch (error) {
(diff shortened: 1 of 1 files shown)


fabis94 (changes requested):
@didimitrie looks fine to me, but there are some ESlint issues now

packages/objectloader/src/index.js
```diff
@@ -253,7 +253,7 @@ export default class ObjectLoader {
       count += 1
       yield obj
     }
-    console.log(`Loaded ${count} objects in: ${(Date.now() - t0) / 1000}`)
+    // console.log(`Loaded ${count} objects in: ${(Date.now() - t0) / 1000}`)
```
didimitrie, line 256: I'd comment these lines out too to make the linter happy, and we're good to merge! 
```
const t0 = Date.now()
let count = 0
```
Happy to keep them as comments as we will be benchmarking loading times on this in the future 😅

didimitrie (changes requested):

PR #802, opened 2022-06-14
Add blob storage backend

gjedlicska:
- feat(server): add server authz pipeline rework first sketch
- feat(server authz): add new server authz middleware poc implementation
- test(server authz): add unittests for the new server authz workflow
- feat(wip rework of fileuploads vs blob storage): add basim impl of separate blob storage service
- feat(fileimport service): refactored file import service to utilize the new asssetstorage service
- refactor(server errors): refactor server errors to use the shared module definitions
- refactor(fileimport service): cleanup after refactor
- feat(frontend fileimports): use the new blob storage for downloading the original file
- refactor(server fileimports): clean up the remnants of S3 storage from file imports
- refactor(server authz): centralize generic authz pipeline configs
- refactor(server blob storage): refactor / rename everything to use the `blob-storage` name
- ci(circleci): add s3 objectstorage environment variables
- ci(circleci): fix missing env variables
- ci(circleci): add minio test container
- ci(circleci): fix minio app startup
- ci(circleci): enable circleci remote docker
- ci(circleci): fix minio startup
- ci(cirleci): detach and wait properly for minio to start
- ci(circleci): revert to additional minio img config, it only fails when the container is stopped ?!
- ci(circleci): disable file uploads
- fix(fileimports): update with blob storage refactor leftovers
- feat(server blob storage): add blob storage graphql api


packages/server/modules/shared/errors/base.js
```diff
@@ -57,31 +57,4 @@ class BaseError extends VError {
...
-
-/**
- * Use this to throw when user tries to access data that he shouldn't have access to
- */
-class UnauthorizedAccessError extends BaseError {
-  static code = 'UNAUTHORIZED_ACCESS_ERROR'
-  static defaultMessage = 'Attempted unauthorized access to data'
-}
-
-class RichTextParseError extends BaseError {
```
gjedlicska, line 76: I've moved these errors into the index package and renamed with the Speckle prefix

.circleci/config.yml
```diff
@@ -146,7 +152,7 @@ jobs:
             - .yarn/cache
             - .yarn/unplugged
 
-      - run: 'dockerize -wait tcp://localhost:5432 -timeout 1m'
+      - run: 'dockerize -wait tcp://localhost:5432 -wait tcp://localhost:6379 -timeout 1m'
```
gjedlicska, line 155: This was missing, waiting for redis to be ready. Luckily it wasn't causing issues so far.

packages/fileimport-service/package.json
```diff
@@ -15,7 +15,8 @@
     "node": "^16.0.0"
   },
   "scripts": {
-    "dev": "cross-env S3_BUCKET=speckle-server POSTGRES_URL=postgres://speckle:speckle@localhost/speckle NODE_ENV=development SPECKLE_SERVER_URL=localhost:3000 nodemon ./src/daemon.js",
+    "dev": "cross-env S3_BUCKET=speckle-server POSTGRES_URL=postgres://speckle:speckle@localhost/speckle NODE_ENV=development SPECKLE_SERVER_URL=http://localhost:3000 nodemon ./src/daemon.js",
+    "parse:ifc": "node ./ifc/import_file.js /tmp/file_to_import/file 33763848d6 2e4bfb467a main File upload: steelplates.ifc",
```
gjedlicska, line 19: This is just an example on how to run the ifc parser without the fileimport service

packages/server/app.js
```diff
@@ -106,15 +105,15 @@ exports.init = async () => {
     app.use(compression())
   }
 
-  app.use(bodyParser.json({ limit: '100mb' }))
-  app.use(bodyParser.urlencoded({ limit: '100mb', extended: false }))
+  app.use(express.json({ limit: '100mb' }))
```
gjedlicska, line 108: this is the same in functionality, but express now bundles the .json method

gjedlicska, line 108: express now bundles the bodyParser functionality with is deprecated

packages/server/modules/fileuploads/index.js
```diff
@@ -27,136 +34,30 @@ exports.init = async (app) => {
...
-
-    try {
-      await authorizeResolver(
-        req.context.userId,
-        req.params.streamId,
-        'stream:contributor'
+  app.post(
+    '/api/file/:fileType/:streamId/:branchName?',
+    contextMiddleware,
+    authMiddlewareCreator(streamWritePermissions),
```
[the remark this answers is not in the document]
didimitrie: Yes, it's hairy. I'm inclined to allow it, as GH allows it - just did a test in a random repo. The use case would be `I'm Random Bob The Builder on site, i just got a link from my super, and I want to attach a pic to him to explain the issue`. 

The how is tricky. I'm getting my head around the code now so we can discuss tomorrow.

packages/server/modules/blobstorage/services/index.js
```diff
@@ -0,0 +1,111 @@
+const knex = require('@/db/knex')
+const {
+  SpeckleNotFoundError,
+  SpeckleResourceMismatch
+} = require('@/modules/shared/errors')
+const BlobStorage = () => knex('blob_storage')
+
```
fabis94, line 7: i don't like the approach of creating a service called `index.js` in our modules - what if we want to create other services? also, I think such an approach encourages devs to grow the file to be bigger and handle more concerns than it needs to

packages/server/modules/shared/authz.js
```diff
@@ -0,0 +1,245 @@
...
+} = require('./errors')
+
+class SpeckleContextError extends Error {
+  constructor(message) {
+    super(message)
+    this.name = 'SpeckleContextError'
+  }
+}
+
+// const authorizeStreamAccess = async ({
```
fabis94, line 16: these should probably be deleted

packages/server/modules/shared/index.js
```diff
@@ -92,14 +92,20 @@ async function contextMiddleware(req, res, next) {
 
 let roles
 
+const getRoles = async () => {
```
fabis94, line 95: The issue I had with blobstorage/services/index.js goes double for this file. It will tend to collect all kinds of random stuff and have a ton of different concerns. Moving stuff to a common/shared folder makes sense to me, but the shared code should still be organized according to concern and no single module should have too many concerns.

I realize that this file already existed beforehand, so this isn't necessarily about your addition, but a comment about the situation generally

fabis94 (changes requested):
@gjedlicska overall looks fine (didn't go too deeply into the core logic tho), i've left some comments regarding some pretty minor issues

packages/server/modules/blobstorage/migrations/202206030936_add_asset_storage.js
```diff
@@ -0,0 +1,31 @@
+/* istanbul ignore file */
+
+const TABLE_NAME = 'asset_storage'
```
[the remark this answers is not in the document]
gjedlicska: lol i thought i fixed this :D

packages/server/modules/blobstorage/migrations/202206030936_add_asset_storage.js
```diff
@@ -0,0 +1,31 @@
...
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = async (knex) => {
+  await knex.schema.createTable(TABLE_NAME, (table) => {
+    table.string('id', 10)
+    // dont cascade on delete, cause it doesn't clean the object storage for the objs
+    // it needs to be exposed as a service, to be able to cleanup fully after a stream
+
+    table.string('streamId', 10).references('id').inTable('streams')
```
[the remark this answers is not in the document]
gjedlicska: 1 -> st makes sense
2 -> we can cross that bridge when we get there, it might even make sense to store them in a separate table or smthing

packages/server/modules/blobstorage/services/index.js
gjedlicska: makes sense, but its an easy refactor when the stuff gets bigger
and while the blobstorage has only 1 type of service, i was having a hard time comming up with a name for a service that is blobstorage / blobstorage.

BTW what is ur preferred limit of stuff in a single module / file?

packages/server/modules/blobstorage/index.js
```diff
@@ -0,0 +1,210 @@
...
+
+  if (!process.env.S3_BUCKET) {
+    debug('speckle:error')(
+      'S3_BUCKET env variable was not specified. 📦 BlobStorage will be DISABLED.'
+    )
+    return
+  }
+}
+
+const errorHandler = async (req, res, callback) => {
```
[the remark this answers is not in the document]
gjedlicska: yes, that is where i was going with this, but lets put that to a separate PR.

I'll rework all the rest endpoints middlewares soon

packages/server/modules/shared/index.js
gjedlicska: here i agree 100%. This will get cleaned up with the authz/middleware rework

packages/server/modules/blobstorage/services/index.js
fabis94: > BTW what is ur preferred limit of stuff in a single module / file?
There's not a hard line, it's more of a feeling, but i'd say that when you go over 500 LOC you're probably starting to put too much stuff in there. it would be nice if we could even implement this as a linting rule, there's a rule for that.

> its an easy refactor when the stuff gets bigger
agreed, and yea until you actually figure out what those extra concerns are, it's not realistic to just split it apart ahead of time. so I agree that it can remain all in one file for now, but I def disagree with it being called 'index.js' then, cause that kinda marks it as the "main" service, which won't make sense when there will be many. even something like "blobStorageService.js" or "blobManagementService.js" would be better.

PR #803, opened 2022-06-14
feat(frontend): typescript in frontend

### LINKED ISSUE #799: TypeScript in the Frontend
frontend

We want to start using TypeScript in the frontend for improved maintainability & DX

fabis94:
closes #799 

- Supports .ts & TS .vue files in `frontend`
- Added `yarn gqlgen` command for (re-)generating GraphQL TS types
- Re-wrote a bit of code to TS as an example


packages/server/modules/core/graph/schemas/streams.graphql
```diff
@@ -70,7 +70,7 @@ type StreamCollaborator {
 type StreamCollection {
   totalCount: Int!
   cursor: String
-  items: [Stream]
+  items: [Stream!]
```
fabis94, line 73: doesn't make sense that Stream was nullable in the first place (do we really allow `[null]` arrays?), do you guys predict any issues with this? @didimitrie @gjedlicska

didimitrie: IIRC, the intention was to allow for empty arrays `[]` - that's all. If the new way allows for that, we're cool

fabis94: yea, it does, should be safe.

fyi:
`[X!]!` -> `[]` or `[X]`
`[X!]` -> `[]` or `[X]` or `null`
`[X]` -> `[]` or `null` or `[X]` or `[X, null]`

PR #804, opened 2022-06-15
fix(frontend): embed viewer bugfixes & speed improvements

### LINKED ISSUE #773: Optimized embed bundle with lazy loading
This relates to the `frontend`, specifically the embed endpoint and its bundle

We already have the play button on the embed UI, which opens up an obvious opportunity to lazy load all of the (very heavy) JS dependencies only once it's pressed.

Considering that these embeds are used all across the web, we want them to load quickly and in a performant way so that the users of those sites don't get degraded performance just because of us.

We should refactor `EmbedViewer.vue` so that it lazy loads as much as possible. When the component is mounted it should only show the play button and the background preview image.

fabis94:
closes #773


packages/frontend/src/embed/EmbedViewer.vue
```diff
@@ -319,26 +186,51 @@ body::-webkit-scrollbar {
   height: 100vh !important;
   width: 100vw !important;
   position: fixed;
+
   &::-webkit-scrollbar {
     display: none;
   }
+
+  top: 0;
```
fabis94, line 194: These originally came from SpeckleViewer.vue. Since it's now asynchronized, we need to copy these styles over

packages/viewer/src/modules/SceneObjectManager.js
```diff
@@ -13,10 +13,11 @@ import { getConversionFactor } from './converter/Units'
  * Manages objects and provides some convenience methods to focus on the entire scene, or one specific object.
  */
 export default class SceneObjectManager {
+  views = []
```
fabis94, line 16: @AlexandruPopovici had to add this as a public class field, otherwise the exported TS type won't have it

packages/viewer/src/modules/Viewer.ts
```diff
@@ -270,11 +270,7 @@ export class Viewer extends EventEmitter implements IViewer {
     this.cameraHandler.toggleCameras()
   }
 
-  public async loadObject(
-    url: string,
-    token: string | undefined,
-    enableCaching = true
-  ) {
+  public async loadObject(url: string, token?: string, enableCaching = true) {
```
fabis94, line 273: Previously the 2nd param wasn't optional

didimitrie (approved):
think i caught all the pokemons 🤞

PR #809, opened 2022-06-20
Various papercuts

### LINKED ISSUE #806: Malformed comments raised on Commits or Objects crash the viewer
Frontend (and by association to the issue discovery, GraphQL)

**Describe the bug**
The GraphQL schema allows for any form of JSONObject to be attached as a `data` property to the comment input. If this data property doesn't include ```{ "location": { "x"…, "y"…, "z"… } }```, the viewer will either not display the model at all or it will load the geometry and interactivity is broken.

This stems from the comment rendering routine accessing the x, y & z values from the `data.location` and failing as accessing properties from `undefined`
https://github.com/specklesystems/speckle-server/blob/f2df716a10d7500f68ffb57090b2871c14122154/packages/frontend/src/main/components/viewer/CommentsOverlay.vue#L504-L508

**To Reproduce**
1. Add a comment via GQL in the form:
```
mutation CommentCreateInput {
  commentCreate(
    input: {
      streamId: STREAM_ID, 
      resources: [
        { 
          resourceType: Commit
          resourceId: …, 
        }
      ], 
      text: "🍌 Bananagram 🍌", 
      data: {}
    }
  )
}
```

This is acceptable according to the `Type CommentCreateInput`

2. Access the commit in the browser.

**Expected behaviour**
A) Data property schema is documented / validated / rejected
or
B) Comment parsing is defensive around properties that are not guaranteed to be present.
or 
C) Both, and.

**Additional context**
Archiving the dodgy comment restores viewer functionality, so it is just the overlay component rendering and no other parsing routine so far as I can tell.

didimitrie:
- Fixes issue reported by @jsdbroughton re embedding that happened on streams with no commits on main. I've cleaned up the whole logic in there to make it more robust. 
- Adds option to have transparent background for embedded iframes (cc @bimgeek)
- Removes `frameborder` around iframes
- Adds rotation preview image in embedded viewer
- Makes stream share button more visible (given the success of @bimgeek's notion tutorial, thought it's worth it!)
- closes #806 (ignores all comments with no location info)
- properly refreshes branches when new branch is created
- no longer edits the prop when editing commits
- displays toast notification that commit has been moved to another branch if that's changed
- re-adds embedding code in stream share dialog for streams and branches
- unifies copy around private/public streams to "link sharing on" & "link sharing off" 
- allows commit edits by stream owners as well as commit authors 
- tsconfig `strict` set to false so I can actually get some work done (cc @fabis94) - this might be a controversial change, but I'm happy to switch it back once web team (read: me!) is more at home with TS in general.


fabis94:
@didimitrie can you please elaborate on the tsconfig change? i don't see any changes to TS files in this PR, so it's not clear how TS strict mode caused any issues here and what they were

PR #810, opened 2022-06-20
feat: stream comment attachments

### LINKED ISSUE #801: Stream comment attachments functionality
frontend/server

Users need to be able to attach files to stream comments

Once #771 & #757 are done, we can implement this functionality

fabis94:
closes #801


packages/server/modules/shared/authz.js
```diff
@@ -52,7 +52,7 @@ const validateStreamRole = ({ requiredRole }) =>
     requiredRole,
     rolesLookup: getRoles,
     iddqd: Roles.Stream.Owner,
-    roleGetter: (context) => context.stream.role
+    roleGetter: (context) => context.stream?.role
```
gjedlicska, line 55: why do we need this?

fabis94: I was getting errors cause of stream being undefined in some scenarios
i'm guessing in some test cases the context.stream resolution fails, and then because of that these auth checks fail as well (but in a messy way, throwing "undefined object" errors)

didimitrie (approved):
This is exciting :D

didimitrie:
Okay, got a bit more excited and added a super basic "attachment preview" function as an intermediary step before download. Helpful to see a picture in context rather than download it, open, etc.

PR #812, opened 2022-06-21
fix: fixing lint-staged/precommit-ci eslint configuration issues

fabis94:
up till now both lint-staged & precommit-ci reported an eslint warning about testing ignored files, this fixes that

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ repos:
       - id: eslint
         types: [file]
         files: \.[jt]s$|vue$ # *.js, *.ts and vue
-        exclude: '(\/|^)((.*generated\/.*)|(\..*\.([jt]sx?|vue)))$'
+        exclude: '(\/|^)((generated\/.*)|(\..*\.([jt]sx?|vue)))$'
         args:
           - '--max-warnings=0'
         additional_dependencies:


.pre-commit-config.yaml
```diff
@@ -5,7 +5,7 @@ repos:
       - id: eslint
         types: [file]
         files: \.[jt]s$|vue$ # *.js, *.ts and vue
-        exclude: 'mocharc.js$'
+        exclude: '(\/|^)((.*\.ignored\.([jt]sx?|vue))|(\..*\.([jt]sx?|vue)))$'
```
fabis94, line 8: the regex is quite scary, but unfortunately pre-commit doesn't give us any better ways to filter out files

packages/frontend/src/main/pages/TheFavoriteStreams.vue
```diff
@@ -44,7 +44,7 @@ import {
   UserFavoriteStreamsQuery,
   UserFavoriteStreamsQueryVariables,
   useUserFavoriteStreamsQuery
-} from '@/graphql/generated/graphql'
+} from '@/graphql/generated/graphql.ignored'
```
gjedlicska, line 47: this is the only part that feels very icky about this :/

gjedlicska (approved):

Issue #750, opened 2022-05-13
File upload time out and upload limits


teocomi:
A few users are reporting getting a timeout or hitting file size limits using our file upload service with IFC files. It's a known limitation set to safeguard the service. We should look into adjusting or removing it as it's blocking adoption and usability.

iainsproat:
As a first step, suggest that this should be configured per server deployment.  This would allow us to set larger limits on a case-by-case basis, especially enterprise deployments where the risk of a bad actor abusing large uploads is lower.

PR #835, opened 2022-07-22
Feat: configurable file limits

iainsproat:
#750


packages/server/modules/blobstorage/index.js
```diff
@@ -50,6 +50,16 @@ const ensureConditions = async () => {
   }
 }
 
+let fileLimit = null
+
+const getFileLimit = () => {
+  if (!fileLimit) {
+    //100Mb default
+    fileLimit = process.env.FILE_SIZE_LIMIT_BYTES || 1024 * 1024 * 100
```
gjedlicska, line 58: you do not need to predefine the fileLimit variable on line 53, it could only be set from sourcecode.
this func could be just

```javascript
const getFileLimit = () => {
  const defaultFileSizeLimit = 1024 * 1024 * 100
  return process.env.FILE_SIZE_LIMIT_BYTES || defaultFileSizeLimit
}
```

gjedlicska:
We should also add a variable to configure the worker time limit.
Lets add a `FILE_IMPORT_TIMEOUT_SECONDS` env var to the `fileimport-service`

And use it in every call of [this function](https://github.com/specklesystems/speckle-server/blob/79f558c726b6c4d35fdfd8fe788006f621566263/packages/fileimport-service/src/daemon.js#L184)

.circleci/config.yml
```diff
@@ -179,6 +179,7 @@ jobs:
   docker-build-and-publish: &docker-job
     docker: &docker-image
       - image: cimg/node:16.15
+    resource_class: xlarge
```
iainsproat, line 182: Should these unrelated changes be in a separate PR?

packages/frontend/nginx/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+set -e
```
iainsproat, line 2: `set -eu pipefail` would be safer

packages/frontend/nginx/docker-entrypoint.sh
```diff
@@ -0,0 +1,11 @@
...
+
+defined_envs=$(printf '${%s} ' $(env | cut -d= -f1))
+
+echo Starting nginx environment template rendering with $defined_envs
+
+envsubst "$defined_envs" < /etc/nginx/templates/nginx.conf.template > /etc/nginx/conf.d/nginx.conf
+
+echo Nginx conf rendered, starting server...
+exec "$@"
```
iainsproat, line 11: missing newline at EOF

packages/server/modules/blobstorage/index.js
```diff
@@ -50,6 +50,16 @@ const ensureConditions = async () => {
   }
 }
 
+const getFileLimit = () => {
+  //100Mb default
+  let sizeLimit = 100
+  const suppliedSize = process.env.FILE_SIZE_LIMIT_MB
+  if (suppliedSize) sizeLimit = parseInt(suppliedSize)
```
iainsproat, line 57: should we check for `NaN`?

utils/helm/speckle-server/templates/deployment-frontend.yml
```diff
@@ -43,3 +43,7 @@ spec:
             port: 80
           initialDelaySeconds: 5
           periodSeconds: 5
+        
+        env: 
+          - name: FILE_SIZE_LIMIT_MB
```
iainsproat, line 48: Does the following line also need to be changed to use this environment variable? https://github.com/specklesystems/speckle-server/blob/a10c49e7310ce84cbbbb9657a8b73514336214cb/packages/frontend/src/main/components/common/file-upload/FileUploadZone.vue#L84

gjedlicska: Right, this is missed.

.circleci/config.yml
gjedlicska: Since the containers are not build (yet) on PRs, there is no way to test them on CI atm. So i thought its fine...

packages/server/modules/blobstorage/index.js
gjedlicska: right, fixed it

utils/helm/speckle-server/templates/deployment-frontend.yml
iainsproat: this is the bit I was struggling with, I didn't know how Vue was able to get environment variables from the server.  Would like an explanation if you know/figure it out!

gjedlicska: you cannot, afaik in static build vue. It needs to come from the server via an API. That is also cleaner anyhow, so thats what im going to add.

packages/frontend/src/main/components/comments/CommentEditor.vue
```diff
@@ -84,10 +85,21 @@ export default Vue.extend({
       default: true
     }
   },
+  apollo: {
+    serverInfo: {
+      query: gql`
+        query serverInfo {
```
fabis94, line 91: This is OK for now, but it lacks types (and this is a TS component), so if you'd use `this.serverInfo` anywhere inside the component it would just have an `any` type and would need to be re-cast.

A better option is to do the following:
1. Move this query to `/packages/frontend/src/graphql/server.js` where the rest of the server queries are defined and make sure you use a globally unique query name (e.g. `query ServerFileSizeInfo { ... }`
2. Run `yarn gqlgen` to re-generate GraphQL Code Generator types
3. Import the typed query document from `'@/graphql/generated/graphql'` - you can import both the actual document to put into `query` - it'll be named ServerFileSizeInfoDocument - and also the type of the retrieved value - `ServerFileSizeInfoQuery`.
4. To type `this.serverInfo` property, create a `data()` prop like this:
```
data() {
 return {
   serverInfo: null as ServerFileSizeInfoQuery | null,
 }
}
```

(this is the suggested way to define and use queries elsewhere as well. an ever better option to use with better ergonomics would be to use the `useQuery` composable (look it up in our codebase if interested), but that's not a requirement)

packages/server/modules/blobstorage/services.js
```diff
@@ -154,6 +154,16 @@ const updateBlobMetadata = async (streamId, blobId, updateCallback) => {
   return { blobId, fileName, ...updateData }
 }
 
+const getFileSizeLimit = () => {
+  //100Mb default
+  let sizeLimit = 100
+  const suppliedSize = parseInt(process.env.FILE_SIZE_LIMIT_MB)
```
fabis94, line 160: I try to keep all env variables in `./modules/core/helpers/envHelper.js` so they're never accessed directly from process.env

packages/frontend/src/main/components/comments/CommentEditor.vue
gjedlicska: this is exactly why i wanted to discuss it with you, how to make the query nicely reusable, since its already there in 2 places. Thanks will do this.

packages/frontend/src/main/components/comments/CommentEditor.vue
```diff
@@ -49,8 +48,11 @@ import {
 import FileUploadProgress from '@/main/components/common/file-upload/FileUploadProgress.vue'
 import { UploadFileItem } from '@/main/lib/common/file-upload/fileUploadHelper'
 import { differenceBy } from 'lodash'
+import { useQuery } from '@vue/apollo-composable'
+import { serverInfoBlobSizeLimitQuery } from '@/graphql/server'
```
fabis94, line 52: You should import the query from '@/graphql/generated/graphql' instead, then it'll have the proper types and everything. Currently I'm not sure how the types are working for you on lines 90-95, I'd think that `result` just has type `any`

packages/server/modules/blobstorage/services.js
gjedlicska: fair enough, i do not like cross module dependency tangles, so moving the env helper to shared.

packages/frontend/src/main/components/comments/CommentEditor.vue
```diff
@@ -157,8 +167,8 @@ export default Vue.extend({
     },
     placeholder(): string {
       return this.addingComment
-        ? 'Your comment... (press enter to send)'
-        : 'Reply... (press enter to send)'
+        ? 'Your comment... (Enter sends it)'
+        : 'Reply... (Enter sends it)'
```
iainsproat, line 171: 'Press enter to send' is the more commonly used phrase

docker-compose-speckle.yml
```diff
@@ -8,6 +8,8 @@ services:
     restart: always
     ports:
       - '0.0.0.0:80:80'
+    environment:
+      FILE_SIZE_LIMIT_MB: 100
```
iainsproat, line 12: Does it still need to be set in the frontend as well as the server?

utils/1click_image_scripts/template-docker-compose.yml
```diff
@@ -41,6 +41,8 @@ services:
     restart: always
     ports:
       - '127.0.0.1:8000:80'
+    environment:
+      FILE_SIZE_LIMIT_MB: 100
```
iainsproat, line 45: again, do we need it set in both containers?

packages/frontend/src/main/components/comments/CommentEditor.vue
gjedlicska: but this hurts my eyes
image

utils/1click_image_scripts/template-docker-compose.yml
gjedlicska: yes, because its also set on the nginx config

gjedlicska: see here
[    client_max_body_size ${FILE_SIZE_LIMIT_MB}m;](https://github.com/specklesystems/speckle-server/blob/10d02da741255c43641292201fe0cbbf3e1555b0/packages/frontend/nginx/templates/nginx.conf.template#L113)

packages/frontend/src/main/components/comments/CommentEditor.vue
fabis94: I agree that "Press enter to send" is more common. Maybe we can just remove the "Your comment..." prefix? It's kinda self-explanatory that your comment can go there

PR #845, opened 2022-07-31
Blobs API SDK Driven Changes

didimitrie:
tl;dr: 
- [server] we're now accepting client file hashes that we use as blob ids
- [server] adds a new endpoint for diffing blobs based on said hashes (allows clients to not send existing blobs) 
- [server] guards with a `on conflict do nothing` blob entry creation 
- [objectloader] filters out blob ids from object's `__closure`


didimitrie:
Leftover todos: 
- [ ] resolve conflicts
- [ ] review with @gjedlicska 
- [ ] merge 🤞

gjedlicska (approved):

Issue #847, opened 2022-08-01
Linter for Dockerfiles


iainsproat:
**What package are you referring to?**
All

**Is your feature request related to a problem? Please describe.**
Dockerfiles should be well formatted and compliant with standards

**Describe the solution you'd like**
Pre-commit hooks should run Hadolint

**Describe alternatives you've considered**

**Additional context**
Package versions are latest from https://packages.debian.org/search

PR #848, opened 2022-08-01
Add Hadolint to pre-commit

iainsproat:
Fixes https://github.com/specklesystems/speckle-server/issues/847

Uses [Hadolint](https://github.com/hadolint/hadolint), which doesn't work in pre-commit.ci
Integrates pre-commit with husky, to run pre-commit every time the git pre-commit hook is triggered.

If `hadolint` is installed, it will lint Dockerfiles during pre-commit step.
This PR addresses the existing issues in Dockerfiles that were detected by Hadolint.


iainsproat:
## Proof of work
### Hadolint passes
Screenshot 2022-08-05 at 15 05 17

### Hadolint can catch linting issues in Docker file
Screenshot 2022-08-05 at 15 05 08

.husky/pre-commit
```diff
@@ -1,6 +1,15 @@
-#!/bin/sh
+#!/usr/bin/env bash
+set -eo pipefail
 [ -n "$CI" ] && exit 0
 
 . "$(dirname "$0")/_/husky.sh"
 
 yarn lint-staged
```
iainsproat, line 7: @gjedlicska - is `yarn lint-staged` now duplicated by `eslint` in pre-commit?  If so, I can make this run only if the `pre-commit` command is not found.

.pre-commit-config.deployment.yaml
```diff
@@ -0,0 +1,10 @@
+# pre-commit for deployment related resources
```
iainsproat, line 1: Or should this file be renamed `.pre-commit-config.requires-additional-binaries.yaml`, which might be more explicit as to why we have it in a separate file?

Issue #849, opened 2022-08-01
Linter for shell files


iainsproat:
**What package are you referring to?**
All

**Is your feature request related to a problem? Please describe.**
It is difficult to determine if shell files comply to standards by reading them

**Describe the solution you'd like**
shellcheck added as a pre-commit hook

**Describe alternatives you've considered**

**Additional context**

PR #850, opened 2022-08-01
Add shellcheck

iainsproat:
Adds shellcheck to pre-commit.

Fixes shellcheck linting issues.

Uses https://github.com/Jarmos-san/shellcheck-precommit instead of the official https://github.com/koalaman/shellcheck-precommit as the official only provides shellcheck within docker, and not shellcheck on the local system.

Fixes https://github.com/specklesystems/speckle-server/issues/849


iainsproat:
Currently fails as pre-commit.ci cannot run docker images.  If shellcheck is available as a binary, then this hook may work: https://github.com/gruntwork-io/pre-commit/blob/master/hooks/shellcheck.sh

Issue #787, opened 2022-06-02
Build container artifacts for all commits on opened PR's


gjedlicska:
ATM we only build (and publish) docker containers on commits for the main branch.

This can cause issues, when the tests are passing, but some other build error results in a broken state on the main branch after merging a PR.
The only safe and automated fix is to build containers for commits on open PR-s

gjedlicska:
This could also be a nice issue to tackle @iainsproat

PR #856, opened 2022-08-02
build: Images should be built on every branch & commit with a PR

iainsproat:
If a commit on branch `main` or `hotfix*` has a tag which matches `/^[0-9]+\.[0-9]+\.[0-9]+$/` then images are published.

For all other commits on any branch with a PR, images will be built but not published.

Fixes shellcheck warnings.

## Proof of work
Screenshot 2022-08-02 at 16 03 35

## Related issues

Fixes https://github.com/specklesystems/speckle-server/issues/787


.circleci/get_version.sh
```diff
@@ -1,14 +1,20 @@
...
 
-echo "$NEXT_RELEASE-alpha.$CIRCLE_BUILD_NUM"
+if [[ "${CIRCLE_BRANCH}" == "main" ]]; then
+    echo "$NEXT_RELEASE-alpha.${CIRCLE_BUILD_NUM}"
+    exit 0
+fi
+
+BRANCH_NAME_TRUNCATED="$(echo "${CIRCLE_BRANCH}" | cut -c -50)" # docker has a 128 character tag limit, so ensuring the branch name will be short enough
+COMMIT_SHA1_TRUNCATED="$(echo "${CIRCLE_SHA1}" | cut -c -7)"
+echo "$NEXT_RELEASE-branch.${BRANCH_NAME_TRUNCATED}.${COMMIT_SHA1_TRUNCATED}.${CIRCLE_BUILD_NUM}"
```
iainsproat, line 19: Generates a unique version number, even if it isn't on main branch.

.circleci/config.yml
```diff
@@ -8,66 +8,70 @@ workflows:
...
       - get-version:
-          filters: &filters-build
+          filters:
             tags:
-              only: &filters-tag /^[0-9]+\.[0-9]+\.[0-9]+$/
-            branches:
-              only:
-                - main
+              # get version for anything, including any tags
+              only: /.*/
```
iainsproat, line 18: We now get a version for everything.

.circleci/build.sh
```diff
@@ -1,26 +1,30 @@
...
-IMAGE_VERSION_TAG="${IMAGE_VERSION_TAG:-0}"
-echo $IMAGE_VERSION_TAG
+# if there is not image version tag, uses the SHA1 of the last git commit of the branch that triggered this build
+IMAGE_VERSION_TAG="${IMAGE_VERSION_TAG:-${CIRCLE_SHA1}}"
+echo "${IMAGE_VERSION_TAG}"
 
 export DOCKER_BUILDKIT=1
 
-docker build --build-arg SPECKLE_SERVER_VERSION=$IMAGE_VERSION_TAG -t $DOCKER_IMAGE_TAG:latest . -f $FOLDER/$SPECKLE_SERVER_PACKAGE/Dockerfile
+docker build --build-arg SPECKLE_SERVER_VERSION="${IMAGE_VERSION_TAG}" -t "${DOCKER_IMAGE_TAG}:${IMAGE_VERSION_TAG}" . --file "${FOLDER}/${SPECKLE_SERVER_PACKAGE}/Dockerfile"
```
iainsproat, line 18: we should always have an IMAGE_VERSION_TAG, so build with this unique tag.  Only if we're publishing will we tag it with the more generic `latest` and `2`.

gjedlicska:
The overall direction is good, but i think we can do some parts a bit nicer.
Since running the get-version job already persists env vars, we could set the should publish trigger there, and it will be auto available in all other jobs. That way, duplicating the workflow, just to set the publish trigger with a different filter becomes redundant and the publication logic is defined in one place.

Also the long running docker build jobs aren't the ones depending on the server tests, its always the frontend build that takes the most time, so that shouldn't need extra optimizations.

gjedlicska (changes requested):
see my comment on the PR

iainsproat:
Played around with this a little more, I tried two options:

The current commit [`bef24c`](https://github.com/specklesystems/speckle-server/pull/856/commits/bef24c4b328a06903ebe876f58dff26825b38e38) uses logic in a script to determine whether to publish.
Pro:
 - logic for determining whether to publish is contained within a script, so can be more complex in future if necessary
 - the matchers for tags & branches are defined in the yaml
 
Con:
  - logic for publishing to helm is in a different place within the circleci config yaml (could probably tidy
  - not clear from the UI if a job will publish.  'build-and-publish' would suggest it will.
  - the build-and-publish non-frontend jobs have to wait until server-test is complete before starting; though the frontend remains the slowest overall time.
  
A previous iteration (at commit [`0bb1a8`](https://github.com/specklesystems/speckle-server/pull/856/commits/0bb1a8b179b7a59b738ead45b573f044597be92f)), since reverted, separated the build and the publish jobs:
Pro:
 - logic for determining whether to publish is always in the CircleCI config, and can be shared between jobs
 - UI is clear whether it will just build or whether it will also publish (publish will show up as a separate job)
 
Con:
 - storing image as a file to be passed between jobs, results in a large files being passed in the workspace (we may need to be more aggressive in the ttl for workspaces)

.circleci/config.yml
```diff
@@ -111,13 +132,54 @@ jobs:
...
             - env-vars
 
+  should-publish:
+    docker:
+      - image: cimg/base:2022.08
+    working_directory: *work-dir
+    environment:
+      # £ delimited strings of regex for matches which should be published
+      PUBLISHABLE_TAGS: '^[0-9]+\.[0-9]+\.[0-9]+$'
+      PUBLISHABLE_BRANCHES: '^main$£^hotfix.*'
```
iainsproat, line 148: @gjedlicska - not 100% sure about this here

.circleci/should_publish.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+IFS='£' read -a PUB_TAGS <<< "${PUBLISHABLE_TAGS}"
+for item in ${PUB_TAGS[@]}; do
+    [[ "${CIRCLE_TAG}" =~ ${item} ]] && echo "true" && exit 0
+done
+
+IFS='£' read -a PUB_BRANCHES <<< "${PUBLISHABLE_BRANCHES}"
```
iainsproat, line 9: @gjedlicska

.circleci/config.yml
gjedlicska: add alpha branch option

.circleci/should_publish.sh
iainsproat: This publishes on _any_ commit to these branches, or _any_ tag (matching the semver regex).

gjedlicska: haha, good Q. The idea of filtering tags for specific branches is unforch not going to work. If you check out a git tag ( which is what CircleCI does ) the git HEAD is in a detached state, with no connections to its parent branch. This i knew, but forgot, when we were planning this, so sorry about that...

What we can do is have two separate trigger conditions
- publish on tag matching semver stable range
- publish on all commit for branches `main | hotfix.* | alpha.*`

for the branch commits, we rely on the version number properly generated by the get version step

gjedlicska (approved):

Issue #868, opened 2022-08-02
Helm: pods should have named ports


iainsproat:
**What package are you referring to?**
Helm

**Is your feature request related to a problem? Please describe.**
To make referencing of ports simple and help understanding of what is happening in the helm chart, we should provide names for ports.

**Describe the solution you'd like**
All ports on pods should be named, and consuming components should refer to the ports by name.

**Describe alternatives you've considered**

**Additional context**
https://github.com/specklesystems/helm/pull/3/files <- port to this repository

PR #873, opened 2022-08-03
style(helm chart): define named ports on pods and target them in Services

iainsproat:
Named ports improve flexibility and reduce configuration errors

Fix https://github.com/specklesystems/speckle-server/issues/868


iainsproat:
## Proof of work

Named ports are deployed on the pods, in this case `http`, and are used by the corresponding service.

Screenshot 2022-08-03 at 17 00 36
Screenshot 2022-08-03 at 17 01 03

The service is accessible:

Screenshot 2022-08-03 at 17 32 36

gjedlicska (approved):

Issue #822, opened 2022-07-08
Viewer/Frontend: Add Navigation Pop up Help


didimitrie:
Thanks to Jeroen for pointing out the obvious: https://speckle.community/t/web-viewer-controls-and-information/3184

PR #878, opened 2022-08-04
Embed Extras

didimitrie:
**Fixes:** 
- fixed background transparency not being correctly set

**Additions:**
Add "Reply in Speckle" button for comment threads in non slideshow mode (see below) ✅ 

Adds query options for:
- hiding viewer controls ✅ 
- hiding viewer sidebar ✅ 
- hiding selection info ✅ 
- autoloading (useful for full page embeds when we don't want the user to click play) ✅ 
- scroll prevention (useful for embedding in places where we want browser natural scroll to continue) ✅ 
- comment slideshow: adds prev/next buttons that cycle through threads, and disables time display ✅ 

WIPs: 
- [x] add iframe options for some of the above features
- [x] extra tests


packages/frontend/src/main/components/common/SpeckleViewer.vue
```diff
@@ -50,7 +56,11 @@ export default {
     this.$refs.rendererparent.appendChild(this.domElement)
 
     await this.viewer.unloadAll()
-
+    console.log(this.viewer)
+    console.log(this.noScroll)
+    if (this.noScroll) {
+      this.viewer.cameraHandler.controls.mouseButtons.wheel = 0
```
fabis94, line 62: Since the viewer instance is re-used each time the viewer is opened, you should also handle the inverse scenario - a viewer with scrolling turned off being used with scrolling enabled again. So you should explicitly enable scrolling if !noScroll

packages/frontend/src/main/pages/stream/CommitObjectViewer.vue
```diff
@@ -546,6 +572,9 @@ export default defineComponent({
     }, 300)
   },
   methods: {
+    disableScroll() {
+      // TODO
```
fabis94, line 576: There are some unfinished methods and commented out templates in this PR, should they be there?

packages/frontend/src/main/components/viewer/CommentsOverlay.vue
```diff
@@ -484,6 +501,24 @@ export default {
         }
       }
     },
+    nextComment(comment) {
+      const index = this.activeComments.findIndex((c) => c.id === comment.id)
```
fabis94, line 505: Both of these methods are almost identical, you could merge them into a single function that handles all of the scenarios in it (idx === -1, idx === MAX, idx === 0)

packages/frontend/src/main/components/common/SpeckleViewer.vue
didimitrie: Checked, and this is actually difficult to get going properly without deeper hacking with the camera controls lib used in the viewer. I'll add a guard to make sure this is happening only in embeds.

packages/frontend/src/main/components/viewer/CommentsOverlay.vue
didimitrie: done

didimitrie:
ah, forgot to mention this also closes #822 (with some nice gifs)

PR #879, opened 2022-08-04
fix(frontend): various minor FE fixes + gitguardian config

fabis94:
- Fixed Comment Thread Archive modal not appearing
- Fixed Password Too Weak message during registration appearing when it shouldn't
- Added Gitguardian config so that we can ignore the warning about our mixpanel token

### DIFF
--- a/packages/frontend/src/main/pages/auth/TheRegistration.vue
+++ b/packages/frontend/src/main/pages/auth/TheRegistration.vue
@@ -160,7 +160,13 @@
               </v-row>
             </v-col>
             <v-col cols="12">
-              <v-btn block large color="primary" @click="registerUser">
+              <v-btn
+                block
+                large
+                color="primary"
+                :disabled="loading"
+                @click="registerUser"
+              >
                 Create Account
               </v-btn>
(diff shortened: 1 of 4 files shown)


packages/frontend/src/main/pages/auth/TheRegistration.vue
```diff
@@ -296,19 +301,31 @@ export default {
     }
   },
   methods: {
-    debouncedPwdTest: debounce(async function () {
+    async validatePasswordStrength() {
```
gjedlicska, line 304: I think there was reason to debounce this test. Is that not needed anymore?

fabis94: @gjedlicska it didn't work very well, it was often delayed and just confused the user. from now on the check is invoked once you actually press the submit button so that it's clear that the message that appears relates to the last submit. so no, it's not necessary anymore

gjedlicska: blah, makes sense, should have read more carefully

PR #880, opened 2022-08-04
fix(frontend container): add default value for FILE_SIZE_LIMIT_MB

gjedlicska:
this prevents breaking the existing deployments which do not provide this env variable


iainsproat (approved):
Kubernetes will override these, so looks good to me 👍

Issue #864, opened 2022-08-02
Helm: Chart should use recommended labels


iainsproat:
**What package are you referring to?**
Helm Chart

**Is your feature request related to a problem? Please describe.**
Helm defines a set of [recommended labels](https://helm.sh/docs/chart_best_practices/labels/).

**Describe the solution you'd like**
Speckle's Helm Chart should conform to Helm's best practices and provide the recommended annotations.

**Describe alternatives you've considered**

**Additional context**
https://github.com/specklesystems/helm/pull/2 <- should be ported to this repository

PR #882, opened 2022-08-04
feat(helm chart): use standard helm labels on all pods and services

iainsproat:
Helm best practices recommend use of standard labels
https://helm.sh/docs/chart_best_practices/labels/#standard-labels

This PR adds these standard labels to all deployed resources.
This is an additive change that is not breaking.

There is some duplication of helper methods, but this allows future flexibility to move the separate components to their own sub-charts in the future.

Additional changes:
* allows `monitoring` and `test` pod resources to be configurable

fixes https://github.com/specklesystems/speckle-server/issues/864

### DIFF
--- a/utils/helm/speckle-server/templates/_helpers.tpl
+++ b/utils/helm/speckle-server/templates/_helpers.tpl
@@ -1,0 +1,51 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "speckle.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "speckle.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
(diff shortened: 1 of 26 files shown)


iainsproat:
## Proof of work
Screenshot 2022-08-04 at 12 29 38
Screenshot 2022-08-04 at 12 30 11

utils/helm/speckle-server/templates/_helpers.tpl
```diff
@@ -0,0 +1,51 @@
...
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "speckle.fullname" -}}
+{{- if .Values.fullnameOverride }}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
```
gjedlicska, line 17: ohh the lovely whalrus

gjedlicska (commented):
Why do we need to duplicate the _helpers.tpl everywhere with minimal changes? Can't that be generated from a common template?

iainsproat:
@gjedlicska - it's certainly possible, and was the approach I took a month or so ago in this (to be closed) PR. https://github.com/specklesystems/helm/pull/2/files

It reduces the code, as can be seen here: https://github.com/specklesystems/helm/blob/044f5cd48bd1569057f1d7a818ce692ab6b9dec9/charts/speckle-server/templates/_helpers.tpl
However, it leads to complexity in the common function which was difficult to debug and will be a pain to maintain in the future. As an example, these lines: https://github.com/specklesystems/helm/blob/044f5cd48bd1569057f1d7a818ce692ab6b9dec9/charts/speckle-server/templates/_helpers.tpl#L24-L35
Another source of complexity is that it is then not possible to just pass the local context, `.` when using the include function e.g. `include "speckle.name" .`.  Instead it requires the global context, `$` is passed instead e.g. `include "speckle.name" $`.  This isn't typical and might be a source of error for callers of the function in the future.

iainsproat:
Have merged as-is, will aim to consolidate some of the functions and reduce duplication where it makes sense.

PR #890, opened 2022-08-04
build(circleci): Run pre-commit within CircleCI

iainsproat:
## Problem this addresses
When using pre-commit hooks that require docker images or non-typical binaries, these are not supported by pre-commit.ci.

## Solution in this PR
Adds a build step that runs pre-commit within CircleCI.  This allows us to run hooks that require docker, or any other dependency we require.

This solution uses pre-built custom image to provide dependencies, such as pre-commit. The source for this image can be found at https://github.com/specklesystems/pre-commit-runner-image

## Additional changes
* Yarn step to validate CircleCI configuration

## Reference
* Cache invalidation referenced from https://circleci.com/developer/orbs/orb/talkiq/linter


iainsproat:
## Proof of work
### pre-commit.ci
Runs in ~34 seconds
Screenshot 2022-08-04 at 18 11 57

### CircleCI running pre-commit
Runs in ~33 seconds (ignoring installation steps, which can be reduced by caching)
Screenshot 2022-08-04 at 18 12 05

.circleci/config.yml
```diff
@@ -18,6 +19,12 @@ workflows:
               only:
                 - main
 
+      - pre-commit:
+          cache_prefix: '7'
+          filters:
+            tags:
+              only: /.*/
```
gjedlicska, line 26: this should run with the same filters as the test job

PR #896, opened 2022-08-05
Helm: select on standard labels

iainsproat:


iainsproat:
## Evidence that manual intervention during upgrade is required
Screenshot 2022-08-10 at 11 01 32

Issue #883, opened 2022-08-04
Reduce image size and attack surface


iainsproat:
**What package are you referring to?**
Docker images in all packages

**Is your feature request related to a problem? Please describe.**
The images are currently based on Debian's Bullseye Slim distribution.  While this is relatively small, it remains a fairly complete distribution. This may include a number of packages we don't require, which increase the image size and the 'attack surface' (the number of components that need to be monitored, patched, and which could be utilised by a threat actor).

**Describe the solution you'd like**
Use a less feature-complete image which just the minimum amount of dependencies to make Speckle work, perhaps a Distroless based image: https://github.com/GoogleContainerTools/distroless/tree/main/nodejs

**Describe alternatives you've considered**

**Additional context**

PR #899, opened 2022-08-07
perf(server, webhook-service): production images are based on distroless

iainsproat:
As well as improving boot-up & restart time (via smaller download & load size), Distroless reduces the
attack surface area by removing almost all binaries & packages (e.g. shell, chown) that are not
necessary to run node or python.

- speckle-server
  - ✅  moved to distroless in this PR. Before (`v2.7.1`): 352Mb, Now: 203Mb — a **42% reduction**! 📉 
- frontend
  -  Before (`v2.7.1`): 114Mb.
  - 🚫  based on openresty to run nginx.  openresty does not currently provide an official distroless base image.
- preview-service
  - Before (`v2.7.1`): 1195Mb 🐳 
  - ❓ installs a number of binaries via apt-get to run chromium to create previews. Moving to distroless would require manually installing packages (wget and untar, similar to [this](https://github.com/chanseokoh/distroless-with-package-manager/blob/master/Dockerfile)).
- test-deployment
  - ✅  moved to distroless in this PR. Before (`v2.7.1`): 201Mb, Now: 113Mb - a **43% reduction**! 📉 
- monitoring-deployment
  - ✅ moved to distroless in this PR. Before (`v2.7.1`): 142Mb, Now: 69Mb - a **51% reduction**! 📉 
- webhook-service
  - ✅  moved to distroless in this PR. Before (`v2.7.1`): 207Mb, Now: 116Mb — a **44% reduction**! 📉 
- fileimport-service
  - ✅  moved to distroless in this PR. Before (`v2.7.1`): 952Mb, Now: 462Mb - a **51% reduction**! 📉 

Total image size reduced: 891Mb 📉  (28%), but - and most importantly - less security surface area 🔐 

Partially address https://github.com/specklesystems/speckle-server/issues/883


utils/test-deployment/Dockerfile
```diff
@@ -1,6 +1,21 @@
...
+    python3 -m venv /venv
 
-COPY utils/test-deployment .
-RUN pip install -r requirements.txt
+COPY utils/test-deployment/requirements.txt /requirements.txt
+RUN /venv/bin/pip install --disable-pip-version-check --requirement /requirements.txt
 
-CMD [ "./run_tests.py" ]
\ No newline at end of file
+FROM gcr.io/distroless/python3-debian11 as production-stage
```
gjedlicska, line 12: This is considered experimental. We could take the risk, it just has to be a known risk.

image

iainsproat: Yes, I'd noticed that.  I wanted to ensure that the things that run python (e.g. fileimporter) were all working before taking this PR out of draft status.

Slightly more details on why it is marked as experimental: 
> special requirements (like a shell) and another included binary ([ld.so.amd64.cache](https://github.com/GoogleContainerTools/distroless/blob/main/experimental/python3/ld.so.amd64.cache)) which have it sitting in experimental -- we also don't have the current capability to enforce build preconditions on arm machines

https://github.com/GoogleContainerTools/distroless/issues/973#issuecomment-1065169760

Assuming this is everything that makes it experimental; I'd argue that we're currently running python in an image with a shell (and presumably the mentioned binary) so distroless is no worse than our current state.  We're also not running on arm architecture in production (and I assume we don't intend to support that).

packages/fileimport-service/Dockerfile
```diff
@@ -1,30 +1,54 @@
...
+COPY --from=build-stage /speckle-server/wait /wait
+COPY --from=build-stage /speckle-server/tini /tini
+COPY --from=build-stage /venv /venv
+COPY --from=build-stage /speckle-server/node_modules ./node_modules
+
+# HACK
+# the node script calls 'python3', which works when running locally.
+# Force use of the python virtual environment in the image,
+# as this has the pip requirements installed.
+ENV python3="/venv/bin/python3"
```
gjedlicska, line 52: This should use the proper environment variables to set the python environment

gjedlicska: image

The way the python venv activation script does it, is actually prefixing the `PATH` env variable, with the path, where the venv python binary lives. Its a good non destructive solution.
Probably the `VIRTUAL_ENV` variable does not need to be set, lets test without that.

packages/server/Dockerfile
```diff
@@ -1,18 +1,23 @@
...
 ARG NODE_ENV=production
 ARG SPECKLE_SERVER_VERSION=custom
+ARG FILE_SIZE_LIMIT_MB=100
+ARG NODE_ENV=production
 
-RUN apt-get update && \
-  DEBIAN_FRONTEND=noninteractive apt-get install -y \
-  --no-install-recommends \
-  tini=0.19.0-1 \
-  fonts-dejavu-core=2.37-2 \
```
iainsproat, line 9: fonts-dejavu-core & fontconfig have been removed by this PR as this allows us to remove apt as a dependency (tini is installed manually by wget & un-tarring)
@gjedlicska @fabis94 - do you know why they were installed?  If they're required, I can look to manual installation for these as well.
https://packages.debian.org/bullseye/fonts-dejavu-core
https://packages.debian.org/bullseye/fontconfig

PR #900, opened 2022-08-08
WIP: Local deploy kind and skaffold

iainsproat:
⚠️ - currently broken and WIP.
This PR is an attempt to use skaffold to deploy dev builds locally to either kind or minikube ( profiles to be added :) ).  It also deploys a kind cluster locally.


gjedlicska:
What does this add, that the current minikube based local installation doesn't provide (besides not using make to do stuff)?

iainsproat:
@gjedlicska - this PR was mostly me experimenting for now in my spare time, you can ignore it for the meantime!

If it was to be merged, I'd split it into two PRs first - one for skaffold and one for kind.

- in theory kind should have a faster startup time over minikube (will have to experiment to verify this in the context of our CI), which would bring down our CI times (especially for PRs, where `server-test` is the longest running job).

- skaffold would replace the Makefile.  It deals with ensuring the deployed helm chart matches up with the correct docker image (a very slight improvement over using a mutable `local` tag).  It may make port-forwarding easier, and it also provides a filewatcher which would update the deployed helm chart & images while working on code.

PR #903, opened 2022-08-09
feat(server): new base email template + implementation for invites emails

### LINKED ISSUE #853: Reusable server email templating
Create a maintainable solution to work with email templates on the server.

The current string literals solution is messy and hard to maintain.
The new solution should give us an easy to use functionality, that provides a consistent look and feel for multiple mail contents.

fabis94:
closes #853 

- Using EJS templates to render .txt and .html templates, cause lodash/template was too limitting (i had issues with unnecessary newlines being added in the .txt file)
- HTML template built in https://github.com/specklesystems/speckle-email-templates


packages/server/assets/emails/templates/basic/basic.html
```diff
@@ -0,0 +1,566 @@
...
+                            role="presentation"
+                            style="border-collapse: collapse; border-spacing: 0px"
+                          >
+                            <tbody>
+                              <tr>
+                                <td style="width: 100px">
+                                  <img
+                                    alt="Speckle"
+                                    height="auto"
+                                    src="https://i.imgur.com/KAvbGEj.png"
```
fabis94, line 180: @gjedlicska @didimitrie FYI the logo is currently loaded from imgur, since we don't currently have a way to just dump static assets for public serving anywhere inside `speckle-server`

packages/server/bootstrap.js
```diff
@@ -3,17 +3,17 @@
  * Bootstrap module that should be imported at the very top of each entry point module
  */
 
-// Conditionally change appRoot and repoRoot according to whether we're running from /dist/ or not (ts-node)
+// Conditionally change appRoot and packageRoot according to whether we're running from /dist/ or not (ts-node)
 const path = require('path')
 const isTsNode = !!process[Symbol.for('ts-node.register.instance')]
 const appRoot = __dirname
-const repoRoot = isTsNode ? appRoot : path.resolve(__dirname, '../')
+const packageRoot = isTsNode ? appRoot : path.resolve(__dirname, '../')
```
fabis94, line 10: Previous name was misleading

packages/server/modules/core/dbSchema.ts
```diff
@@ -0,0 +1,168 @@
...
+  used?: boolean
+  message?: string
+  resourceTarget?: string
+  resourceId?: string
+  role?: string
+  token: string
+}
+
+/*
+ * TABLE HELPERS
```
fabis94, line 91: Basically just re-did these with TypeScript to get vastly improved DX with a lot less code - to define a new helper you just have to invoke `buildTableHelper` and pass in the table and column names

didimitrie (changes requested):
Looking good. One last (hopefully quick) addition would be to give same treatment to the password reset email! We'd be leaving the job half done (on the end user "visible" side) without that extra last step :) 

For copy on the pwd reset, I've added it as an example template in the template repo (see: https://github.com/specklesystems/speckle-email-templates/pull/1)

PR #907, opened 2022-08-10
refactor(server authz): refactor authz module to TypeScript

gjedlicska:


packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
+import { Scopes } from '@/modules/core/helpers/mainConstants'
+import { getRoles } from '@/modules/shared'
+import { getStream } from '@/modules/core/services/streams'
+
+import {
+  BaseError,
+  ForbiddenError as FE,
```
fabis94, line 12: this was probably already here before, but these shortened error names are really unintuitive. i'm guessing originally there were other errors with the same name imported here? there don't seem to be any more, tho, so maybe the aliases can be removed?

or you could do something like
```js
import * as Errors from '@/modules/shared/errors'
...
throw new Errors.ForbiddenError()
```

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+}
+
+interface AuthFailedResult extends AuthResult {
+  authorized: false
+  error: BaseError | null
+  fatal?: boolean
+}
+
+interface Stream {
+  role?: StreamRoles
```
fabis94, line 30: Better to keep DB entity types in some centralized place so that they can be re-used everywhere (and they def will be).

I've kept them (TS types or JSDoc Typedefs) in these places before:
- In the entity's repository module (e.g. `ServerInviteRecord` in `modules/serverinvites/repositories/index.js`)
- In the relevant speckle module's types helper (in this case - `@/modules/core/helpers/type.ts`

Additionally - you shouldn't mark fields in these kinds of types as optional (with ?), cause the DB schema being a strictly defined relational schema - all of the fields will always be there. Sure they might be null, then you can mark the value as Nullable<T>, but they can't be optional or undefined.

You might say that you might have a more limited query that only pulls specific fields from the DB, but in that case - wouldn't all of the fields be optional? Any and all fields can be omitted from a query, so theoretically they'd all have to be optional then.

The way I suggest dealing with this is making sure that this interface defines all of the fields with none of them being optional and with values correctly marked as Nullable, when it's possible, and then if you ever need to have a more limited version of this time you can use the `Pick`, `Omit` or `Partial` TS utility types to specify that some of the fields don't have to be there.

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+  isPublic: boolean
+  allowPublicComments: boolean
+}
+
+interface AuthContext {
```
fabis94, line 35: maybe some other ones as well, but I know that this AuthContext type already has a JSDoc typedef in `modules/shared/index.js` (AuthContextPart)

to make things more DRY, you should delete the JSDoc Typedef, and use this new TS type instead. you can import & use TS types in JSDoc comments as well, through the dynamic import syntax.

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+} from '@/modules/shared/errors'
+// import { getbAllRoles } from '../core/services/generic'
+
+interface AuthResult {
+  authorized: boolean
+}
+
+interface AuthFailedResult extends AuthResult {
+  authorized: false
+  error: BaseError | null
```
fabis94, line 25: You can do `Nullable<BaseError>` as well, but that's up to you

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+  User = 'server:user',
+  ArchivedUser = 'server:archived-user'
+}
+
+enum StreamRoles {
+  Owner = 'stream:owner',
+  Contributor = 'stream:contributor',
+  Reviewer = 'stream:reviewer'
+}
+type AvailableRoles = ServerRoles | StreamRoles
```
fabis94, line 71: Not great that ServerRoles & StreamRoles essentially duplicate everything that we already have in `mainConstants.js`.

Not just that, but in TS the actual literal string value of an enum does not equal a string literal with that same value -  `ServerRoles.Admin !== 'server:admin'` in the eyes of TypeScript. So if you predict wanting to use a string literal from outside the enum (e.g. `Roles` imported from `mainConstants.js`), you should avoid typing the constant values with enums.

I've solved both of those problems by converting `mainConstants.js` to TS, and extracting ` ServerRoles` and `StreamRoles` types from the already defined constants there. Instead of being enum values, they will just be a union type of all of the allowed role values. This results in a less duplicated codebase and also in being able to use the `Roles` constant values wherever you expect `StreamRoles` or `ServerRoles`.

(I've pushed these changes into your branch)

image

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+      return authFailed(context, new FE('You do not have the required privileges.'))
+    if (
+      context.scopes.indexOf(requiredScope) === -1 &&
+      context.scopes.indexOf('*') === -1
+    )
+      return authFailed(context, new FE('You do not have the required privileges.'))
+    return authSuccess(context)
+  }
+
+type StreamGetter = ({
```
fabis94, line 163: you don't need to destructure the object twice here

```ts
type StreamGetter = (params: { streamId: string; userId?: string }) => Promise<Stream>
```

packages/server/modules/shared/authz.ts
```diff
@@ -0,0 +1,293 @@
...
+    }
+    // validate auth result a bit...
+    if (authResult.authorized && authHasFailed(authResult))
+      throw new Error('Auth failure')
+    return { context, authResult }
+  }
+  return pipeline
+}
+
+import Express from 'express'
```
fabis94, line 239: In ESM all imports should always be at the top, you can only invoke an import at other arbitrary locations if they're dynamic/async imports. During compilation I bet the import gets moved up there anyway, but to keep the source code more readable it would be nice to move it up there as well.

packages/server/modules/shared/errors/base.ts
```diff
@@ -17,11 +17,7 @@ class BaseError extends VError {
    */
   static defaultMessage = 'Unexpected error occurred!'
 
-  /**
-   * @param {string | null} message
-   * @param {import('verror').Options | Error} options
-   */
-  constructor(message, options) {
+  constructor(message: string | null, options: Options | Error | null = null) {
```
fabis94, line 20:
```suggestion
  constructor(message: string | null | undefined, options: Options | Error | undefined = undefined) {
```

fabis94 (changes requested):

packages/server/modules/shared/authz.ts
fabis94: example of the ServerRoles.admin !== 'server:admin` issue:
image

fabis94: there was an inverse solution btw:
1. define ServerRoles & StreamRoles as enums
2. update the `Roles` constant and change all of its string literal values to enum values

this way you won't get TS errors when using a value off `Roles` in a variable/param that expects the `StreamRole` or `ServerRole` enum. if you would ever use a string literal directly, however, it wouldn't work

packages/server/modules/shared/authz.ts
gjedlicska: yeah, it was some name duplication. I'm not a fan of * imports, but we can def drop the aliases

packages/server/modules/shared/authz.ts
gjedlicska: yeah, that is true, but since duck typing I thought its fine. I sort of like the strictly necessary interface definition.
Plus this doesn't match the db record. The role? is there, because that is dynamically patched to the db record by the `getStream` function from `@/modules/core/repositories/streams.js`

Anyhow, we can start to define these types somewhere, my reasoning here was, to define the least amount needed for the auth context

fabis94: BTW In my own branch I've already created a `core` types helper at `@/modules/core/helpers/types.ts`, and I already defined a `StreamRecord` type there, so I guess the best path forward would be for you to either do the same and solve the merge conflict afterwards, or leave your Stream type as local to your module

fabis94: > yeah, that is true, but since duck typing I thought its fine. I sort of like the strictly necessary interface definition. Plus this doesn't match the db record. The role? is there, because that is dynamically patched to the db record by the `getStream` function from `@/modules/core/repositories/streams.js`
> 
> Anyhow, we can start to define these types somewhere, my reasoning here was, to define the least amount needed for the auth context

sure, if it's something very specific to your own module and not really reusable, then you can keep it there. considering the other comment I just made, it might be the best option

fabis94 (approved):

Issue #860, opened 2022-08-02
Helm: network policies


iainsproat:
**What package are you referring to?**
Helm

**Is your feature request related to a problem? Please describe.**
At present network traffic is allowed from anywhere to everywhere, and vice-versa, both within and without the kubernetes cluster.

**Describe the solution you'd like**
For additional security, [NetworkPolicies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) should be defined for each service.

Create NetworkPolicies using [an editor](https://editor.cilium.io/).

**Describe alternatives you've considered**

**Additional context**
* This is limited as Kubernetes Network Policy does not allow egress definitions with domain names (FQDN) and expects IPs.  For any service where we don't know the IP behind the FQDN (or it's likely to change) we have to allow egress to 0.0.0.0/0 (i.e. everything external).  Cilium Network Policies do allow egress by FQDN, but the presence of Cilium cannot be assumed for all deployments.

PR #909, opened 2022-08-10
feat(helm chart): network policies are provided for all services

iainsproat:
## Description & motivation

[Network policies](https://kubernetes.io/docs/concepts/services-networking/network-policies/) are used to deny arbitrary network egress and ingress to a pod.  They effectively act as a simple 'firewall' around each service.   This provides some additional security at the kubernetes network-mesh level, for example preventing an attacker connecting to other pods within the cluster or exfiltrating data to arbitrary locations.

Fixes https://github.com/specklesystems/speckle-server/issues/860

## Changes:

Frontend:
- only allow ingress to port 80 from within the cluster (which should be the ingress loadbalancer)
- block all egress

Backend:
- only allow ingress to port 3000 from within the cluster (which should be the ingress loadbalancer)
- explicitly allow ingress from test and file import service.
- allow egress to redis, postgres, & blob storage

Preview service
- only allow ingress from the monitoring stack to the monitoring port
- allow egress to postgres

Monitoring
- only allow ingress from the monitoring stack to the monitoring port
- allow egress to postgres

File Import Service
- only allow ingress from the monitoring stack to the monitoring port
- allow egress to postgres
- allow egress to the speckle-server (backend) port 3000

Test
- block all ingress
- only allow egress to the speckle-server (backend) port 3000

Webhook service
- only allow ingress from the monitoring stack to the monitoring port
- only allow egress to outside of the cluster.  This assumes that applications called by the webhook service are not within the cluster (or are accessible via an ingress from the internet).  This prevents the webhook service being abused to call other kubernetes services within the cluster.

Screenshot 2022-08-15 at 10 47 54

_diagram of the configured ingress (red/purple) and egress (greens), and both ingress/egress (black)_

## To-do before merge

No special requirements

## Screenshots:

See comment below.

## Validation of changes:

See comment below.

## Things that aren't perfect

- Kubernetes Network Policies don't support domain names, only IP addresses.  Hence if an endpoint (e.g. redis or postgres) is a domain name, and we don't know the IP address will be static (very likely), we have to allow egress to everywhere 0.0.0.0/0.  Separate issue to address this for Cilium network meshes at some point in the future: https://github.com/specklesystems/speckle-server/issues/913
- Kubernetes Network Policies only support IP v4 CIDRs.
- I don't think it's possible to explicitly select on the nginx ingress, so ingress to the frontend and server is allowed from anywhere in the cluster.

## Checklist:

- [x] I have added appropriate tests. - was deployed to a cluster, see screenshots below.


iainsproat:
## Proof of work

- No policies are deployed when `.values.*.networkPolicy.enabled` is `false`:
Screenshot 2022-08-10 at 15 56 47

- Policies are deployed when `.values.*.networkPolicy.enabled` is `true` (default):
Screenshot 2022-08-10 at 16 09 38

utils/helm/speckle-server/templates/fileimport_service/networkpolicy.yml
```diff
@@ -0,0 +1,47 @@
...
+      ports:
+        - port: 53
+          protocol: UDP
+    # allow egress to speckle-server
+    - to:
+        - podSelector:
+            matchLabels:
+{{ include "server.selectorLabels" $ | indent 16 }}
+      ports:
+        - port: 3000
```
iainsproat, line 40: Tried using named port `http`, but this wasn't working 🤔

utils/helm/speckle-server/templates/_helpers.tpl
```diff
@@ -93,3 +93,137 @@ Part-of label
...
+Creates a network policy egress definition for connecting to Postgres
+*/}}
+{{- define "speckle.networkpolicy.egress.blob_storage" -}}
+{{- $port := (default "443" .Values.s3.networkPolicy.port ) -}}
+{{- if .Values.s3.networkPolicy.inCluster.enabled -}}
+{{ include "speckle.networkpolicy.egress.internal" (dict "podSelector" .Values.s3.networkPolicy.inCluster.podSelector "namespaceSelector" .Values.s3.networkPolicy.inCluster.namespaceSelector "port" $port) }}
+{{- else if .Values.s3.networkPolicy.externalToCluster.enabled -}}
+  {{- $host := (urlParse .Values.s3.endpoint).host -}}
+  {{- $ip := "" -}}
+  {{- if eq (include "speckle.isIPv4" $host) "true" -}}
```
iainsproat, line 141: Struggled to have the include return a boolean (i.e. not a string) and avoid using `eq` against a string value of "true".  Kept failing for some reason I was unable to debug.

gjedlicska:
2 things are missing I think:
- [ ] allow egress to sentry from those services which report stuff (just the server afaik)
- [ ] allow egress to mixpanel for metrics

gjedlicska:
what's the difference between the colors of egress lines?

iainsproat:
@gjedlicska - the shades of green are just to make it easier to trace the line.
greens are egress only. purples are ingress only, and black is both ingress/egress on the respective service (e.g. egress for fileimport, ingress for server).

iainsproat:
@gjedlicska - sentry_dns value was only ingested by the server, so I assume that is the only service that was using it.

Mixpanel is only referenced in the frontend code, and my understanding is that mixpanel runs in the user's browser.  i.e. there's no need for egress from frontend.

Fixed the services with unnecessary access to s3 and redis.

Diagram is updated.

gjedlicska (approved):

PR #910, opened 2022-08-10
Allow save object to S3 in different region

peterjgrainger:
Can we add the ability to save objects to different regions rather than hardcode it? Some projects require data to be stored in a particular country.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -248,6 +248,7 @@ jobs:
       S3_SECRET_KEY: 'minioadmin'
       S3_BUCKET: 'speckle-server'
       S3_CREATE_BUCKET: 'true'
+      S3_REGION: '' # optional, defaults to 'us-east-1'
     steps:
       - checkout
       - restore_cache:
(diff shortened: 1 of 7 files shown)


iainsproat:
Thanks for the contribution @peterjgrainger !
Would you have time to add the environment variable to the docker-compose and the helm charts?  (If not, let me know and  I can do that if you're ok with me collaborating here and contributing to your PR!)

peterjgrainger:
`docker-compose` should be OK right? It's defaulting to the same value as before plus it's using a local version of S3. For the helm charts I haven't much experience so please contribute @iainsproat !

iainsproat:
Hi @peterjgrainger - thanks for letting me contribute!

I was keen to add the variable to docker-compose and helm chart as this will help others by documenting that the variable is available to be configured.

I've pushed a commit to your fork.  Please review and let me know if you're happy with my changes.

peterjgrainger:
Changes look good. What about the .env.example file? Should I change that too?

iainsproat:
@peterjgrainger - yes, please!

gjedlicska (approved):

PR #912, opened 2022-08-11
Adds gitguardian ggshield pre-commit step

iainsproat:
* Intended to be run locally and not in CI
* currently incompatible with how we intend to run CI

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,5 +25,12 @@ repos:
     rev: 'v2.7.1' # Use the sha / tag you want to point at
     hooks:
       - id: prettier
+
+  - repo: https://github.com/gitguardian/ggshield
+    rev: 'v1.13.0'
+    hooks:
+      - id: ggshield
+        language_version: python3
+        stages: [commit]
 ci:
   autoupdate_schedule: quarterly


.gitguardian.yml
```diff
@@ -1,3 +1,7 @@
-matches-ignore:
-  - name: MIXPANEL_TOKEN
-    match: acd87c5a50b56df91a795e999812a3a4
+secret:
+  ignored-paths:
+    - 'packages/server/modules/emails/tests'
+  ignored-matches:
+    - match: acd87c5a50b56df91a795e999812a3a4
```
fabis94, line 5: so verbose! can't believe you have to repeat the token twice and even refer to the full path to file...

anyway, I'm guessing that's not something we can do much about, but the issue that can be resolved is the fact that there's an absolute path to `bootstrapper.ts` from your own workstation, which isn't going to work for anyone else. are relative paths supported?

fabis94: @iainsproat is the 'name' key just a display name/title for the rule? cause if so, we can make it shorter and more descriptive at the same time by just setting it to "mixpanel token". the token being put in the 'name' key doesn't help (we can already see it from the 'match' key) and the file name doesn't give us anything either cause we can always just do a full text search in the project to find where the token is used. it doesn't tell us what kind of secret/token it is either.

fabis94 (approved):

Issue #859, opened 2022-08-02
Helm: the service accounts should have minimum privileges for the service


iainsproat:
**What package are you referring to?**
Helm for kubernetes

**Is your feature request related to a problem? Please describe.**
At present the default service account is used; this may provide too may privileges to each service.

**Describe the solution you'd like**
A Service Account with minimum privileges should be provided.  If some services require more privileges, they should have distinct service accounts.
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
Privileges should be defined by RBAC (Role & RoleBinding resources).
Pod Security Admissions https://kubernetes.io/docs/concepts/security/pod-security-admission/

**Describe alternatives you've considered**

**Additional context**

PR #922, opened 2022-08-12
feat(helm chart): serviceAccounts are provided for each service

iainsproat:
ServiceAccounts are provided for each service.

These ServiceAccounts do not mount service account token (which allows access to the kubernetes API).
The ServiceAccount only allows mounting of the secret that the pod requires access to.  i.e. it blocks mounting of arbitrary secrets.  For test and frontend, no secrets need to be mounted.

Fixes https://github.com/specklesystems/speckle-server/issues/859

### DIFF
--- a/utils/helm/speckle-server/templates/fileimport_service/deployment.yml
+++ b/utils/helm/speckle-server/templates/fileimport_service/deployment.yml
@@ -85,14 +85,13 @@ spec:
             value: {{ .Values.fileimport_service.time_limit_min | quote }}
 
       priorityClassName: low-priority
-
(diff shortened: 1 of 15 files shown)


iainsproat:
## Proof of work

If the service account does not list the required secret then the pod is not created (in this example speckle-monitoring is missing).  i.e. it is not possible to create a pod which is requesting secrets its service account does not have access to!
Screenshot 2022-08-12 at 18 52 55

With the service account correctly configured to provide the secret requested by the pod, the pod can be created:
Screenshot 2022-08-12 at 18 53 37

PR #933, opened 2022-08-16
feat(server): notify users on comment mentions

### LINKED ISSUE #904: Basic Notifications API + Comment mentions emails
- Basic Notifications API that relies on Redis PubSub, as discussed with Gergo
- Comment mentions emails as the first feature that uses this API

fabis94:
closes #904 

Features:
- Notifications on comment mention: When a user is mentioned in a comment, they should receive an e-mail about this

Notable changes
- Notifications module introduced which allows us to queue up notifications in a [Bull](https://github.com/OptimalBits/bull) queue and process them asynchronously (currently in the same server process, in the future possibly in separate worker processes)
- Introduced optional `shutdown` function for `server` modules, which allow you to do some cleanup once the server has been signaled to exit
- Useful CLI commands for debugging the notifications queue accessible at `yarn cli bull`. Push stuff onto the queue, run a consumer or run the Bull Web UI for monitoring the queues.
- Introduced a sort of event-based architecture for the `comments` module, which allows other modules from within that same process to listen for events and respond to them asynchronously. The point is to make the architecture more [SOLID](https://en.wikipedia.org/wiki/SOLID) - Instead of modifying the `createComment` service call every time we want to do something once a comment is created, the `createComment` call emits a "Comment created" event that can be handled by other modules elsewhere
- Typed many DB schema tables. Whenever you're querying for streams, comments, ACL stuff - you can use type aliases defined in this PR to type the results.


packages/server/modules/comments/events/emitter.ts
```diff
@@ -0,0 +1,69 @@
+import { CommentRecord } from '@/modules/comments/helpers/types'
+import { MaybeAsync } from '@/modules/shared/helpers/typeHelper'
+import { modulesDebug } from '@/modules/shared/utils/logger'
+import { EventEmitter } from 'node:events'
+
```
fabis94, line 5: @gjedlicska Besides the Redis/Bull based message queue, I've also introduced the approach of having a speckle-module-scoped (e.g. "comments" module) event emitter to allow other modules to listen for and handle certain events within that same process. Don't confuse this with the inter-process event queue we discussed previously, this only works with events fired within that same process and solves a different issue - making our code more [SOLID](https://en.wikipedia.org/wiki/SOLID).

In this PR I wanted to trigger an activity (notification) when a comment is created. While I could have just added this code to the `createComment` service call, it wouldn't be very SOLID, because it would add extra responsibilities to the call and also break the open-closed principle by requiring that devs actually modify this function to extend it.

So now, instead of modifying the `createComment` function every time we need to do something when a comment is created, you can just listen for the "COMMENT CREATED" event in some other module of yours and respond accordingly (e.g. see if it has any mentions and send out an email).

Here's an example of how this emitter is listened to: `modules/comments/services/notifications.ts`

I think this may be a useful approach to explore further as that would make our modules less coupled with each other and possibly allow custom/third-party speckle modules in the future that could use these events to do custom things.

packages/server/modules/notifications/services/queue.ts
```diff
@@ -0,0 +1,173 @@
...
+          limiter: {
+            max: 10,
+            duration: 1000
+          }
+        }
+      : {}),
+    defaultJobOptions: {
+      attempts: 1,
+      timeout: 10 * 1000, // 10s execution timeout
+      removeOnComplete: isProdEnv(),
```
fabis94, line 54: @gjedlicska all finished jobs are removed from the queue in prod so as to not fill it up, but that means we can't check them out afterwards as some kind of log output. if we need more insight into what's going on in the queues, we can re-think this

packages/server/modules/notifications/services/queue.ts
```diff
@@ -0,0 +1,173 @@
...
+  Success = 'success',
+  ValidationError = 'validation-error'
+}
+
+const handlers = new Map<NotificationType, NotificationHandler>()
+
+export const NOTIFICATIONS_QUEUE_MAIN = `default:user-notifications`
+export const NOTIFICATIONS_QUEUE_TEST = `test:user-notifications`
+
+export const NOTIFICATIONS_QUEUE = isTestEnv()
```
fabis94, line 35: separate queues for main app & tests

packages/server/test/notificationsHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+
+type AckEvent = {
+  result?: NotificationJobResult
+  err?: Error
+  jobId: JobId
+}
+
+const NEW_ACK_EVENT = 'new-ack'
+
+export async function buildNotificationsStateTracker() {
```
fabis94, line 14: Utility you can use in tests to await when certain notifications finish executing

packages/server/modules/comments/events/emitter.ts
```diff
@@ -0,0 +1,69 @@
...
+export function onCommentEvent<T extends CommentsEvents>(
+  eventName: T,
+  handler: CommentsEventHandler<T>
+) {
+  const onHandler = (event: CommentsEventPayloadMap[T]) => {
+    // This is important: Without this errors in async event handlers will
+    // kill the process
+    try {
+      // In case handler is async
+      Promise.resolve(handler(event)).catch(errHandler)
```
fabis94, line 57: as mentioned above, currently there's no way for these event handlers to throw without the error being logged and suppressed, just so that you don't have a situation where a down-stream handle throws and breaks the part of the code that emitted the event

we can re-think this however

gjedlicska (changes requested):
One thing is def missing, the mention user event needs to be saved in the activities module, to be able to query it later. Call this [func](https://github.com/specklesystems/speckle-server/blob/main/packages/server/modules/activitystream/services/index.js#L32) with the appropriate data

packages/server/modules/activitystream/services/streamActivityService.ts
```diff
@@ -137,10 +143,29 @@ async function addStreamInviteDeclinedActivity({
...
+  commentId: string
+  threadId: string
+}) {
+  const { streamId, mentionAuthorId, mentionTargetId, commentId, threadId } = params
+  await saveActivity({
+    streamId,
+    resourceType: ResourceTypes.Comment,
+    resourceId: commentId,
+    actionType: ActionTypes.Comment.Mention,
+    userId: mentionTargetId,
```
gjedlicska, line 162: This should be the mentionAuthorId, we're recording the activities of the user who made the activity.

packages/server/modules/comments/services/notifications.ts
```diff
@@ -0,0 +1,101 @@
...
+            commentId: id,
+            threadId
+          })
+        ]
+      })
+    )
+  )
+}
+
+async function processNewCommentMentions(newComment: CommentRecord) {
```
gjedlicska, line 67: This could be merged into the function as below, with an optional previousComment argument that defaults to an empty list of previous mentions.

packages/server/modules/notifications/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    ).default
+  }
+
+  registerNotificationHandlers(
+    typeWhitelist?.length ? pick(allHandlers, typeWhitelist) : allHandlers
+  )
+
+  initializeQueue()
+
+  if (isTestChildServer()) {
```
gjedlicska, line 34: Baking in test specific code into the production codebase is always itchy. Can we work around this somehow?

packages/server/modules/notifications/helpers/types.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { MaybeAsync, Optional } from '@/modules/shared/helpers/typeHelper'
+import { Job } from 'bull'
+import debug from 'debug'
+import { isObject, has } from 'lodash'
+
+export enum NotificationType {
+  MentionedInComment = 'mentioned-in-comment',
+  Test = 'test'
```
gjedlicska, line 9: Can we not add the `Test` case in the production code? Maybe make this an extendable type and extend in the tests.

packages/server/modules/shared/helpers/typeHelper.ts
```diff
@@ -2,9 +2,29 @@ import { Express } from 'express'
...
 export type SpeckleModule = {
-  init: (app: Express) => MaybeAsync<void>
-  finalize: (app: Express) => MaybeAsync<void>
+  /**
+   * Initialize the module
+   * @param app The Express instance
+   * @param isInitial Whether this initialization method is being invoked for the first time in this
+   * process. In tests modules can be initialized multiple times.
+   */
+  init: (app: Express, isInitial: boolean) => MaybeAsync<void>
```
gjedlicska, line 16: Why are these functions MaybeAsync? Doesn't it complicate our lives?

fabis94: @gjedlicska I don't think so, if you look inside ./modules/index.js you'll see that both init() & finalize() are awaited, so they already support being async. And if init() is synchronous that will work also, the `await` will be essentially ignored.

So it appears that already before this PR these functions supported both async and non-async return types.

packages/server/modules/activitystream/services/streamActivityService.ts
fabis94: good catch, i'll fix it

packages/server/modules/comments/services/notifications.ts
fabis94: true, i'll merge em

packages/server/modules/activitystream/services/streamActivityService.ts
fabis94: fixed

packages/server/modules/notifications/index.ts
fabis94: there's no way to get around having some sort of check here, but I renamed the env var to something less test specific - DISABLE_NOTIFICATIONS_CONSUMPTION

packages/server/modules/notifications/helpers/types.ts
fabis94: got rid of it everywhere. this does make the test-push and test-consume commands of the CLI a bit confusing, because I'm having to use the real "Mentioned in Comment" notification type to test out pushing & consuming, while overriding its handler to be a "test" handler

PR #956, opened 2022-08-23
feature(frontend): dl manager link now leads to releases page

### LINKED ISSUE #953: Update Manager download link
The link to download Manager in the frontend should be updated to point to the releases page: 

https://releases.speckle.systems/

Instead of automatically downloading the Windows version:

https://github.com/specklesystems/speckle-server/blob/b55f12d6bcb8b5a04a99691aab3aad02ff02e22e/packages/frontend/src/main/components/common/NoDataPlaceholder.vue#L29

fabis94:
closes #953

### DIFF
--- a/packages/frontend/src/main/components/common/NoDataPlaceholder.vue
+++ b/packages/frontend/src/main/components/common/NoDataPlaceholder.vue
@@ -26,7 +26,8 @@
                   link
                   class="primary mb-4"
                   dark
-                  @click="downloadManager"
+                  href="https://releases.speckle.systems/"
+                  target="_blank"
                 >
                   <v-list-item-icon>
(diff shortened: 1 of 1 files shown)


packages/frontend/src/main/components/common/NoDataPlaceholder.vue
```diff
@@ -190,13 +190,14 @@ export default {
         type: 'action'
       })
```
teocomi, line 191: This ` this.$mixpanel.track` can be removed, sorry I should have mentioned it, didn't think you were going to reuse this method ;)
Could actually opt for a `:href` on the button directly if you prefer ;)

teocomi (changes requested):
Minor fiddles

PR #967, opened 2022-08-25
refactor: email verification flow & email template

### LINKED ISSUE #826: Emails Redesign
Emails sent by the server currently look unprofessional, or their message is difficult to read. 

General: 
- [x] shared template with header (image) and footer

Content reworks, bearing the following: email should have a clear **big blue button** as the main call to action
- [x] content rework stream invites
- [x] content rework server invites
- [x] content rework password reset
- [x] email verification
- [ ] others?

### LINKED ISSUE #958: Email Verification Improvements
- [x] fire off email verification on user signup automatically 
- [x] allow infinite verification emails to be sent
- [x] top banner behaviour change: 
if a verification "email" is already present for the user, "please check your inbox for your verification email. [resend button]"
if a verification "email" is  not already present for the user, "please verify your email [send email verification]"

fabis94:
closes #958 
closes #826


packages/server/modules/emails/migrations/20220825082631_drop_email_verifications_used_col.ts
```diff
@@ -0,0 +1,23 @@
+import { Knex } from 'knex'
+
+const TABLE_NAME = 'email_verifications'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TABLE_NAME, (table) => {
```
fabis94, line 6: dropped unused 'used' column and migrated timestamp precision

packages/frontend/src/main/layouts/TheMain.vue
```diff
@@ -62,9 +61,9 @@
 </template>
 <script>
 import { gql } from '@apollo/client/core'
-import { mainUserDataQuery } from '@/graphql/user'
 import { useNavigationDrawerAutoResize } from '../lib/core/composables/dom'
 import { ref } from 'vue'
+import { useIsLoggedIn } from '../lib/core/composables/core'
```
iainsproat, line 66: is this a typo? - `user` or `use`?

fabis94: "use", it's a common prefix for composables https://vuejs.org/guide/extras/composition-api-faq.html

packages/server/modules/core/services/users.js
```diff
@@ -73,25 +73,17 @@ module.exports = {
     const usr = await userByEmailQuery(user.email).select('id').first()
     if (usr) throw new Error('Email taken. Try logging in?')
 
-    const res = await Users().returning('id').insert(user)
+    await Users().insert(user)
+    const newUser = await module.exports.getUserById({ userId: newId })
```
gjedlicska, line 77: Why do we need to query back the user? It feels like we do not trust, that if the line above succeeds, we have the user in the DB...
The users.insert should return a failure if it cannot create the user...

packages/server/test/mockHelper.ts
```diff
@@ -0,0 +1,222 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/no-var-requires */
+import { MaybeAsync } from '@/modules/shared/helpers/typeHelper'
+import { isArray, isFunction } from 'lodash'
+import mock from 'mock-require'
+import { ConditionalPick } from 'type-fest'
+
+export type MockedFunctionImplementation = (...args: any[]) => MaybeAsync<any>
```
gjedlicska, line 8: I hate mocking...

fabis94: Yea, I don't really see any problem with mocking/faking per se, cause you inevitably need to set a fake boundary when it comes to integration tests. It's all about how the mocking is set up and used (like you once said - mock only what you own), and this kind of JS module mocking is standard practice in the JS ecosystem

packages/server/modules/core/services/users.js
fabis94: fixed https://github.com/specklesystems/speckle-server/pull/967/commits/f003270b5b6b5aa2e2fa113b19a1714eaaf67e16

PR #972, opened 2022-08-29
refactor(server): load secrets from file instead of env var

### LINKED ISSUE #897: Helm: mount secrets as files instead of env vars
Helm Chart and all services

https://www.tenable.com/audits/items/CIS_Kubernetes_v1.20_v1.0.0_Level_2_Master.audit:98de3da69271994afb6211cf86ae4c6b

Secrets should be mounted as files instead of environment variables.

**Describe alternatives you've considered**

**Additional context**

iainsproat:
## Description & motivation

Security recommendations for Kubernetes are to mount secrets from file instead of env vars.  Env
vars are available to all users of a container, often get leaked in log output etc..  If file does
not exist, secrets are loaded from env var for backward compatibility.

fix #897

## Changes:

- direct call to environment variable for secret is replaced with a helper method, this attempts to retrieve the secret from a file and falls back to environment variable.

## Checklist:

### DIFF
--- a/packages/server/modules/shared/helpers/secretsHelper.ts
+++ b/packages/server/modules/shared/helpers/secretsHelper.ts
@@ -1,0 +1,74 @@
+import { MisconfiguredSecretError } from '@/modules/shared/errors'
+import { existsSync, readFile } from 'fs'
+import { join } from 'path'
+
+let redisUrl: string | undefined
+let postgresUrl: string | undefined
+let sessionSecret: string | undefined
+let emailPassword: string | undefined
+let s3SecretKey: string | undefined
+let googleClientSecret: string | undefined
+let githubClientSecret: string | undefined
+let azureADClientSecret: string | undefined
+let apolloKey: string | undefined
+
+export function getRedisUrl(): string | undefined {
+  if (!redisUrl) redisUrl = getSecret('redis_url')
+  return redisUrl
+}
+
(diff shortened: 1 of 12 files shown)


packages/server/modules/shared/helpers/secretsHelper.ts
```diff
@@ -0,0 +1,75 @@
...
+  return azureADClientSecret
+}
+
+export function getApolloKey(): string {
+  if (!apolloKey) apolloKey = getSecret('apollo_key')
+  return apolloKey
+}
+
+function getSecret(secretName: string): string {
+  const secretPath = join('/etc/secrets', secretName)
```
fabis94, line 61: (How?) Will this work on local development environments? Also I'm wondering if this shouldn't be an infra only solution (e.g. storing env vars in /etc/secrets in prod deployments), not something that changes how our app reads environment config values

iainsproat: Thanks for challenging this, it led to a [better solution](https://github.com/specklesystems/speckle-server/pull/973).
To summarise what we discussed for future reference; the file won't exist on a dev environment, so this will fallback to using environment variables.  It is backwards-compatible with our current dev & test setups.
On production the files would exist, so we wouldn't need the secrets in environment variables (secrets in env vars are relatively insecure).

PR #976, opened 2022-08-31
feat: stream access requests

### LINKED ISSUE #433: Request access to a private stream
Would be nice to add a button that allows users with a stream link to request access to it if it is private. This should probably be paired with a stream setting toggle that turns on/off the option to allow users to request access

Related to #118

fabis94:
closes #433


packages/frontend/src/main/pages/stream/CommitObjectViewer.vue
```diff
@@ -1,15 +1,22 @@
 <template>
   <div class="commit-object-viewer">
-    <div v-if="(isMultiple || isCommit || isObject) && !singleResourceError">
+    <div
```
fabis94, line 3: The changes you see in seemingly unrelated Vue files are from when I was testing out Volar and fixing TS errors I found in some of the templates

gjedlicska (approved):

PR #977, opened 2022-09-01
Feature: Viewer Redux Integration

didimitrie:
## Description & motivation

A completely revamped 3D viewer. New viewer [API docs are here](https://www.notion.so/speckle/Viewer-API-Documentation-11f7bcbf3d2547c2985b0c988fb9889e).

## TODOs

- [x] fix linting errors in PropertyManager.ts
- [x] ensure colour hashes are generated consistently in a non-random fashion based on input values
- [x] re-add ghosting params to filtering manager 
- [x] re-add real time user object selection highlight 
- [x] center bubbles on selection
- [x] update viewer package readme
- [ ] update docs https://speckle.guide/dev/js.html (TODO after merge)
## Validation of changes:

Local testing with various streams. When merged into main, "fuzz" testing on latest will start. 

## Checklist:


packages/frontend/src/main/components/comments/CommentThreadViewer.vue
```diff
@@ -694,6 +694,13 @@ export default {
     window.addEventListener('beforeunload', async () => {
       await this.sendTypingUpdate(false)
     })
+
+    window.addEventListener('keydown', (e) => {
```
fabis94, line 698: Any global (e.g. window) event listeners we add should be explicitly removed, otherwise they remain even after this component is gone. And then on each creation of this component another event listener is added, eventually bogging the users device down with all of the duplicate event listeners.

Best practice is to remove all event listeners, when they're no longer necessary, even local ones on local DOM elements, but considering that the DOM is destroyed after the component is unmounted, it's possibly fine that we don't explicitly remove local event listeners.

packages/frontend/src/main/components/common/SpeckleViewer.vue
```diff
@@ -93,10 +99,23 @@ export default {
...
+
+      this.viewer.on(ViewerEvent.ObjectClicked, (selectionInfo) => {
+        handleViewerSelection(selectionInfo)
+      })
+
+      this.viewer.on(ViewerEvent.ObjectDoubleClicked, (selectionInfo) => {
+        handleViewerDoubleClick(selectionInfo)
+      })
+
+      window.addEventListener('keydown', (e) => {
```
fabis94, line 114: event listeners should be cleaned up https://github.com/specklesystems/speckle-server/pull/977/files#r961520356

packages/frontend/src/main/components/viewer/CommitInfoResource.vue
```diff
@@ -126,46 +125,36 @@ export default {
...
-          filterKey: '__parents',
-          filterValues: [id]
-        })
-      else
-        isolateObjects({
-          filterKey: '__parents',
-          filterValues: [id]
-        })
+      if (this.isolated) {
+        unIsolateObjects([id], 'ui-iso', true)
```
fabis94, line 146: ui-iso and ui-vis repeat multiple times. do they have some sort of special meaning? if so, shouldnt they be constants?

packages/frontend/src/main/components/viewer/FilterCategoryActive.vue
```diff
@@ -134,66 +110,91 @@ export default {
...
   methods: {
-    mashColorLegend(colorLegend) {
-      // just adds to our colors
-      if (!colorLegend) return
-      const keys = Object.keys(colorLegend)
-      for (const key of keys) {
-        if (!this.legend[key]) this.$set(this.legend, key, colorLegend[key])
+    async toggleFilter(prop) {
+      if (this.isolatedLegend[prop.value]) {
+        unIsolateObjects(prop.ids, 'ui-filters')
```
fabis94, line 167: is 'ui-filters' some sort of special value? might make sense being a constant

packages/frontend/src/main/components/viewer/FilterNumericActive.vue
```diff
@@ -1,20 +1,9 @@
...
-          @click.stop="colorBy = !colorBy"
-        >
-          <v-icon small :class="`${colorBy ? 'primary--text' : ''}`">
-            mdi-palette
-          </v-icon>
-        </v-btn>
-      </v-list-item-action>
-    </portal>
     <v-row no-gutters class="my-1 property-row rounded-lg">
+      <v-col cols="12">
```
fabis94, line 4: unnecessary

packages/frontend/src/main/components/viewer/FilterNumericActive.vue
```diff
@@ -75,15 +65,36 @@
           @finish="setFilterHistogram"
         />
       </v-col>
+      <!-- TODO -->
```
fabis94, line 68: unnecessary

packages/frontend/src/main/components/viewer/FilterNumericActive.vue
```diff
@@ -104,59 +115,66 @@ export default {
...
     }
   },
   mounted() {
-    this.$set(this.range, 0, this.filter.data.minValue)
-    this.$set(this.range, 1, this.filter.data.maxValue)
-    this.setFilter()
+    this.$set(this.range, 0, this.filter.min)
+    this.$set(this.range, 1, this.filter.max)
+
+    setTimeout(() => {
```
fabis94, line 156: ditto re: timeouts. is there some sort of event we could use here instead?

packages/frontend/src/main/components/viewer/LightsMenu.vue
```diff
@@ -0,0 +1,66 @@
...
+    const config = reactive({ ...DefaultLightConfiguration })
+    return {
+      config
+    }
+  },
+  watch: {
+    config: {
+      deep: true,
+      handler(newVal) {
+        getInitializedViewer().setLightConfiguration(newVal)
```
fabis94, line 61: dont use getInitializedViewer() outside of stateManager, use the `useInjectedViewer()` composable in your Vue components to get the viewer instance

packages/frontend/src/main/components/viewer/ObjectInfoResource.vue
```diff
@@ -98,38 +97,30 @@ export default {
...
-        unisolateObjects({
-          filterKey: '__parents',
-          filterValues: [id]
-        })
-      else
-        isolateObjects({
-          filterKey: '__parents',
-          filterValues: [id]
-        })
+      if (this.isolated) unIsolateObjects([id], 'ui-res', true)
```
fabis94, line 117: ui-res - another special value that could be a constant? if all of these values have special meaning for Viewer, maybe the Viewer package should export them out. if not, they could still be constants somewhere else, because they're repeated multiple times

packages/frontend/src/main/components/viewer/ObjectPropertiesRow.vue
```diff
@@ -160,16 +165,17 @@ export default {
...
+          currentFilterState
         }
       }
     `)
     const viewerState = computed(
       () => viewerStateResult.value?.commitObjectViewerState || {}
     )
 
-    return { viewerState }
+    const viewer = getInitializedViewer()
```
fabis94, line 176: useInjectedViewer

packages/frontend/src/main/components/viewer/ObjectSelection.vue
```diff
@@ -120,26 +121,30 @@ export default {
       })
     },
     isolated() {
+      // TODO
```
fabis94, line 124: there are many TODOs in this PR, some of which are probably valid and some seem like comments that were forgotten about and should be deleted, so it would be nice if youd go over the comments & todos left in this PR and get rid of the unnecessary ones

packages/frontend/src/main/components/viewer/ObjectSelection.vue
```diff
@@ -120,26 +121,30 @@ export default {
...
-          filterValues: ids
-        })
-      else
-        isolateObjects({
-          filterKey: '__parents',
-          filterValues: ids
-        })
+      if (!this.isolated) {
+        clearSelectionDisplay()
+        isolateObjects(ids, 'ui-sel')
```
fabis94, line 144: another special const value

packages/frontend/src/main/components/viewer/ViewerBubbles.vue
```diff
@@ -228,45 +241,59 @@ export default {
       this.updateBubbles(false)
     )
 
+    // TODO: this needs to be more intelligent...
     this.updateInterval = window.setInterval(this.sendUpdateAndPrune, 2000)
     window.addEventListener('beforeunload', async () => {
       await this.sendDisconnect()
     })
+
+    this.viewer.on(
```
fabis94, line 250: regarding all of the event listeners on 'Viewer' - shouldnt these be cleaned up as well? the Viewer instance is re-used, so it seems that these event listeners will keep piling up similar to the window ones

packages/frontend/src/main/components/viewer/ViewerFilters.vue
```diff
@@ -89,6 +79,11 @@
 import { useQuery } from '@vue/apollo-composable'
 import gql from 'graphql-tag'
 import { computed } from 'vue'
+import {
+  getObjectProperties,
```
fabis94, line 83: local state values should be read using queries and written to using mutator functions. judging from the name it seems like this reads some values, so this should be moved to the query instead.

if this does in fact mutate the state instead, then the name is very misleading

packages/frontend/src/main/lib/viewer/commit-object-viewer/stateManager.ts
```diff
@@ -40,28 +45,33 @@ let globalViewerData: Nullable<GlobalViewerData> = null
...
   }>,
   addingComment: false,
   preventCommentCollapse: false,
   commentReactions: ['❤️', '✏️', '🔥', '⚠️'],
-  emojis
+  emojis,
+  // New viewer & filter vars
+  currentFilterState: null as Nullable<FilteringState>,
+  selectedObjects: [] as UnknownObject[],
+  objectProperties: [] as PropertyInfo[],
```
fabis94, line 59: remember to update local-only/schema.gql with these as well, otherwise we wont get proper types in queries

packages/frontend/src/main/lib/viewer/commit-object-viewer/stateManager.ts
```diff
@@ -101,7 +111,7 @@ function getOrInitViewerData(): GlobalViewerData {
   return globalViewerData
 }
 
-function getInitializedViewer(): Viewer {
+export function getInitializedViewer(): Viewer {
```
fabis94, line 114: remove the export, this is meant to be private to the module

packages/frontend/src/main/lib/viewer/commit-object-viewer/stateManager.ts
```diff
@@ -201,301 +211,195 @@ export function setSelectedCommentMetaData(
...
-  getInitializedViewer().applyFilter(state.appliedFilter)
+export function sectionBoxOn() {
+  getInitializedViewer().sectionBoxOn()
+  updateState({ sectionBox: true })
 }
 
-export function showObjects(params: FilterKeyAndValues) {
-  const { filterKey, filterValues } = params
-  const state = { ...commitObjectViewerState() }
+export function getObjectProperties() {
```
fabis94, line 277: misleading name, the 'get' prefix implies that this function only retrieves some data and doesn't mutate anything. a more appropriate name would be 'loadObjectProperties'

packages/frontend/src/main/components/viewer/CommentsOverlay.vue
```diff
@@ -511,31 +517,40 @@ export default {
...
+      })
+      // TODO: If it's an (ortho) isometric cam.
+      // NOTE: currently not supported as parallel cam is disabled due to
+      // comment bubbles projection complications.
+      // if (camToSet[6] === 1) {
+      //   this.viewer.cameraHandler.activeCam.controls.zoom(camToSet[7], true)
+      // }
+
+      // NOTE: this is a "hack" to prevent jank - let the camera animation end
+      // before applying some heavy filters
```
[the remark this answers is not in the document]
didimitrie: we could - currently camera events are not exposed nicely though the viewer api, and camera controls will need reworking in the viewer itself - so i'm hesitant to do so right now.

packages/frontend/src/main/components/viewer/CommitInfoResource.vue
didimitrie: No special meaning really - it's a way of splitting up commands coming from different controls (model explorer, filters, selection) so the viewer filtering api can know whether to reset its internal state or not (e.g., model explorer show/hide filtering commands stack, but if a new show/hide command comes from, say, the selection info component - the prev visibility state should be cleared).

packages/frontend/src/main/components/viewer/ViewerBubbles.vue
didimitrie: Agreed in principle. For now, this would looks like a larger cleanup I don't have the energy for right now (we'd need to handle viewer events centrally somewhere, a bit like a `useViewerEvent()` from `@vueuse`).

packages/frontend/src/main/components/viewer/FilterNumericActive.vue
didimitrie: The case where this happens: we're setting a filter directly from a URL/comment with separate passMin and passMax values (e.g., `area > 4 && area < 10`). 

There's no way to update that component besides via $refs (that I found); it looks like because of the dynamic imports the `this.$refs.histogramSlider` is `undefined` in mounted. Keeping those webpack chunks works with the timeout. 

Removing the dynamic imports works with a slightly less hacky `$nextTick` call in mounted. I'll push these changes now.

AlexandruPopovici (approved):

packages/frontend/src/main/lib/viewer/commit-object-viewer/stateManager.ts
```diff
@@ -520,91 +424,139 @@ export type Filter = {
...
     }
-  } else if (filter.colorBy) {
-    toggleColorByCategory({ filterKey: filter.colorBy.property })
+    setSectionBox(box)
+    sectionBoxOn()
   }
 }
 
-export function resetFilter() {
+// NOTE: keeping legacy function around for now as it will help implement a legacy filter fallback later on.
```
fabis94, line 480: not a fan of keeping large comment chunks around just cause they might be useful later on, we have the git history for that

fabis94 (changes requested):
a couple more notes, if these are fixed the PR can be merged

packages/frontend/src/graphql/local-only/schema.gql
```diff
@@ -7,21 +7,16 @@ extend type Query {
...
-  isolateCategoryKey: String
-  isolateCategoryValues: [String!]!
-  hideCategoryKey: String
-  hideCategoryValues: [String!]!
   selectedCommentMetaData: SelectedCommentMetaData
   addingComment: Boolean!
   preventCommentCollapse: Boolean!
   commentReactions: [String!]!
   emojis: [String!]!
+  currentFilterState: JSONObject
```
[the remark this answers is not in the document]
didimitrie: Agreed! They're defined types in the viewer, but I expect some extra changes on them.

AlexandruPopovici (approved):

packages/frontend/src/main/lib/viewer/commit-object-viewer/stateManager.ts
didimitrie: ha, that was actually a reminder to do something. fixed it :)

fabis94 (approved):

AlexandruPopovici (approved):

PR #980, opened 2022-09-02
fix(server prometheus metrics): speckle_server_apollo_calls prometheus metric should be published

iainsproat:
## Description & motivation

The move from app.js to app.ts had broken `speckle_server_apollo_calls` prometheus metric and this metric was not being published.  This PR reverts the import of apolloPlugin to use the previous inline require.

## Changes:

- require apolloPlugin in server app.ts in an inline `require` call

## Checklist:


fabis94 (approved):
maybe the metrics dont work because the metric is initialized while parsing apolloPlugin.js, and maybe at that point PrometheusClient isnt initialized yet?

this is why side effects can be sketchy, code is being triggered just by triggering an import and then it can break or not break depending on when you do the import

would be cleaner to do the metric & prometheus initializations explicitly with a function, instead of it all being set up during module import

approved for now, but would be nice to clean this up in the future

PR #997, opened 2022-09-07
feat: @speckle/shared & some comments fixes

### LINKED ISSUE #982: Allow for comment creation with only a mentioned user as content
Currently, one cannot create a comment with just a single mentioned user as content. For simple quick tagging/pinging of people, we should allow this!

### LINKED ISSUE #983: Comments API: Get simple unformatted text
Allow for getting a simple plain-text (no formatting) version of the comment text. This is useful for @teocomi and other connector developers to display comments in their host applications. 

Possibly add a simple new field called `rawText` or something to the comments!

fabis94:
closes #982 
closes #983 

Example of @speckle/shared usage:
```js
import { Roles } from '@speckle/shared' // get main exports (from 'core')
import { isDocEmpty } from '@speckle/shared/dist/rich-text-editor' // get exports from sub-modules
```

We could've gotten `@speckle/shared/rich-text-editor` to work, which would look nicer, but we can't do it because the `frontend` has an old webpack version


fabis94:
This is blocked until https://github.com/specklesystems/speckle-server/pull/1000 gets merged. When it is, we can get all of the nice export maps working

fabis94:
@gjedlicska Still didn't get to use export maps for this (it's a whole can of works relating to TS support of new ESM module resolution rules, don't get me started), but at the very least I'm making all exports available through `@speckle/shared`.

This should work fine now, the only remaining issue is the CI/CD - tests fail, because @speckle/server doesn't have @speckle/shared built, and there need to be adjustments to properly publish this package. I'll leave that up to you!

gjedlicska:
We an issue, that is blocking the nice way of doing things. 
`yarn workspaces foreach -pt run build` command fails with the error below.

If the build command is invoked from inside the `packages/shared` folder, it succeeds.
Also running without the `-pt` flags, it works, its sl

image

gjedlicska (approved):

PR #1032, opened 2022-09-26
fix(server): streamUpdatePermission wont allow adding new users to streams

fabis94:


packages/server/modules/core/graph/resolvers/commits.js
```diff
@@ -216,12 +215,7 @@ module.exports = {
     },
 
     async commitReceive(parent, args, context) {
-      // if stream is private, check if the user has access to it
-      const stream = await getStream({ streamId: args.input.streamId })
-
-      if (!stream.public) {
-        await authorizeResolver(context.userId, args.input.streamId, 'stream:reviewer')
-      }
+      await authorizeResolver(context.userId, args.input.streamId, 'stream:reviewer')
```
fabis94, line 218: The stream.public check is redundant now that authorizeResolver properly checks for isPublic

packages/server/modules/core/services/streams/streamAccessService.js
```diff
@@ -13,9 +13,25 @@ const {
...
 
 /**
- * Validate that the user has the required permission level (or one above it) for the specified stream
+ * Check if user is a stream collaborator
+ * @param {string} userId
+ * @param {string} streamId
+ * @returns
+ */
+async function isStreamCollaborator(userId, streamId) {
+  const stream = await getStream({ streamId, userId })
```
gjedlicska, line 25: there is a bit of overhead here, looking up the stream is not necessary, could do with just the act table lookup

fabis94: I thought it would be better to re-use a general-purpose existing stream repository function, instead of creating a new one, considering that performance wise they would be more or less the same

gjedlicska (approved):

PR #1048, opened 2022-09-28
feat(helm): s3 configuration can be loaded from configmap

iainsproat:
## Description & motivation

Currently, when deploying a new cluster the s3 compatible storage must first be created and then the values entered into the helm chart. For deployment tooling used to create s3 compatible storage, such as Terraform, the templating of helm charts can be difficult and so is often avoided. This limitation currently requires manual intervention to copy the values of the s3 compatible storage from the deployment tooling into the helm chart. 

This PR enables variables for s3's configuration to be read by the Helm Chart from an existing configmap. This is similar to how Secrets are handled. This allows deployment tooling, such as Terraform or CloudFormation, to dynamically create an s3 bucket and then create a ConfigMap with the necessary values. This decouples the cluster deployment from the helm release, and reduces the need for manual intervention.

## Validation of changes:

This was tested in the following situations:
- deployed the Helm Chart with previous values.yaml file (i.e. `s3.configMap.enabled: false`, and `s3.endpoint` etc. set to value values). ✅
- deployed a ConfigMap with valid values, and upgraded the Helm release. The Helm Release's values were amended in the upgrade to reference the ConfigMap (i.e. `s3.configMap.enabled: true`, `s3.configMap.name` set correctly, and `s3.endpoint` etc. set to empty values) ✅
- deployed a fresh Helm release in a new namespace which referenced the ConfigMap. (i.e. `s3.configMap.enabled: true`, `s3.configMap.name` set correctly, and `s3.endpoint` etc. set to empty values) ✅

## Checklist:

### DIFF
--- a/utils/helm/speckle-server/values.schema.json
+++ b/utils/helm/speckle-server/values.schema.json
@@ -121,12 +121,12 @@
             "secretName": {
               "type": "string",
               "description": "Required. A secret containing the full connection string to the Postgres database (e.g. in format of `protocol://username:password@host:port/database`) stored within the Kubernetes cluster as an opaque Kubernetes Secret. Ref: https://kubernetes.io/docs/concepts/configuration/secret/#opaque-secrets",
-              "default": "server-vars"
+              "default": ""
             },
             "secretKey": {
               "type": "string",
               "description": "Required. The key within the Kubernetes Secret holding the connection string.",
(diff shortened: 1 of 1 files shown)


gjedlicska:
Why did you remove all the baked in default values?

gjedlicska (approved):

PR #1061, opened 2022-10-01
feat(frontend): adds missing appnames and colors

teocomi:
Adds missing appnames and colors


fabis94 (approved):
@teocomi If `apps` would be an object keyed by `name`, you wouldn't need to loop over all items and you could just do `apps[appname]`. That said - considering the size of the collection, the performance difference is negligible so it's also fine the way it is.

teocomi:
Thanks, @fabis94 , the thing is that the appnames we get don't always match exactly the hardcoded ones, that's why we need to loop through them anyways and use `appname.includes()`

PR #1073, opened 2022-10-04
added dos2unix to frontend dockerfile to avoid problems on windows wi…

flecheria:
…th nginx/docker-entrypoint.sh

### DIFF
--- a/docker-compose-deps.yml
+++ b/docker-compose-deps.yml
@@ -63,6 +63,9 @@ services:
 volumes:
   postgres-data:
   redis-data:
+  minio-data:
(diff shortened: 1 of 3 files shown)


flecheria:
Hi
yes, the problem was that docker on windows throw an error that basically said that the docker-entrypoint.sh was not inside the folder, but instead it was present.
I notice many times before the same kind of problem, and basically it happens when docker find an incorrect and of line, dos2unix solve the problem and is a best practice that I use very often on my deployments.

I have to sync the repository and try the push?

PR #1099, opened 2022-10-11
chore: securing apollo server cache against DDoS attacks

fabis94:
more info https://www.apollographql.com/docs/apollo-server/performance/cache-backends/#ensuring-a-bounded-cache

we don't use any of the server-side apollo caching functionality, so I don't anticipate any problems

### DIFF
--- a/packages/server/app.ts
+++ b/packages/server/app.ts
@@ -158,6 +158,7 @@ export async function buildApolloServer(
     introspection: true,
     cache: 'bounded',
     persistedQueries: false,
+    csrfPrevention: true,
     formatError: buildErrorFormatter(debug),
     debug,
     ...optionOverrides
--- a/packages/server/app.ts
+++ b/packages/server/app.ts
@@ -156,6 +156,8 @@ export async function buildApolloServer(
         : [])
     ],
     introspection: true,
+    cache: 'bounded',
+    persistedQueries: false,
     formatError: buildErrorFormatter(debug),
     debug,
     ...optionOverrides


didimitrie (dismissed):

fabis94:
@iainsproat Nope, this information is dispersed throughout their docs. That said, I did a quick overview of their security docs just now and I think I've got a good idea of the remaining security improvements we should do at some point:

* CORS - We do have a CORS middleware that we use on some REST API endpoints, but it doesn't appear that we have it enabled for the GraphQL API. Should we enable it, though? Is querying our API from random third party origins a valid use case (e.g. some kind of plugin/third party app)? If it is, then we can't really use a more restrictive CORS policy and we have to accept all origins.
* CSRF Protection - This is a new one I just read about in the docs - https://www.apollographql.com/docs/apollo-server/security/cors#preventing-cross-site-request-forgery-csrf. They have a CSRF prevention setting that essentially expects preflight requests to be used before all operations so that CORS settings can be validated. I just updated the PR to include this setting. That said, our CORS policy is very permissive so this won't really change much. Our auth data is stored in the frontend's localStorage, however, so while other origins can try to secretly hit the API, they won't ever have access to the users authentication data.
* Operation complexity analysis - We have some custom GraphQL rate limiting for some operations that was implemented by Cristi before he left, but at some point we should replace that with a full-on query complexity analysis library. These would allow us to assign complexity scores to our operations and then limit the maximum allowed request complexity to avoid DDoS attacks on our GraphQL API.

Issue #1094, opened 2022-10-10
Expanding some fields on the viewer sidebar freezes the tab:


AlanRynne:
From this commit: https://speckle.xyz/streams/f0532359ac/commits/98678e2a3d

When expanding the `@blocks` property on the commit object, the whole tab freezes as if it were waiting for something to load, but in my case it never came back... 🥶

This can be reproduced also with the `__closure` property in that same commit, but in that case it does come back after several seconds frozen.

@didimitrie said something on discord about this being possibly caused by our _"eager hover effect"_, which I guess @fabis94 or @AlexandruPopovici would know what it means 😅

Dim's note: this needs investigation - i suspect it's to do with the frontend telling the viewer to highlight a lot of objects at once, but unsure! 

![2022-10-10 09 30 39](https://user-images.githubusercontent.com/2316535/194820854-d6751b07-b5de-42c7-acf0-5eb8bd87c1ea.gif)

fabis94:
As discussed with Dim, this is more of a @speckle/viewer issue than a @speckle/frontend issue, so it will have to be look at and optimized by Dim or Alex. Moving back to Backlog!

AlexandruPopovici:
@fabis94  @didimitrie  Cool, I'll have a look. Did you guys find out what frontend operation is making the viewer hang?

fabis94:
@AlexandruPopovici It's FilteringManager:highlightObjects(), it triggers a lot of walking in three.js (Dim will know more)

AlexandruPopovici:
Addressed here: https://github.com/specklesystems/speckle-server/pull/1133
Filtering can still be slow in very dense streams, but speeding it up further will require a more general solution

PR #1133, opened 2022-10-21
Removing redundancy from filtering

AlexandruPopovici:
## Description & motivation
When dealing with streams holding a large number of objects, filtering gets very slow. The main reason for this is the large redundancy we currently have in working with the stream's data tree. This pull request aims at reducing some of that redundancy.
#1094 Signals such a case where the data tree is dense and filtering is very slow

## Changes:
- Removed a lot of redundant walks inside the `populateGenericState `function of the `FilteringManager`. There is still some redundancy left, but I think that fixing that will require a more general solution.
- **Speedup X6** for the particular case reported in #1094

## Checklist:


didimitrie:
I suspect I am doing similar crappy things in other parts of the filtering manager!

didimitrie (approved):

PR #1188, opened 2022-11-08
Dim fe2 styling

fabis94:
## Checklist:


packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
```diff
@@ -1,43 +1,42 @@
...
 
 :root {
-  /* background */
-  --theme-color-background: #fff; /* white */
-  --theme-color-background-2: #f3f4f6; /* gray-100 */
-  --theme-color-background-3: #e5e7eb; /* gray-200 */
+  /* blank background */
+  --bg-base: #f1f5f9;
+  /* background for elements that need to sand out from the blank backgorund, e.g. cards. */
+  --bg-default: #fff;
```
fabis94, line 17: Ok so "base" is just the base of the page and not ever supposed to be used for any panels on top of it, right? And then "default" is the default BG color for any panels or blocks on top of the base. And then "accent" is like an emphasis state.

Makes sense so far, but I think "default" might be improved considering that we might want to add different variations of colors to use on top of "base". And then it would be useful to organize all of those panel variations together, and make them stand out as a group and not mix with "base" and "accent".

So maybe "default"->"panel", and then we could have "panel-secondary", "panel-ternary" in the future? That way it's immediately clear that "base" and "accent" are their own colors and all of the "panel-*" ones relate to the blocks put on top of "base".

It's also important to evaluate whether or not we will want "accent" to differ depending on the "panel"/"default" variation chosen.

packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
```diff
@@ -1,43 +1,42 @@
...
+  /* background used to higlight elements. */
+  --bg-accent: #dbeafe;
 
-  /* foreground */
-  --theme-color-foreground: #111827; /* gray-900 */
-  --theme-color-foreground-2: #374151; /* gray-700 */
-  --theme-color-foreground-3: #6b7280; /* gray-500 */
-  --theme-color-foreground-4: #d1d5db; /* gray-300 */
+  /* default foreground color */
+  --foreground: #334155;
```
fabis94, line 22: it triggers me that foreground colors start with "foreground", but background colors start with "bg" 😄

packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
```diff
@@ -1,43 +1,42 @@
...
+  /* dimmer foreground color, e.g. caption text */
+  --foreground-dim: #94a3b8;
```
fabis94, line 24: We will likely need more than 2 foreground color variations. We don't necessarily need to add them now, but we should at least future proof the naming scheme. If the 2nd variation is "-dim" what would the third one be called? Is it intuitive to understand?

packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
```diff
@@ -1,43 +1,42 @@
...
 
-  /* link */
-  --theme-color-link: #2563eb; /* blue-600 */
-  --theme-color-link-lighter: #3b82f6; /* blue-500 */
+  /* primary color */
+  --primary: #3b82f6;
+  /* focused primary color */
+  --primary-focus: #2563eb;
```
fabis94, line 29: It's definitely going to be used in more scenarios than just ":focus" states, so we need a more abstract name that makes the meaning a bit more general. -emphasis or -accent maybe?

packages/frontend-2/assets/css/tailwind.css
```diff
@@ -51,15 +52,18 @@
   .caption {
     @apply text-xs;
   }
-}
 
-/**
- * Ensuring ring offset color always defaults to bg color
- */
-*,
-[type='checkbox']:focus,
```
fabis94, line 60: Why did you remove the ring offset background styles? 🤔 If these aren't set, then you get ugly white ring offsets in the dark theme. Are we scrapping ring offsets altogether as a design choice? If so, then I guess it makes sense to remove this

packages/frontend-2/assets/css/tailwind.css
```diff
@@ -51,15 +52,18 @@
...
-[type='radio']:focus {
-  @apply ring-offset-background;
+  .default-width {
```
fabis94, line 56: Can you elaborate on what this is? Preferably `tailwind.css` should only contain either global styles or globally re-usable classes that will be used often, and I'm not sure if it's the case for "default-width". The naming is confusing as well - the default width of what? The body? An element inside the body? A column?

packages/frontend-2/assets/css/tailwind.css
```diff
@@ -51,15 +52,18 @@
...
+    @apply mx-auto max-w-7xl px-4 sm:px-6 lg:px-8;
+  }
+
+  .background-special {
+    @apply bg-[radial-gradient(ellipse_at_top_left,_var(--tw-gradient-stops))] from-background-primary to-primary dark:bg-blue-700 dark:from-sky-400 dark:via-blue-700 dark:to-indigo-900;
```
fabis94, line 61: This feels like a class that you're going to add once in maybe app.vue or something, so I'm not sure if it has to be put inside `tailwind.css` as a globally re-usable class.

packages/frontend-2/assets/css/tailwind.css
```diff
@@ -51,15 +52,18 @@
...
+  }
+
+  .link {
+    @apply text-blue-500 dark:text-blue-600;
```
fabis94, line 65: Make sure TextLink.vue is adjusted with these styles

packages/frontend-2/assets/css/tailwind.css
```diff
@@ -77,3 +81,8 @@ body,
 div#__nuxt {
   height: 100%;
 }
+
+body {
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Helvetica, Arial,
```
fabis94, line 86: Is this correct? I'm not seeing the fonts imported at the top of the file here. Also I'm concerned about setting the main/highest priority font to be something Apple specific that might not work on other operating systems, this could make fonts differ across OSs making it more difficult for us to ensure a comparable experience for everyone

packages/frontend-2/components/form/Button.vue
```diff
@@ -104,31 +105,44 @@ const colorClasses = computed(() => {
   switch (props.type) {
     case 'outline':
       return `${
-        isDisabled ? 'text-foreground-3' : 'text-foreground'
-      } bg-transparent border-foreground-3 focus:ring-primary-lighter`
+        isDisabled
+          ? 'bg-neutral-300 dark:bg-neutral-800 text-neutral-100 dark:text-neutral-400 focus:ring-0 hover:ring-0'
```
fabis94, line 109: 1. Is it OK that all disabled states are styled the same between all button types? How will they differ when put next to each other?
2. Should the colors used here not be added as semantic colors?

packages/frontend-2/components/form/TextInput.vue
```diff
@@ -26,8 +25,8 @@
         :type="type"
         :name="name"
         :class="[
-          'block w-full shadow-sm rounded-md focus:outline-none sm:text-sm text-black',
-          'disabled:cursor-not-allowed  disabled:bg-background-2 disabled:text-foreground-3',
+          'block h-12 w-full rounded-xl focus:outline-none sm:text-sm bg-background-base text-foreground transition-all',
+          'disabled:cursor-not-allowed disabled:bg-neutral-50 disabled:dark:bg-neutral-800 disabled:text-neutral-100 disabled:dark:text-neutral-400',
```
fabis94, line 29: I see "neutral" used in many places across multiple components and with more than 3 different weights, feels like these should be semantic and the set of variations should be reduced

packages/frontend-2/components/global/AvatarGroup.vue
```diff
@@ -0,0 +1,89 @@
+<template>
+  <div class="flex items-center">
```
fabis94, line 2: We can merge these in `frontend-2` as well, but we should remember to clean up/delete our mock/test/fake components once we don't need them anymore

packages/frontend-2/components/header/NavBar.vue
```diff
@@ -1,35 +1,36 @@
...
-              :aria-current="item.current ? 'page' : undefined"
-            >
-              {{ item.name }}
-            </a>
+          <HeaderLogoBlock :active="false" class="mr-1" />
+          <div class="hidden sm:flex flex-shrink-0 items-center">
+            <HeaderNavLink to="/" name="Dashboard" :separator="false" class="ml-2" />
+
+            <HeaderNavLink
+              v-for="(nl, i) in nav.filter((n) => !!n)"
```
fabis94, line 15: these probably can't be null, and if they can, better to clean up `nav` (or create a computed?) instead of introducing filtering logic here in the template

packages/frontend-2/components/header/NavBar.vue
```diff
@@ -85,25 +86,18 @@
...
-          ]"
-          :aria-current="item.current ? 'page' : undefined"
-        >
-          {{ item.name }}
-        </DisclosureButton>
+    <DisclosurePanel class="sm:hidden" on-pointerleave="">
+      <div class="flex flex-col space-y-6 pt-2 pb-6 px-5">
+        <HeaderNavLink to="/" name="Dashboard" class="" />
+        <HeaderNavLink
+          v-for="(nl, i) in nav.filter((n) => !!n)"
```
fabis94, line 93: Duplicate template should be cleaned up at some point (we can leave this in while we're in the prototyping phase)

packages/frontend-2/components/header/NavLink.vue
```diff
@@ -0,0 +1,39 @@
...
+    type: String,
+    default: '/'
+  },
+  name: {
+    type: String,
+    default: null
+  }
+})
+const route = useRoute()
+const active = computed(() => {
```
fabis94, line 36: there's no need to manually build this logic, you can rely on `NuxtLink`s `activeClass` prop for this https://v3.nuxtjs.org/api/components/nuxt-link/#props

packages/frontend-2/components/model/Card.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <!-- eslint-disable-next-line vuejs-accessibility/mouse-events-have-key-events -->
+  <div
+    class="group bg-background hover:bg-primary rounded-xl shadow hover:shadow-xl hover:scale-[1.04] transition-all dark:shadow-slate-500/10 hover:cursor-pointer"
```
fabis94, line 4: Maybe shadow colors should also be semantic? Each of these non-semantic colors will make it more difficult for the designer to follow the entire color palette and its appropriate uses

packages/frontend-2/components/model/Card.vue
```diff
@@ -0,0 +1,125 @@
...
+    @mouseenter="hover = true"
+    @mouseleave="hover = false"
```
fabis94, line 6: I haven't seen visually what the effect you're going for is, but maybe the 'hover:' classes are enough?

packages/frontend-2/components/model/DashboardList.vue
```diff
@@ -0,0 +1,145 @@
...
+    <div class="flex items-center justify-between gap-4 mt-8 mb-4">
+      <div class="flex items-center gap-4">
+        <h1 class="h4 text-foreground font-bold flex items-center">
+          {{ title }}
+          <span class="ml-2 caption">{{ numModels }}</span>
+        </h1>
+      </div>
+      <div v-if="enableSearch">
+        <FormTextInput
+          name=""
```
fabis94, line 12: `name` is required for a reason, its accessible labels rely on it (even if there's no visible label, we should support them for screen readers)

packages/frontend-2/components/model/DashboardList.vue
```diff
@@ -0,0 +1,145 @@
...
+          label=""
+          placeholder="search"
+          class="bg-background shadow hover:shadow-lg"
+        />
+      </div>
+    </div>
+    <div class="grid grid-cols-1 sm:grid-cols-2 lg:grid-cols-4 gap-4">
+      <ModelCard
+        v-for="model in models.splice(0, numModels)"
```
fabis94, line 21: `splice()` actually mutates the source array, is that really what you're going for here? maybe you meant to use `slice()` instead?

packages/frontend-2/components/form/Button.vue
```diff
@@ -104,31 +105,44 @@ const colorClasses = computed(() => {
...
+          : 'bg-danger'
+      } text-white hover:bg-danger-darker ring-danger-lighter`
     case 'warning':
       return `${
-        isDisabled ? 'bg-warning-darker' : 'bg-warning'
-      } text-white hover:bg-warning-darker focus:ring-warning-lighter`
-    case 'secondary':
+        isDisabled
+          ? 'bg-neutral-300 dark:bg-neutral-800 text-neutral-100 dark:text-neutral-400 focus:ring-0 hover:ring-0'
```
fabis94, line 121: Ideally you shouldn't need to use `dark:` at all - semantic colors would be set up in such a way that they look correct in both themes. Maybe it makes sense to have this be an exception, but I just wanted to note this so that we're on the same page about this being something that we shouldn't need to do often.

packages/frontend-2/components/project/List.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <section>
+    <div class="flex items-center justify-between gap-4 my-8">
+      <div class="flex items-center gap-4">
+        <h1 class="h4 text-foreground font-bold flex items-center">
+          Projects
+          <span class="ml-2 caption text-foreground-dim">{{ count }}</span>
+        </h1>
+        <button
+          class="rounded-lg bg-blue-100 hover:bg-blue-200 dark:bg-slate-800 hover:dark:bg-slate-900 text-blue-500 xxxdark:text-gray-500 text-sm shadow hover:shadow-md px-4 py-2 cursor-pointer transition"
```
fabis94, line 10: 1. multiple non-semantic colors here, does it really make sense to have a completely unique button just for this? why not use one of the button types we already have, or introduce a new variation to it that could then be used elsewhere as well?
2. what's the "xxxdark" thing mean? 🤔

packages/frontend-2/composables/states.ts
```diff
@@ -0,0 +1,7 @@
+type NavItem = {
+  to: string
+  name: string
+  separator: boolean
+}
+
+export const useNav = () => useState<NavItem[]>('nav', () => [])
```
fabis94, line 7: this should be re-done as we discussed in DMs

packages/frontend-2/pages/index.vue
```diff
@@ -1,212 +1,22 @@
...
+    <!-- <header class="default-width">
+      <section>
+        <h1 class="h3 text-foreground">Henlo 👋</h1>
+        <p class="text-foreground">
+          Welcome {{ user?.name }}! This is a super big work in progress currently.
+          <NuxtLink class="link" to="/designsystem">Design System Demo</NuxtLink>
+        </p>
+      </section>
+    </header> -->
+    <main class="default-width">
```
fabis94, line 12: IMO the semantic HTML5 `<main>` element should be used somewhere at the layout level, cause otherwise devs will forget to add it in all pages, making the semantics of the page inconsistent

packages/frontend-2/pages/project.vue
```diff
@@ -0,0 +1,62 @@
...
+
+const { result, loading } = useQuery(projectQuery, () => {
+  return { id: route.params.projectId as string }
+})
+
+const project = computed(() => result.value?.project || null)
+const nav = useNav()
+
+onMounted(() => {
+  nav.value[0] = reactive({
```
fabis94, line 43: this will cause a hydration mismatch, cause onMounted is only invoked on the client side

packages/server/app.ts
```diff
@@ -131,6 +131,9 @@ function buildApolloSubscriptionServer(
  * Define mocking config in dev env
  * https://www.apollographql.com/docs/apollo-server/v3/testing/mocking
  */
+
+const roles = ['owner', 'contributor', 'reviewer']
```
fabis94, line 135: We already have all roles as constants in `mainConstants.ts` and also `@speckle/shared`

packages/server/app.ts
```diff
@@ -139,7 +142,28 @@ function buildMocksConfig(): { mocks: boolean | IMocks; mockEntireSchema: boolea
     mocks: {
       Query: () => ({
         testNumber: () => faker.datatype.number(),
-        testList: () => [...new Array(faker.datatype.number({ min: 1, max: 10 }))]
+        testList: () => [...new Array(faker.datatype.number({ min: 1, max: 10 }))],
```
fabis94, line 145: I'm fine with merging the mocks into `frontend-2` :+1:

packages/frontend-2/assets/css/tailwind.css
didimitrie: adding it as a separate class, and pref through tailwind config!

packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
didimitrie:
```
bg-background -> bg-panel
bg-backgorund-accent -> bg-panel-2
(+ others in the future)

bg-background-page -> bg-page


foreground -> foreground
foreground-dim -> foreground-2
(+ others in the future)

[danger, warning] -> [danger, warning]
[danger, warning]-darker
[danger, warning]-lighter

// Note: we want primary colors to become lighter when focused/hovered in a dark theme, and darker when focused/hovered in a light theme, hence the differentnaming scheme
primary -> primary
primary-focus -> primary
primary-muted -> primary-muted

```

packages/frontend-2/assets/css/tailwind.css
didimitrie: unused, comment it out. if we reintroduce it, name it better (e.g. background-panel-gradient)

packages/frontend-2/components/form/Button.vue
didimitrie: TODO: add **disabled** semantic colors

packages/frontend-2/assets/css/tailwind.css
didimitrie: second conclusion: we need it, and let's name it `layout-single-column`

packages/frontend-2/components/header/NavBar.vue
didimitrie: YOLO, we use your system (once you code it up!)

packages/frontend-2/components/model/Card.vue
didimitrie: cleanup needed, these were experiments

packages/frontend-2/components/model/DashboardList.vue
didimitrie: legit, i'll re-add it

packages/frontend-2/components/project/List.vue
didimitrie: needs to be migrated to an actual button

packages/frontend-2/assets/css/tailwind.css
didimitrie: removed, it was a superfluous class

packages/frontend-2/assets/css/tailwind-setup/semantic-colors.css
didimitrie: Realised a thing: having a `background` color meant that we'd type `bg-background`, `bg-background-2` as actual css class names (`<div class="bg-background-2">`), which triggered me (too many bg-background-backgrounds). Above i was confusing the tailwind class and the css color name 🙃 

`bg-page`/`bg-panel` as a tailwind class would mean `page`/`panel` as a tailwind.config color, which doesn't really work (or felt weird). 

I propose we try to use `base`, `base-2`, and `base-page` as css vars, and added the tailwind.config color `base` with these variations. This means that you can now do `bg-base`, `bg-base-2` and `bg-base-page` when setting element backgrounds.

packages/frontend-2/assets/css/tailwind.css
didimitrie: tried a thing, tbd if the best way 😅

packages/frontend-2/components/form/TextInput.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <label v-show="showLabel" :for="name" class="block label">
+    <label :for="name" class="block label" :class="{ 'sr-only': !showLabel }">
```
didimitrie, line 3: nice!

PR #1213, opened 2022-11-16
Rate limit all endpoints

iainsproat:
## Description & motivation

We should have Express middleware to comprehensively rate limit all endpoints.

## Changes:

- adds a rate limiting module to the server package
- rate limiting applied as a middleware to Express
- refactors existing rate limiting to use the new rate limiting module

## To-do before merge:

- [ ] take a look at the middleware function, and try to determine why it's always using `ip` address.
- [ ] should add more rate limit tests. There's only one test, for user create rate limit.

## Screenshots:

Screenshot 2022-11-17 at 13 49 03

Screenshot 2022-11-16 at 16 22 50

## Checklist:

## References

https://github.com/animir/node-rate-limiter-flexible/wiki/Express-Middleware


packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,184 @@
...
+    res.setHeader('Retry-After', rateLimiterRes.msBeforeNext / 1000)
+    res.setHeader('X-RateLimit-Remaining', rateLimiterRes.remainingPoints)
+    res.setHeader(
+      'X-RateLimit-Reset',
+      new Date(Date.now() + rateLimiterRes.msBeforeNext).toISOString()
+    )
+  }
+  if (action) {
+    const opts = LIMITS[action]
+    res.setHeader('X-RateLimit-Limit', opts.limitCount)
```
gjedlicska, line 124: lets send a retry after header too with the cooldown period
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,170 @@
...
+  res: express.Response,
+  next: express.NextFunction
+) => {
+  if (isTestEnv()) next()
+
+  const burstAction = 'ALL_REQUESTS_BURST'
+  const action = 'ALL_REQUESTS'
+
+  const source: string = (req as RequestWithContext)?.context?.userId
+    ? ((req as RequestWithContext)?.context?.userId as string)
```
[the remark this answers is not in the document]
gjedlicska: its going to be a problem, because the ratelimiter middleware is now called before the context middleware and the context doesn't exist on the req object when ratelimiting is evaluated. I'll look into this.

packages/server/modules/core/services/ratelimiter.ts
gjedlicska: lol im blind...

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,184 @@
...
+export async function isWithinRateLimits({
+  action,
+  source
+}: isWithinRateLimitsConfig): Promise<RateLimiterRes> {
+  const rlOpts = LIMITS[action]
+  if (!rlOpts) return Promise.reject(null) // the rate limits for the action have not been defined, so prevent use of the action
+
+  // provides a backup store for the rate limiter
+  // incase Redis is unreachable
+  const rateLimiterMemory = new RateLimiterMemory({
```
gjedlicska, line 174: you are creating a new instance of the memory rate limiter for every call of this function. 

That doesn't really do anything, that instance needs to be created once at startup.

Possibly its also inefficient to create a new instance of the RateLimiterRedis for earch request

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,344 @@
...
+  RateLimiterRedis,
+  RateLimiterRes
+} from 'rate-limiter-flexible'
+import { TIME } from '@speckle/shared'
+import { AuthContext } from '@/modules/shared/authz'
+import { BaseError } from '@/modules/shared/errors'
+import { getIpFromRequest } from '@/modules/shared/utils/ip'
+import { Options } from 'verror'
+
+export class RateLimitError extends BaseError {
```
fabis94, line 21: move the error to `/core/errors`

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,344 @@
...
+  static code = 'RATE_LIMIT_ERROR'
+
+  rateLimitBreached: RateLimitBreached
+
+  constructor(
+    rateLimitBreached: RateLimitBreached,
+    message?: string | null | undefined,
+    options: Options | Error | undefined = undefined
+  ) {
+    super(message ?? RateLimitError.defaultMessage, options)
```
fabis94, line 33: you rarely need to use `??` in JS, in this case it would be better to use `||` instead. `??` is only going to check for null while `||` is going to check for all falsy values

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,344 @@
...
+type RateLimiterOptions = {
+  [key in RateLimitAction]: BurstyRateLimiterOptions
+}
+
+interface RateLimitContext extends AuthContext {
+  ip: string
+}
+
+interface RequestWithContext extends express.Request {
+  context: RateLimitContext
```
fabis94, line 72: Preferably we should use type augmentations (see `type-augmentations` folder) to augment the global express.Request type so that we can get the context object anywhere.

We can mark the property as optional (`?`), but then we'll always have to check if its not undefined, so I think we should skip that, even if it makes writing code before the ctx object is initialized (which is a very small part of the project) a bit less type-safe

fabis94 (changes requested):

packages/server/type-augmentations/express.d.ts
```diff
@@ -0,0 +1,10 @@
+import { Request } from 'express'
+import { AuthContext } from '@/modules/shared/authz'
+
+declare module 'express' {
+  interface RequestWithAuthContext extends Request {
```
fabis94, line 5: Just extend `Request` directly, no need to introduce a new interface. This will ensure that whenever you use the main express Request type, it has the property on it.

```ts
declare module 'express' {
  interface Request {
    context: AuthContext
  }
}
```

packages/server/modules/core/errors/ratelimit.ts
```diff
@@ -0,0 +1,20 @@
...
+export class RateLimitError extends BaseError {
+  static defaultMessage =
+    'You have sent too many requests. You are being rate limited. Please try again later.'
+  static code = 'RATE_LIMIT_ERROR'
+
+  rateLimitBreached: RateLimitBreached
+
+  constructor(
+    rateLimitBreached: RateLimitBreached,
+    message?: string | null | undefined,
```
fabis94, line 14: I don't really see the point of allowing `null` here, if you don't want to add a message then skip the parameter, which is going to resolve to `undefined`. Additionally, I don't think you can put an optional parameter before more required parameters (message? before options).

```ts
constructor(
rateLimitBreached: RateLimitBreached,
message?: string,
options?: Options | Error
) {}
```

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,317 @@
...
+    'X-RateLimit-Reset',
+    new Date(Date.now() + rateLimitBreached.msBeforeNext).toISOString()
+  )
+  res.setHeader('X-Speckle-Meditation', 'https://http.cat/429')
+  return res.status(429).send({
+    err: 'You are sending too many requests. You have been rate limited. Please try again later.'
+  })
+}
+
+export const getActionForPath = (path: string, verb: string): RateLimitAction => {
```
fabis94, line 190:
```ts
export const getActionForPath = (path: string, verb: string): RateLimitAction => {
    const maybeAction = `${verb} ${path}` as keyof typeof RateLimitAction
    return RateLimitAction[maybeAction] || RateLimitAction.ALL_REQUESTS
}
```

packages/server/modules/core/services/ratelimiter.ts
```diff
@@ -0,0 +1,317 @@
...
+        return next()
+      } catch (err) {
+        if (!(err instanceof RateLimitError)) throw err
+        return sendRateLimitResponse(res, err.rateLimitBreached)
+      }
+    }
+  }
+}
+
+// we need to take the Bursty specific type because its not an Abstract.
```
fabis94, line 238: are these comments still applicable? if not, please delete them, if yes convert to a `todo` with some kind of less opaque message that details what needs to be done here

gjedlicska (approved):

PR #1223, opened 2022-11-23
Improves IFC parsing speed ~30x & captures object psets

didimitrie:
See log & changes here: https://www.notion.so/speckle/IFC-Season-3-877a777373fc441cb400cfbe2c349849

### DIFF
--- a/packages/fileimport-service/ifc/import_file.js
+++ b/packages/fileimport-service/ifc/import_file.js
@@ -7,7 +7,7 @@ const { parseAndCreateCommit } = require('./index')
 async function main() {
   const cmdArgs = process.argv.slice(2)
 
-  const [filePath, userId, streamId, branchName, commitMessage] = cmdArgs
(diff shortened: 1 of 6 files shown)


packages/fileimport-service/ifc/import_file.js
```diff
@@ -7,7 +7,7 @@ const { parseAndCreateCommit } = require('./index')
 async function main() {
   const cmdArgs = process.argv.slice(2)
 
-  const [filePath, userId, streamId, branchName, commitMessage] = cmdArgs
+  const [filePath, userId, streamId, branchName, commitMessage, fileId] = cmdArgs
```
didimitrie, line 10: @gjedlicska, this is something i'd want to flag: i'm introducing here the file id as an extra argument for the parsing process. currently this only used (and passed to) the ifc converter. 

The intention is that we'd be able to store a link to the blob within the object it was generated from. Ideally we'd move this down later to the other parsers, but my py skills are `zero` and i'm afraid to muck around in there.

PR #1232, opened 2022-11-30
Frontend 2 docker build

gjedlicska:
Add Docker build and CI to FE2


packages/frontend-2/nuxt.config.ts
```diff
@@ -19,12 +15,11 @@ export default defineNuxtConfig({
       }
     ]
   ],
-
   runtimeConfig: {
     public: {
-      API_ORIGIN,
-      MIXPANEL_API_HOST,
-      MIXPANEL_TOKEN_ID
+      apiOrigin: 'foo',
```
fabis94, line 20: i'm guessing the values will be changed by .env values during runtime, right? if so we should still set some kind of sensible defaults or at least something like 'UNDEFINED' so that if we ever forget to set the env vars we get more reasonable default values

PR #1249, opened 2022-12-06
refactor(frontend-2 liveness/readiness): adds checks for server health

iainsproat:
## Description & motivation

Liveness checks that nuxt server is responding, readiness checks it can connect to the backend and the database.

Addresses https://github.com/specklesystems/speckle-server/issues/1250

## Checklist:


packages/frontend-2/server/api/v2/healthz/readiness.get.ts
```diff
@@ -0,0 +1,80 @@
...
+  ApolloQueryResult,
+  gql,
+  HttpLink,
+  InMemoryCache,
+  NormalizedCacheObject
+} from '@apollo/client/core'
+
+type GraphQLClient = ApolloClient<NormalizedCacheObject>
+
+const serverInfoVersionQuery = gql`
```
fabis94, line 12: Use `graphql()` out of `~~/lib/common/generated/gql` instead of `gql()`, this will ensure TypeScript types are auto-generated (when you run the dev server through `yarn dev`, or separately through `yarn gqlgen`) and you'll get properly typed result structures. There's various examples of such queries/mutations across the `frontend-2` package also.

packages/frontend-2/server/api/v2/healthz/liveness.get.ts
```diff
@@ -0,0 +1,5 @@
+export default defineEventHandler(() => {
```
fabis94, line 1: I'm not sure of the `/api/v2` prefix for fe2 API routes. It kind of implies that these are a v2 REST API of the `server` routes. Better to use a completely different prefix altogether like `/api-frontend` or something.

packages/frontend-2/server/api/v2/healthz/readiness.get.ts
```diff
@@ -0,0 +1,80 @@
...
+  query serverInfo {
+    version
+  }
+`
+
+export default defineEventHandler(() => {
+  // as part of creating the apollo client it checks that the server can be reached
+  return createApolloClient('apiOrigin') //FIXME there's some magic going on with the config that clearly isn't applied here (yet?)
+    .then(() => {
```
fabis94, line 21: use async/await syntax

packages/frontend-2/server/api/v2/healthz/readiness.get.ts
```diff
@@ -0,0 +1,80 @@
...
+        message:
+          'Frontend is unable to reach the Graphql server, or the server is unable to query the database',
+        statusMessage: 'Internal Server Error'
+      })
+    })
+})
+
+// TODO copied from packages/server/modules/cli/commands/download/commit.ts
+// most likely a better place to put the below
+const createApolloClient = async (origin: string): Promise<GraphQLClient> => {
```
fabis94, line 39: No need to copy this from anywhere, because the new frontend already has its own Apollo Client setup.

Try any of the following options to get the client:
1. `useApolloClient().client`
2. `const {$apollo} = useNuxtApp()`

Now these will likely not work, because these API routes run in a very different context from the other app routes, but I want to test this out anyway.

If it doesn't work, then you'll have to create a new ApolloClient instance, but even then you can re-use the existing config, instead of setting everything up from scratch. `~~/lib/core/configs/apollo.ts` exports a function for building this config, so use that.

I do see that this function relies on some Vue/Nuxt composables like `useRuntimeConfig()` and `useAuthCookie()` which might not be available in API routes, but even in that case the function could just be split apart into a new one that allows you to feed in apiOrigin & authToken some other way. E.g.:

```ts

export const buildApolloConfig(params: {apiOrigin: string, authToken: Ref<Optional<string>>}) {
 ...
}

const defaultConfigResolver: APolloConfigResolver = async () => {
  const {
    public: { apiOrigin }
  } = useRuntimeConfig()

  const authToken = useAuthCookie()
  return buildApolloConfig({apiOrigin, authToken})
}

export default defaultConfigResolver;
```

And inside your API route:
```ts
import {ref} from 'vue'

const apiOrign = ... // still taken from runtime config somehow?
const authToken = ref(undefined); // i'm guessing we don't need a token here so we can leave it undefined
const config = await buildApolloConfig({apiOrigin, authToken});
```

packages/frontend-2/server/api/v2/healthz/readiness.get.ts
```diff
@@ -0,0 +1,80 @@
...
+  // as part of creating the apollo client it checks that the server can be reached
+  return createApolloClient('apiOrigin') //FIXME there's some magic going on with the config that clearly isn't applied here (yet?)
+    .then(() => {
+      return {
+        status: 'ready'
+      }
+    })
+    .catch(() => {
+      throw createError({
+        statusCode: 500,
```
fabis94, line 28: Should we not at the very least log the actual error that was thrown?

fabis94 (changes requested):

Issue #1361, opened 2023-02-07
Review server logging and improve sensitivity and quality of logged messages


iainsproat:
Follows on from: https://github.com/specklesystems/speckle-server/issues/1105

Now that we have structured logs and these are being aggregated, we have the opportunity to improve our logging.

We should log more relevant events.
We should reduce unhelpful log events.

PR #1416, opened 2023-02-24
feat(logging): log all http requests and responses

iainsproat:
## Description & motivation

Uses pino's http logger to automatically log requests and responses to all REST API endpoints.
Headers are removed so sensitive information is not logged, with the exception of select 'allow-listed' headers.
Log level is adjusted depending on http status code (error for 500<=x, silent for 300>=x<400, info otherwise).

Part of https://github.com/specklesystems/speckle-server/issues/1361

## Changes:

- enables pino-http auto logger, removing all but an allow list of headers from the logs.
- adds a `x-request-id` header and `req.id` property to all requests and responses.
- delegates most response logging to pino-http auto logger.

## Checklist:

### DIFF
--- a/packages/server/modules/core/tests/generic.spec.js
+++ b/packages/server/modules/core/tests/generic.spec.js
@@ -1,4 +1,12 @@
 /* istanbul ignore file */
+const { mockRequireModule } = require('@/test/mockHelper')
+const envHelperMock = mockRequireModule(
+  [
+    '@/modules/shared/helpers/envHelper',
+    require.resolve('../../shared/helpers/envHelper')
+  ],
+  ['@/modules/shared/index']
+)
 const expect = require('chai').expect
 
 const { beforeEachContext } = require('@/test/hooks')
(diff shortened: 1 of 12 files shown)


packages/server/logging/expressLogging.ts
```diff
@@ -1,7 +1,75 @@
...
 import HttpLogger from 'pino-http'
+import { IncomingMessage } from 'http'
+import { NextFunction, Response } from 'express'
+import pino, { SerializedResponse } from 'pino'
 
 export const LoggingExpressMiddleware = HttpLogger({
   logger,
-  autoLogging: false
+  autoLogging: true,
+  genReqId: (req) => {
```
gjedlicska, line 11: if you extracted the function out from the `DetermineRequestIdMiddleware`, to take the existing id first and fall back to the header, than the guid, you could pass that function in here to the `genReqId` argument and reuse it in the middleware

gjedlicska (changes requested):

Issue #1200, opened 2022-11-11
Verification Emails Do Not Display Button in Outlook Desktop Dark Mode


didimitrie:
## What package are you referring to?
Speckle Server / Emails module

## Describe the bug

Our big friendly blue button from emails does not display properly in MS Outlook Desktop in Dark Mode. 

Reported on twitter: https://twitter.com/gschleusner/status/1591151312699330575

## Expected behavior

The button should display. 

## Screenshots

See twitter thread linked above.

ahofbeck:
It is not only happening with the Darkmode
![image](https://user-images.githubusercontent.com/32340026/214433394-75431bf9-34d2-4203-a1cf-a998255883e5.png)

but also old school with the white background
![image](https://user-images.githubusercontent.com/32340026/214433518-a8c51bb2-c4b5-4131-8d7a-438a530f4e0a.png)

![image](https://user-images.githubusercontent.com/32340026/214433898-4b21370a-befb-4a7a-9afd-55fea11fc897.png)

gjedlicska:
Well, it looks like Outlook on especially windows [doesn't support](https://www.campaignmonitor.com/css/color-background/css-gradients/) CSS gradients.

This causes the header logo component and the CTA button to be not rendered at all in the desktop client.

gjedlicska:
Hey @ahofbeck,

This has been released to `speckle.xyz` would u mind testing a server / stream invite email in your desktop outlook mail?

Thanks :)

ahofbeck:
Hi @gjedlicska,

it is still the same issue :(. We did it on our own deployment with newest version, but it was the same for speckle.xyz

![MicrosoftTeams-image (21)](https://user-images.githubusercontent.com/32340026/224033407-4ca9881b-eced-44a9-ad9d-71302cd02230.png)
![MicrosoftTeams-image (20)](https://user-images.githubusercontent.com/32340026/224033412-944a9d0a-f947-4b9b-bce0-96238e3063b6.png)

gjedlicska:
Thanks for testing @ahofbeck, reopening for further investigation.

gjedlicska:
Ok, I think I fixed this:

Rounding the corners is still too much to ask, but hey, atleast you can click the button...

![image](https://user-images.githubusercontent.com/57442769/224943717-6cb33acc-bf5d-4b0c-90a7-e13afc7527ef.png)


[...]

PR #1448, opened 2023-03-13
refactor(server email templates): use the mjml ejs rendering for all emails

gjedlicska:
This has been left half baked for a while. Now all outgoing emails should use the ejs mjml on the fly rendering.

Unforch my local server is refusing to send emails with mailjet, so I'd like to get this out to latest for a proper testing.
Once that is done, I'm gonna create an additional cleanup PR, that removes the traces of the old rendering system.


gjedlicska:
hopefully fixes #1200

Issue #1471, opened 2023-03-19
Speckle viewer `buildBatch` TypeError on loading model `batch[0] is undefined`


m-clare:
## What package are you referring to?
`@speckle/viewer: "^2.12.5"`

## Describe the bug

Loading models with the viewer results in the following console error:

```
Uncaught (in promise) TypeError: batch[0] is undefined
    buildBatch Batcher.ts:97
    makeBatchesAsync Batcher.ts:78
    addRenderTreeAsync SpeckleRenderer.ts:483
    loadObjectAsync Viewer.ts:464
    loadModel SpeckleSceneWithHooks.jsx:7
    SpeckleSceneWithHooks SpeckleSceneWithHooks.jsx:38
    React 8
    workLoop scheduler.development.js:266
    flushWork scheduler.development.js:239
    performWorkUntilDeadline scheduler.development.js:533
Batcher.ts:97:39
```
based on the logic here:
https://github.com/specklesystems/speckle-server/blob/4ebb776feb4d0b78817eaad182aed86240b7180a/packages/viewer/src/modules/batching/Batcher.ts#L94

Upon inspecting the `batch` object, it appears that it eventually is an empty array:
https://github.com/specklesystems/speckle-server/blob/4ebb776feb4d0b78817eaad182aed86240b7180a/packages/viewer/src/modules/batching/Batcher.ts#L92

## To Reproduce

Using a minimal Create-React-App with a Speckle Viewer as a component (https://github.com/m-clare/react-speckle-viewer)

Add in proper environment variables to a public speckle stream (`.env`)

Run:
```
$ yarn 
$ yarn dev
```

## Expected behavior

No errors in the DevTools console.

## System Info

If applicable, please fill in the below details - they help a lot!

### Desktop (please complete the following information):

- OS: iOs
- Browser: Firefox
- Version: 110.0.1

## Proposed Solution (if any)

PR #1472  (fixed some logic within `viewer/src/modules/batching/Batcher.ts`)

PR #1472, opened 2023-03-19
Fix: add logic to handle empty batch array after filter

m-clare:
Suggested fix for #1471. Adds conditional logic to deal with empty arrays being passed to Batcher, causing errors here:
https://github.com/specklesystems/speckle-server/blob/4ebb776feb4d0b78817eaad182aed86240b7180a/packages/viewer/src/modules/batching/Batcher.ts#L94

## Description & motivation

Small change to the batch logic to handle if the filter on `batch` in `buildBatch` returns an empty array (feel free to do otherwise, it's just been annoying for working with the viewer elsewhere).

### DIFF
--- a/packages/viewer/src/modules/batching/Batcher.ts
+++ b/packages/viewer/src/modules/batching/Batcher.ts
@@ -75,9 +75,11 @@ export default class Batcher {
 
     for (let i = 0; i < materialHashes.length; i++) {
       const batch = this.buildBatch(subtreeId, rendeViews, materialHashes[i], batchType)
-      this.batches[batch.id] = batch
-      yield this.batches[batch.id]
-      await pause()
+      if (batch) {
+        this.batches[batch.id] = batch
+        yield this.batches[batch.id]
+        await pause()
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


AlexandruPopovici:
Hello @m-clare 

Good catch! It's true that the current implementation doesn't handle batches with no render views to build from, however I'm curious how did you stumble upon this situations. Off the top of my head, the only way there would be no render views when building a batch would be to either supply an empty array of renderviews to the `buildBatch` , either supply some, but they all have invalid geometry i.e. `validGeometry` fails for all of them.

Do you have a stream that you could share which falls into any of these scenarios? 

We definitely need a guard for when this happens, however the solution would be to not create the batch at all, instead of creating an empty batch.

m-clare:
Hi @AlexandruPopovici, 

I was running into this issue when loading a public stream to try out a very bare bones implementation I have for loading speckle objects into a react component.

The relevant branch is here:
https://github.com/m-clare/react-speckle-viewer/tree/bug/batch-empty-array

And this is what happens when I start up the app.
image

Interestingly, sometimes I only get parts of the model, not the whole thing. When I first wrote this viewer, I think it was before batching was implemented, so that's also another possible wrinkle. I'm working on updating the component with the new Viewer functionality.

AlexandruPopovici:
Hi @m-clare 

I've run the branch. Thank you for sharing! It turns out the root cause for the issue you reported is something else. It's caused by having multiple `Viewer` instances running in parallel, like you have on your branch. You can get rid of the error by setting the `keepGeometryData` flag from the `DefaultViewerParams` to `true`. However, even with the error gone, you'll get weird behavior in the two viewer instances caused by race conditions.

We haven't addressed the topic of multiple parallel viewer instances running yet, but it's on our TO DO list in the near future. I'll keep you posted!

AlexandruPopovici:
Hi @m-clare 

This [PR](https://github.com/specklesystems/speckle-server/pull/1518) has been merged and it's vailable on `latest`. It addresses the root cause of the issue you've reported. Additionally, we've added a guard [here](https://github.com/specklesystems/speckle-server/blob/7ea5e35761d9202250f80f9b16fa7d3abf843b81/packages/viewer/src/modules/batching/Batcher.ts#L211) as per your PR, since it is technically possible to have a batch composed of only invalid geometry caused by connector or otherwise issues

Issue #1447, opened 2023-03-10
Efficient per-object rendering and interaction


AlexandruPopovici:
Details and log [here](https://www.notion.so/speckle/Efficient-per-object-rendering-and-interaction-06e831c7591346cfbea6c380c79bc7f4)

- [x] https://github.com/specklesystems/speckle-server/issues/1051
- [x] Add index attribute and per object uniform structs array to the shaders
- [x] Implement opaque per object uniform management in the batch
- [x] Build BVHs at object level.
- [x] User individual object's inverse transformation when querying the BVH
- [ ] Implement top level BVH and bottom level BVH ❓ 
- [x] Review and make required changes to the material system for better handling of current concept changes
- [x] Restructure and add missing abstractions regarding Meshes, Objects and Batches
- [x] Add minimal object transform updates

Issue #1051, opened 2022-09-29
Limit maximum batch size


AlexandruPopovici:
We need to limit the maximum batch size. On chromium for example, for extra-large batches I think we're seeing scene-splits which is bad.

gjedlicska:
Is this about batches inside the viewer, or the batch size, how stream data is received from the backend?

AlexandruPopovici:
It's about the geometry batches inside the viewer. Currently there is not size limit, so for streams with dense geometries and little to no material variation we end up with batches so large that the driver is forced to do scene splits. This is bad for performance

PR #1488, opened 2023-03-28
Alex/#1447 advanced batching

AlexandruPopovici:
## Description & motivation
Addresses #1447 
This PR brings new functionality to the viewer regarding how our entire bataching system works. We are now able to individually transform objects, while still rendering geometry in batches, with little to no performance overhead. Additionally, the viewer custom material system was overhauled partly required by the new batching, partly because the new viewer context required some changes.
With these changes, the viewer starts to get closer to a general purpose viewer, and extends the range of user applications

## Changes:
- Implemented batch splitting #1051 . Batches are now split based on a configurable max vert count as well as max object count
- Implemented individual object transformation using:
  - Vertex textures using `texelFetch` for WebGl 2.0 contexts that support fp32 textures
  - Vertex textures with regular texture sampling for WebGL 1.0 contexts that support fp32 textures
  - Uniform arrays for all context where fp32 textures are not supported
 - Implemented per-object BVH with minimal acceleration for the groups of BVHs inside a batch
 - The new BVH approach avoids the need to be rebuilt whenever individual objects are transformed
 - Implemented instancing materials for minimal context switching and minimal instancing frequency
 - Implemented our own method of overriding materials on the entire scene (since three's default way was messing things up)
 - Overhauled the entire custom material system. Our custom materials will be exposed to the users soon enough, so we needed to make sure they hide most of not all of the complexity required for RTE and TRANSFORM_STORAGE
 - Implemented a form of mixins to our custom material system so we avoid repeated boilerplate code 
 - Added `explode` function the API. Current explosion is just a simple spherical one, centered around the stream's center
 - There is also the `getObjects` function in the API which recieves an object id and returns the `BatchObject` instances which can be used to transform them individually. But I'm inclined to remove it for now until we figure out how to expose this entire functionality to the outside world

## To-do before merge:
- Re-test the uniform array based TRANSFORM_STORAGE on multiple devices
- Test some more using the frontend. 

## Checklist:


didimitrie (dismissed):
Honestly, only Alex can understand this right now, but LGTM post some poor-man's testing in the viewer sandbox!

PR #1489, opened 2023-03-29
feat: proper sign up tracking

fabis94:
more info https://github.com/specklesystems/admin/issues/381


gjedlicska:
Looks good, we only need the network policy to warp this up right?

PR #1508, opened 2023-04-07
feat(server): authentication middleware should log auth context creation status

iainsproat:
## Description & motivation

Part addresses https://github.com/specklesystems/admin/issues/393

- this uses the pino http logger provided via prior express middleware, ensuring a request ID is associated with the log messages
- userID, scopes and roles will be logged

This is a slight workaround as we cannot add the auth context to the request or response log messages generated by pino http logger. As the auth context does not yet exist when pino http logger middleware is called prior to the auth middleware. Additionally, the auth context is not returned in the response, so pino http cannot log it in the response.

## Checklist:

### DIFF
--- a/packages/server/modules/shared/middleware/index.ts
+++ b/packages/server/modules/shared/middleware/index.ts
@@ -87,14 +87,14 @@ export async function authContextMiddleware(
 ) {
   const token = getTokenFromRequest(req)
   const authContext = await createAuthContextFromToken(token)
-  const logger = req.log.child({ authContext })
+  req.log = req.log.child({ authContext })
   if (!authContext.auth && authContext.err) {
     let message = 'Unknown Auth context error'
     let status = 500
     message = authContext.err?.message || message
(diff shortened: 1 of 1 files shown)


packages/server/modules/shared/middleware/index.ts
```diff
@@ -87,15 +87,18 @@ export async function authContextMiddleware(
...
     let message = 'Unknown Auth context error'
     let status = 500
     message = authContext.err?.message || message
     if (authContext.err instanceof UnauthorizedError) status = 401
     if (authContext.err instanceof ForbiddenError) status = 403
+    logger.warn('Auth context creation failed.')
     return res.status(status).json({ error: message })
   }
   req.context = authContext
+  logger.info('Auth context created')
```
gjedlicska, line 101: I dont think we need to log the success here. The warning is good enough

gjedlicska (changes requested):

gjedlicska (approved):

PR #1589, opened 2023-05-23
fix: switching over to new WS implementations (conditionally)

fabis94:
Shouldn't be merged until we can validate that this won't break connectors!

**New implementation must be enabled through env vars:**
- **server**: `NEW_WS_IMPLEMENTATION=true`
- **frontend**: `VITE_NEW_WS_IMPLEMENTATION=true`
- **frontend-2**: `NUXT_PUBLIC_NEW_WS_IMPLEMENTATION=true`


fabis94:
@iainsproat no its not, please dont close the PRs ive intentionally left open

PR #1608, opened 2023-06-04
Dim/fe2/view changes

didimitrie:
View changes in FE2. Some things still todo, namely checking the interactions with open threads and diff mode.


packages/frontend-2/assets/css/tailwind.css
```diff
@@ -8,6 +8,23 @@
  * Don't pollute this - it's going to be bundled in all pages!
  */
 
+@layer components {
```
didimitrie, line 11: should probably go in the extracted package, but i don't have enough chops to do it

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -328,7 +333,7 @@ function useViewerCameraIntegration() {
       // Only now set projection, we can't do it too early
       orthoProjectionUpdate(isOrthoProjection.value)
     } else {
-      loadCameraDataFromViewer()
+      if (!diff.diffString.value) loadCameraDataFromViewer() // NOTE: a bit of a hack
```
didimitrie, line 336: for some reason, threads with diffs, when opened up, fail to set the correct view IF there's another model loaded in the scene besides the ones being diffed..

packages/frontend-2/components/viewer/compare-changes/Version.vue
```diff
@@ -0,0 +1,33 @@
...
+      {{ isNewest ? 'New' : 'Old' }} Version
+    </div>
+  </div>
+</template>
+<script setup lang="ts">
+import dayjs from 'dayjs'
+import localizedFormat from 'dayjs/plugin/localizedFormat'
+import { ViewerModelVersionCardItemFragment } from '~~/lib/common/generated/gql/graphql'
+
+dayjs.extend(localizedFormat)
```
fabis94, line 21: this should be done in a nuxt plugin (we already have one for dayjs)

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
fabis94: this should be a comment in the code imo

packages/frontend-2/lib/viewer/composables/diffs.ts
```diff
@@ -0,0 +1,110 @@
...
+  const diff = async (modelId: string, versionA: string, versionB: string) => {
+    const { data, errors } = await apollo.query({
+      query: viewerDiffVersionsQuery,
+      variables: {
+        projectId: state.projectId.value,
+        modelId: modelId,
+        versionAId: versionA,
+        versionBId: versionB
+      },
+      fetchPolicy: 'network-only'
```
fabis94, line 29: why 'network-only'?

PR #1634, opened 2023-06-15
feat(server): opt-in audience subscribing on registration

didimitrie:
Adds a newsletter registration consent box on the register screen.


utils/helm/speckle-server/templates/server/deployment.yml
```diff
@@ -276,6 +276,25 @@ spec:
           - name: EMAIL_FROM
             value: "{{ .Values.server.email.from }}"
           {{- end }}
+          
+          # *** Mailchimp ***
```
didimitrie, line 280: cc @iainsproat: this is where your magic will be needed. PS: we have these values in the shared vault.

packages/frontend-2/components/auth/RegisterWithEmailBlock.vue
```diff
@@ -43,6 +43,23 @@
...
     />
     <FormButton submit full-width class="mt-4" :disabled="loading">Sign up</FormButton>
+    <div
+      class="mt-3 text-xs flex items-center justify-center text-foreground-2 space-x-2"
+    >
+      <!-- 
+        Note the newsletter consent box is here because i got very confused re layout of the panel
+        and didn't figure out a better way to put it where i needed it to be
+       -->
+      <label for="newsletter" class="sxxr-only">
```
fabis94, line 53: Why not use `<FormInputCheckbox/>`?

packages/server/modules/shared/helpers/envHelper.ts
```diff
@@ -81,6 +81,22 @@ export function getOidcName() {
   return process.env.OIDC_NAME
 }
 
+export function getMailchimpConfig() {
+  if (
+    !process.env.MAILCHIMP_API_KEY ||
```
fabis94, line 86: How will this work locally? Should this just be disabled or should we set up actual mailchimp lists for dev purposes?

fabis94 (changes requested):

packages/frontend-2/components/auth/RegisterPanel.vue
```diff
@@ -63,6 +63,10 @@ graphql(`
   }
 `)
 
+const newsletterConsent = ref(false)
+
+provide('newsletterconsent', newsletterConsent)
```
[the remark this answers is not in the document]
didimitrie: It's tricky - i need the checkbox to be visually in the register with email block, but that var needs to be available in the third party call; this was an easier way to declare it up and inject it where in needed it without mutating props and stuff. I felt weird passing in a ref via props...

packages/frontend-2/components/auth/RegisterWithEmailBlock.vue
didimitrie: i needed to style it a bit more and didn't yet know how to publish a new version of the ui components.

packages/server/modules/shared/helpers/envHelper.ts
didimitrie: added an extra env var, we can ignore it locally

packages/server/modules/auth/services/mailchimp.ts
```diff
@@ -0,0 +1,46 @@
...
+      email_address: user.email,
+      merge_fields: {
+        EMAIL: user.email,
+        FNAME: first,
+        LNAME: second,
+        FULLNAME: user.name // NOTE: this field needs to be set in the audience merge fields
+      }
+    })
+  } catch (e) {
+    logger.warn(e, 'Failed to register user to newsletter.')
```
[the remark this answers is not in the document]
didimitrie: extra enabled env var added helps resolve this - we'll log only when the newsletter is explicitly on and something else goes wrong.

PR #1655, opened 2023-06-26
Feat (DUI3): Get local accounts from connectors

oguzhankoral:
## Description & motivation

Here I will do tests for integrations DUI3 with connectors.
Motivation is to understand how we will communicate with different connectors like SketchUp, Rhino, Revit, Archicad. So this PR will require ongoing reviews to keep code quality up while trying to find best way to handle communication.

### 🔼 SketchUp

Once `UI::HtmlDialog` in `DUI3View` created, we have method to execute scripts by calling `html_dialog.execute_script(...)` method. Problem with SketchUp case is to be able to send `hostApp` and it's callback object `sketchup` and callback function `exec` information after dialog showed. Not yet figured it out how we can send this information to html_dialog immediately after it loaded. For now we check existence of `window.sketchup.exec` to understand it is running in SketchUp or not.

### 🦏 Rhino

In Rhino we will change `window.location = myschema:navigation` to trigger callbacks as below.
```c#
var webView = new WebView();
webView.DocumentLoading += (sender, e) =>
{
	if (e.Uri.Scheme == "myscheme")
	{
		e.Cancel = true; // prevent navigation
		
		var path = e.Uri.PathAndQuery;
		if (path == "dosomething")
		{
			// do something..
		}
	}
};
```

### 🧱 Revit

Need to figure it out.

### 🧱 Archicad

Need to figure it out.

## Checklist:


packages/dui3/lib/common/generated/gql/graphql.ts
```diff
@@ -209,7 +209,7 @@ export type Comment = {
    * Legacy comment viewer data field
    * @deprecated Use the new viewerState field instead
    */
-  data?: Maybe<LegacyCommentViewerData>;
+  data?: Maybe<Scalars['JSONObject']>;
```
oguzhankoral, line 212: @fabis94 Is it normal that these lines in `packages/dui3/lib/common/generated/gql/graphql.ts` changed automatically? I haven't changed them.

didimitrie: yes this is normal. generally you can ignore all generated files.

PR #1732, opened 2023-07-27
feat(server): guests cannot own stream projects

gjedlicska:
## Checklist:

### DIFF
--- a/packages/server/modules/core/services/streams/streamAccessService.js
+++ b/packages/server/modules/core/services/streams/streamAccessService.js
@@ -2,7 +2,7 @@ const { authorizeResolver } = require(`@/modules/shared`)
 
 const { Roles } = require('@/modules/core/helpers/mainConstants')
 const { LogicError } = require('@/modules/shared/errors')
-const { ForbiddenError } = require('apollo-server-express')
+const { ForbiddenError, UserInputError } = require('apollo-server-express')
 const { StreamInvalidAccessError } = require('@/modules/core/errors/stream')
 const {
   addStreamPermissionsAddedActivity,
@@ -14,6 +14,8 @@ const {
   revokeStreamPermissions,
   grantStreamPermissions
 } = require('@/modules/core/repositories/streams')
+
+const { ServerAcl } = require('@/modules/core/dbSchema')
 
 /**
  * Check if user is a stream collaborator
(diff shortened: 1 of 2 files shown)


packages/server/modules/core/services/streams/streamAccessService.js
```diff
@@ -129,6 +131,13 @@ async function addOrUpdateStreamCollaborator(
 
   await validateStreamAccess(addedById, streamId, Roles.Stream.Owner)
 
+  // make sure server guests cannot be stream owners
+  if (role === Roles.Stream.Owner) {
+    const userServerRole = await ServerAcl.knex().where({ userId }).first()
```
fabis94, line 136: I'd prefer using the repository layer instead. You could re-use the `getStream()` function which also returns role metadata if you feed in a userId.

packages/server/modules/core/tests/streams.spec.ts
```diff
@@ -215,6 +219,35 @@ describe('Streams @core-streams', () => {
...
+        guestRoleEnabled: true
+      })
+
+      await addOrUpdateStreamCollaborator(
+        testStream.id,
+        guestGuy.id,
+        Roles.Stream.Owner,
+        userOne.id
+      )
+        .then(() => {
```
fabis94, line 244: The promise is awaited anyway, so instead of `then()`/`catch()` you can convert this to a standard try-catch

gjedlicska: i always get bamboozled by this. Not having a nice way of testing for a promise rejection is grinding my gears.

fabis94: > Not having a nice way of testing for a promise rejection is grinding my gears.

What do you mean by this? You should be able to convert that into this:

```ts
try {
   await addOrUpdateStreamCollaborator(
        testStream.id,
        guestGuy.id,
        Roles.Stream.Owner,
        userOne.id
      )
    throw new Error('This should have thrown')
} catch (e) {
  expect(err.message).to.include('Server guests cannot own streams')
}
```

gjedlicska: i mean in automate we have [this](https://github.com/specklesystems/speckle-automate/blob/519e4edcecdd51ad21b76285a0470020855af0d1/components/automate-app/tests/unit/authz/pipeline.spec.ts#L85) which is super nice

but ofcourse i can convert this into async / await

PR #1764, opened 2023-08-15
Feature - Frontend2 - Server Settings Pages

andrewwallacespeckle:
## Description & motivation
Adding the "Server Management" section to Frontend2. Should only be visible to admin users. 

Designs for this can be found [here](https://www.figma.com/file/WHdLVa7947kGoVXq2Dqp7I/%F0%9F%8E%A8-Miscellaneous-FE-design-(Admin%2C-webhook%2C-developer-settings)?type=design&node-id=114%3A9391&mode=dev)

Main elements are
- Avatar Menu Item visible to admin
- Server Management homepage with server stats
- Server Settings dialog
- Active Users page
- Invited Users page
- Projects page

## To-do before merge:
- Finish projects page - this has been parked until Fabians has had a chance to review the existing pages.
- Fix issue with the User Role select. When the user change role, but does not confirm it in the dialog, the dropdown should not show the new role. 
- Check Version function, sometime it appears as N/A causing a hydration mismatch. 

## Screenshots:
image
image
image
image

## Validation of changes:
No change to tests. I need to chat with Fabians to confirm the testing requirements. 

## Checklist:


packages/frontend-2/components/server-management/UserRoleSelect.vue
```diff
@@ -0,0 +1,53 @@
...
+const emit = defineEmits<{
+  (e: 'update:modelValue', v: ValueType): void
+}>()
+
+const { selectedValue } = useFormSelectChildInternals<ServerRoles>({
+  props: toRefs(props),
+  emit
+})
+
+const roleLookupTable = {
```
fabis94, line 41: If you put this in a helper and export it out, you can use it elsewhere as well (e.g. that Role dialog where you needed these names)

packages/frontend-2/lib/server-management/helpers/types.ts
```diff
@@ -0,0 +1,16 @@
+import { Get } from 'type-fest'
+import { UserListQuery } from '~~/lib/common/generated/gql/graphql'
+
+export type UserItem = NonNullable<Get<UserListQuery, 'admin.userList.items[0]'>>
+
+export interface CTA {
+  type: 'button' | 'link'
+  label: string
+  action: () => void | Promise<void>
```
fabis94, line 9: 1. `MaybeAsync<void>`
2. I saw the same type duplicated in one or more components, just import this there instead (the Button type as well, IIRC)

packages/frontend-2/middleware/admin.ts
```diff
@@ -0,0 +1,28 @@
...
+export default defineNuxtRouteMiddleware(async () => {
+  const client = useApolloClientFromNuxt()
+
+  const { data } = await client
+    .query({
+      query: activeUserQuery
+    })
+    .catch(convertThrowIntoFetchResult)
+
+  // If not logged in, redirect to login. If logged in but not an admin, redirect home.
```
fabis94, line 18: inaccurate comment

packages/frontend-2/middleware/admin.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (data?.activeUser?.role !== Roles.Server.Admin) {
+    return abortNavigation(
+      createError({
+        statusCode: 403,
+        message: 'You do not have access to this project'
```
fabis94, line 23: Inaccurate message

packages/frontend-2/pages/admin-only.vue
```diff
@@ -0,0 +1,8 @@
+<template>
```
fabis94, line 1: I'm guessing this was a test? If so, please delete it

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
+<template>
+  <div>
+    <Portal to="navigation">
+      <HeaderNavLink
+        :to="'/server-management'"
```
fabis94, line 5: If the value to a prop is a string literal, you can remove the `:` to make Vue not treat the value as JS code and just do `to="/server-management"`

Same goes for the other link

packages/frontend-2/components/server-management/Table.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <div class="mt-8 text-foreground">
```
fabis94, line 2: Not a good practice to add outer margins to a component, cause you may want to use it in different kinds of layouts, some of which may not need that margin or may need a different kind of margin.

Better to add this margin to the template where you actually use this component (`<Table class="mt-8"/>`)

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+    <Table
+      :headers="[
+        { id: 'name', title: 'Name' },
+        { id: 'email', title: 'Email' },
+        { id: 'emailState', title: 'Email State' },
+        { id: 'company', title: 'Company' },
+        { id: 'role', title: 'Role' }
+      ]"
+      :items="users"
+      :buttons="[{ icon: TrashIcon, label: 'Delete', action: openUserDeleteDialog }]"
```
fabis94, line 39: There's a TS typing error on `action`

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+      :column-classes="{
+        name: 'col-span-3',
+        email: 'col-span-3',
+        emailState: 'col-span-2',
+        company: 'col-span-2',
+        role: 'col-span-2'
+      }"
+      :overflow-cells="true"
+    >
+      <template #name="{ item }">
```
fabis94, line 49: Since `Table` is not supposed to be generic, you can properly type `item` (inside Table.vue) so that it's the correct type. Same goes for other templates here.

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+        <div class="flex items-center gap-2">
+          <Avatar :user="item" />
```
fabis94, line 51: There's a TS error here, cause of a type mismatch. This is because `Avatar` doesn't support `AdminUserListItem` objects at this point in time.

You can fix this like so:
image

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+
+      <template #company="{ item }">
+        {{ item.company }}
+      </template>
+
+      <template #role="{ item }">
+        <UserRoleSelect
+          :model-value="item.role"
+          @update:model-value="
+            (newRoleValue) => openChangeUserRoleDialog(item as UserItem, newRoleValue)
```
fabis94, line 81: Since `item` is going to be typed like `UserItem | ProjectItem` you can introduce a type guard like this so that you don't have to forcefully recast types and can capture scenarios where `item` is unexpectedly not `UserItem`

Example:
```ts
const asUser = (val: UserItem | ProjectItem): val is UserItem => {
  if (has(val, 'userId')) return true
  throw new Error('Unexpectedly item is not a user item! Wtf, we are in active-users.vue how did this happen?')
}

// then in your vue template
(newRoleValue) => openChangeUserRoleDialog(asUser(item), newRoleValue)
```

packages/frontend-2/components/server-management/ChangeUserRoleDialog.vue
```diff
@@ -0,0 +1,65 @@
...
+import Avatar from '~~/components/user/Avatar.vue'
+
+const emit = defineEmits<{
+  (e: 'update:open', val: boolean): void
+}>()
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  user: UserItem | null
```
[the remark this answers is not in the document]
fabis94: Never mind, I figured it out - it allows for null so that you don't have to unmount the dialog component when it gets closed and all of the CSS transitions work properly.

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+          verified
+          company
+        }
+      }
+    }
+  }
+`)
+
+const adminDeleteUser = graphql(`
+  mutation Mutation($userConfirmation: UserDeleteInput!) {
```
fabis94, line 183: Vague mutation name `Mutation`. Change it to something like `AdminPanelDeleteUser`

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+  convertThrowIntoFetchResult,
+  getFirstErrorMessage
+} from '~~/lib/common/helpers/graphql'
+
+definePageMeta({
+  middleware: ['admin']
+})
+
+const getUsers = graphql(`
+  query UserList($limit: Int!, $cursor: String, $query: String) {
```
fabis94, line 163: Vague query title. If we'd see this in our metrics it wouldn't help us understand where its coming from and what purpose does it have. Something like `AdminPanelUsersList` would be more descriptive.

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+  }
+`)
+
+const changeRoleMutation = graphql(`
+  mutation AdminChangeUseRole($userRoleInput: UserRoleInput!) {
+    userRoleChange(userRoleInput: $userRoleInput)
+  }
+`)
+
+const userToModify = ref<UserItem | null>(null)
```
fabis94, line 194: `X | null` can be shortened to `Nullable<X>`. `Nullable` comes from `@speckle/shared`. There's also `Optional<X>` for `X | undefined` amongst some other useful utility types in that package.

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+const deleteConfirmed = async () => {
+  if (!userToModify?.value?.email) {
+    return
+  }
+  const result = await adminDeleteUserMutation({
+    userConfirmation: { email: userToModify.value.email }
+  }).catch(convertThrowIntoFetchResult)
+  if (result?.data?.adminDeleteUser) {
+    closeUserDeleteDialog()
+    refetchUsers()
```
fabis94, line 252: Refetch is fine for now, but in the future it would be preferred to use post-mutation cache updates to just remove the deleted item from the cached list

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,332 @@
...
+      type: ToastNotificationType.Danger,
+      title: 'Failed to update role',
+      description: errorMessage
+    })
+  }
+  showChangeUserRoleDialog.value = false
+  userToModify.value = null
+}
+
+const handleSearchChange = (newSearchString: string) => {
```
fabis94, line 295: Redundant function, just use `searchUpdateHandler` directly

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
+<template>
+  <div>
+    <Portal to="navigation">
+      <HeaderNavLink
+        :to="'/server-management'"
```
fabis94, line 5: `to="/server-management"`

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+      <Card :server-info="projectData" />
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { graphql } from '~~/lib/common/generated/gql'
+import { useQuery } from '@vue/apollo-composable'
+import { ref, computed, onMounted, reactive } from 'vue'
+import Card, { CardInfo } from '../../components/server-management/Card.vue'
```
fabis94, line 49: Remove all manual imports of local vue components in all files, Nuxt auto-imports local components.

https://nuxt.com/docs/guide/concepts/auto-imports
https://nuxt.com/docs/guide/directory-structure/components#component-names

fabis94: + Make sure you always use absolute path imports

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+          {
+            text: 'Save',
+            props: { color: 'primary', fullWidth: true, outline: false },
+            onClick: saveSettings
+          }
+        ]"
+        @server-info-updated="fetchData"
+      />
+
+      <Card :server-info="userData" />
```
fabis94, line 39: TS typing errors here

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+      <div>
+        <h5 class="h4 font-bold">Your server at a glance</h5>
+      </div>
+    </div>
+
+    <div class="grid grid-cols-1 md:grid-cols-2 gap-6 mt-10">
+      <Card :server-info="serverData" @cta-clicked="showDialog = true" />
+
+      <SettingsDialog
+        ref="settingsDialog"
```
fabis94, line 21: Very rarely if ever you would need to use `refs` and the ability to invoke some internal component methods. In this case it does make sense, cause there's no other way to trigger submission, but I don't see the need to have buttons be injectable from outside anyway. You already know what the buttons are going to be inside the `SettingDialog.vue` component, so why not define them there and thus remove the need to expose submission outwards?

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+  tag_name: string
+}
+
+interface SettingsDialogRef {
+  onSubmit: () => void
+}
+
+const router = useRouter()
+const showDialog = ref(false)
+const settingsDialog = ref<SettingsDialogRef | null>(null)
```
fabis94, line 76: Nullable<X>

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+const latestVersion = ref<string | null>(null)
+const isLatestVersion = ref<boolean>(false)
```
fabis94, line 78: redundant type argument, it's already clear it's a boolean from the `false` value

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+
+const dataQuery = graphql(`
+  query Admin {
```
fabis94, line 81: vague query name

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+    admin {
+      serverStatistics {
+        totalPendingInvites
+        totalProjectCount
+        totalUserCount
+      }
+      inviteList {
+        totalCount
```
fabis94, line 89: You're asking for invite count twice, both in serverStatistics and inviteList

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+    }
+    serverInfo {
+      name
+      version
+    }
+  }
+`)
+
+const { result, refetch } = useQuery(dataQuery)
+const versionInfo = reactive({
```
fabis94, line 100: Why not just use the standard `ref()` here?

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+    icon: HomeIcon,
+    cta: {
+      type: 'button',
+      label: 'Manage',
+      action: async () => await router.push('/server-management/projects/')
+    }
+  }
+])
+
+onMounted(async () => {
```
fabis94, line 171: 1. Don't do this in `onMounted`, its going to cause hydration mismatch warnings.
2. You don't need to manually trigger `fetchData()` the initial load for queries happens automatically

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+      const data = (await response.json()) as GithubRelease
+      return data.tag_name
+    }
+  } catch (err) {
+    console.error('Error fetching latest version:', err)
+    return null
+  }
+}
+
+const fetchData = () => {
```
fabis94, line 207: redundant function

packages/frontend-2/pages/server-management/index.vue
```diff
@@ -0,0 +1,210 @@
...
+      'https://api.github.com/repos/specklesystems/speckle-server/releases/latest'
+    )
+    if (!response.ok) {
+      throw new Error(`HTTP error! status: ${response.status}`)
+    } else {
+      const data = (await response.json()) as GithubRelease
+      return data.tag_name
+    }
+  } catch (err) {
+    console.error('Error fetching latest version:', err)
```
fabis94, line 202: use logger

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
+<template>
+  <div>
+    <Portal to="navigation">
+      <HeaderNavLink
+        :to="'/server-management'"
```
fabis94, line 5: `to="/server-management"/` (check these everywhere)

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+        email: 'col-span-5',
+        invitedBy: 'col-span-4',
+        resend: 'col-span-3'
+      }"
+    >
+      <template #email="{ item }">
+        {{ item.email }}
+      </template>
+
+      <template #invitedBy="{ item }">
```
fabis94, line 64: It appears that this table is used for Users, Invites and Projects, so the type at the end is going to be `UserItem | InviteItem | ProjectItem` and thus just like in that other page, you'll need to create a type guard that ensures the type is properly `InviteItem` and either returns true or throws

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+        <div class="flex items-center gap-2">
+          <img
+            :src="item.profilePicture"
+            :alt="'Profile picture of ' + item.invitedBy"
+            class="w-6 h-6 rounded-full"
+          />
+          {{ item.invitedBy }}
```
fabis94, line 71: IIRC `invitedBy` is an object, not a string

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+        </div>
+      </template>
+
+      <template #resend="{ item }">
+        <button class="font-semibold text-primary" @click="resendInvitation(item)">
```
fabis94, line 76: Why not `FormButton`? If the designs differ from any of the existing FormButton options, I'd bring that up with Agi - is this really worth introducing an entirely new FormButton variation for?

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+          Resend Invitation
+        </button>
+      </template>
+    </Table>
+
+    <DeleteInvitationDialog
+      v-model:open="showDeleteInvitationDialog"
+      :user="user ?? userToModify"
```
fabis94, line 84: There are various TS errors in this file, one of which is that `user` doesn't exist

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+import Table from '../../components/server-management/Table.vue'
+import DeleteInvitationDialog from '../../components/server-management/DeleteInvitationDialog.vue'
+
+import { TrashIcon } from '@heroicons/vue/24/outline'
+
+definePageMeta({
+  middleware: ['admin']
+})
+
+const userToModify = ref<User | null>(null)
```
fabis94, line 118: Nullable<X>`

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+</template>
+
+<script setup lang="ts">
+import {
+  MagnifyingGlassIcon,
+  ChevronLeftIcon,
+  ChevronRightIcon
+} from '@heroicons/vue/20/solid'
+
+import Table from '../../components/server-management/Table.vue'
```
fabis94, line 109: 1. Relative imports
2. Manually imported vue components (Nuxt has auto-imports for local components)

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,272 @@
...
+  userToModify.value = null
+}
+
+const resendInvitation = (item) => {
+  console.log(item)
+}
+
+const showDeleteInvitationDialog = ref(false)
+
+export interface User {
```
fabis94, line 143: 1. Don't export things from Vue components
2. Use a type derived from the GQL query result type

packages/frontend-2/pages/server-management/projects.vue
```diff
@@ -0,0 +1,256 @@
...
+}
+
+const router = useRouter()
+
+const handleProjectClick = (project: Project) => {
+  router.push(`/projects/${project.id}`)
+}
+
+const getProjects = graphql(`
+  query Query($query: String, $orderBy: String, $limit: Int!, $visibility: String) {
```
fabis94, line 167: vague query name

packages/ui-components/src/components/layout/Dialog.stories.ts
```diff
@@ -47,6 +47,33 @@ export const Default: StoryObj = {
   }
 }
 
+export const HeaderAndFooter = {
```
fabis94, line 50: great that you added a story for this

packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -27,21 +27,35 @@
...
-            >
-              <div class="relative">
+            <DialogPanel :class="dialogPanelClasses">
+              <div
+                v-if="title"
+                class="flex items-center justify-center shadow p-4 relative z-10 bg-foundation"
+              >
+                <h4 class="text-2xl font-bold">{{ title }}</h4>
+              </div>
+              <button
```
fabis94, line 37: You should use `FormButton` wherever possible

packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -70,13 +85,23 @@ const props = defineProps<{
...
 const open = computed({
   get: () => props.open,
   set: (newVal) => emit('update:open', newVal)
 })
 
+const dialogPanelClasses = computed(() => {
+  return [
+    'transform rounded-lg overflow-hidden text-left shadow-xl transition-all relative',
+    widthClasses.value,
+    props.title || props.buttons ? 'bg-white' : 'bg-foundation'
```
fabis94, line 101: Why are we forcing it to be `bg-white`? I'd think that `bg-foundation` is fine already, but also `bg-white` isn't a semantic color and thus isn't good look in dark theme

fabis94 (changes requested):
@andrewwallacespeckle Added some comments. For some I didn't repeat myself when I saw the same issue in other files, so please try to fix recurring issues in all files.

I know that Invite/Project pages aren't done yet, but I did a cursory review anyway to point out some issues I saw there.

One general comment: Try to fix all TS/ESLint errors you can see in all files, otherwise CI won't let this PR through at the end. I know that you had some issues with `logger` not being typed properly by your IDE, but we can troubleshoot that at the end.

Regarding the 2 issues I had to check, i'm going to check them after this

packages/ui-components/src/components/layout/Dialog.vue
andrewwallacespeckle: The Dialog in the designs have a #ffffff background. The header and footer of the dialog is bg-foundation, so when these are added I made the background white.

I can revert this to just bg-foundation and mention it to Agi when she is back.

packages/ui-components/src/components/layout/Dialog.vue
andrewwallacespeckle: I'm not sure why this is highlighted as a change. I didn't change this, and its a <button> on main. I can take a look at changing it to FormButton if you want?

fabis94: Aren't these the new buttons you added to the dialog component?

fabis94: Ah right, that button already existed before. And it wasn't a FormButton because it was supposed to be a very simple closer button.

packages/ui-components/src/components/layout/Dialog.vue
andrewwallacespeckle: Fixed

packages/frontend-2/pages/server-management/projects.vue
andrewwallacespeckle: Updated to getAdminProjectsQuery

fabis94: Operation names should always be in PascalCase
Class/Type names should always be in PascalCase
Component props in templates should always be in kebab-case
Component names in templates should always be in PascalCase
Functions/constants/variables should always be in camelCase

Very rarely global constants can be in uppercase SNAKE_CASE

packages/frontend-2/components/server-management/CardRow.vue
```diff
@@ -0,0 +1,46 @@
...
+defineEmits<{
+  (e: 'cta-clicked', v: MouseEvent): void
+}>()
+
+defineProps<{
+  title: string
+  value: string
+  icon: ConcreteComponent
+  cta?: {
+    type: 'button' | 'link'
```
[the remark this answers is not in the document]
andrewwallacespeckle: This is to allow the Button style for the "Update Available" as it is in designs. It is a different button to the primary button in the other cardrows.

packages/frontend-2/pages/server-management/pending-invitations.vue
andrewwallacespeckle: Can you confirm this one for me, removing the single quotes breaks this for me.

packages/frontend-2/components/server-management/ChangeUserRoleDialog.vue
```diff
@@ -0,0 +1,60 @@
...
+  oldRole: ServerRoles | undefined
+  newRole: ServerRoles | undefined
+  buttons?: Array<{ text: string; props: Record<string, unknown>; onClick: () => void }>
```
fabis94, line 53: Why should these be passed in from the parent component? It's always going to be the same set of buttons

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,46 @@
+<template>
```
fabis94, line 1: Would be better to handle the actual GQL mutation from this component, otherwise if you want to use this dialog elsewhere you'll have to duplicate the mutation logic each time from the parent component.

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,46 @@
...
+
+const emit = defineEmits<{
+  (e: 'update:open', val: boolean): void
+}>()
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  invite: InviteItem | null
+  buttons?: Array<{ text: string; props: Record<string, unknown>; onClick: () => void }>
```
fabis94, line 39: Why should these be passed in from the parent component? It's always going to be the same set of buttons

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
```diff
@@ -0,0 +1,48 @@
+<template>
+  <LayoutDialog v-model:open="isOpen" max-width="sm" :title="title" :buttons="buttons">
```
fabis94, line 2: Would be better to handle the actual GQL mutation from this component, otherwise if you want to use this dialog elsewhere you'll have to duplicate the mutation logic each time from the parent component.

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
```diff
@@ -0,0 +1,48 @@
...
+
+const emit = defineEmits<{
+  (e: 'update:open', val: boolean): void
+}>()
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  project: ItemType | null
+  buttons?: Array<{ text: string; props: Record<string, unknown>; onClick: () => void }>
```
fabis94, line 41: Why should these be passed in from the parent component? It's always going to be the same set of buttons

packages/frontend-2/components/server-management/DeleteUserDialog.vue
```diff
@@ -0,0 +1,42 @@
+<template>
```
fabis94, line 1: Would be better to handle the actual GQL mutation from this component, otherwise if you want to use this dialog elsewhere you'll have to duplicate the mutation logic each time from the parent component.

packages/frontend-2/components/server-management/DeleteUserDialog.vue
```diff
@@ -0,0 +1,42 @@
...
+
+const emit = defineEmits<{
+  (e: 'update:open', val: boolean): void
+}>()
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  user: UserItem | null
+  buttons?: Array<{ text: string; props: Record<string, unknown>; onClick: () => void }>
```
fabis94, line 35: Why should these be passed in from the parent component? It's always going to be the same set of buttons

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+  <LayoutDialog v-model:open="isOpen" max-width="sm" :title="title" :buttons="buttons">
+    <div class="flex flex-col gap-6 text-sm text-foreground">
+      <p>Are you sure you want to delete the selected invitation?</p>
+      <div v-if="invite" class="flex flex-col gap-2">
+        <div class="flex gap-1">
+          <div class="w-20">Email:</div>
+          <strong>{{ invite?.email }}</strong>
```
fabis94, line 8: Redundant (& misleading) optional chaining operator, its already established that `invite` is defined on line 5

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
```diff
@@ -0,0 +1,48 @@
...
+import { isProject } from '~~/lib/server-management/helpers/utils'
+
+const emit = defineEmits<{
+  (e: 'update:open', val: boolean): void
+}>()
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  project: ItemType | null
```
fabis94, line 40: Why is this `ItemType` (Project or User or Invite)? It's a "delete project" dialog, passing in invites and/or users is not a valid use case. Additionally, because of this you need the extraneous `isProject` checks in the template

packages/frontend-2/components/server-management/SettingsDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+
+const isOpen = computed({
+  get: () => props.open,
+  set: (newVal) => emit('update:open', newVal)
+})
+
+const emailRules = [isEmail]
+const requiredRule = [isRequired]
+
+const onSubmit = handleSubmit(async () => {
```
fabis94, line 144: Doesn't appear that the changes made in this mutation to `serverInfo` fields are actually refetched or manually updated in the GQL local cache. I see there's a `server-info-updated` event that is handled in the parent component, but that one only refetches a couple of `serverInfo` fields - name and version.

So I assume that after closing the dialog, if there's any component on the page that renders server info again, it will show old/stale values

packages/frontend-2/components/server-management/SettingsDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+    triggerNotification({
+      type: ToastNotificationType.Danger,
+      title: 'Saving failed',
+      description: `Failed to update server info`
+    })
+  }
+})
+
+watch(isOpen, (newVal, oldVal) => {
+  if (newVal && !oldVal) {
```
fabis94, line 174: Reduce nesting for clarity

E.g.:
```ts
if (!newVal || !oldVal) return
if (!result.value?.serverInfo) return

name.value = ....
```

packages/frontend-2/components/server-management/Table.vue
```diff
@@ -0,0 +1,106 @@
...
+        </div>
+      </div>
+      <div
+        class="divide-y divide-outline-3 h-full overflow-visible"
+        :class="{ 'pb-32': overflowCells }"
+      >
+        <div
+          v-for="item in items"
+          :key="item.id"
+          class="relative grid grid-cols-12 items-center gap-6 px-4 py-1 min-w-[900px] bg-white"
```
fabis94, line 24: There still seem to be various non-semantic styles that look broken in dark mode, please validate everything looks OK in both themes
image

packages/frontend-2/components/server-management/Table.vue
```diff
@@ -0,0 +1,106 @@
...
+          :style="{ paddingRight: paddingRightStyle }"
+          :class="{ 'cursor-pointer hover:bg-primary-muted': !!props.onRowClick }"
+          tabindex="0"
+          @click="handleRowClick(item)"
+          @keydown.enter.space="handleRowClick(item)"
```
fabis94, line 29: You can use `keyboardClick` from `~/lib/common/helpers/accessibility.ts` for these kinds of use cases

packages/frontend-2/components/server-management/Table.vue
```diff
@@ -0,0 +1,106 @@
...
+            <div :class="getClasses(column.id, colIndex)" tabindex="0">
+              <slot :name="column.id" :item="item">
+                <div class="text-gray-900 font-medium order-1">
+                  {{ (item as any)[column.id] }}
+                </div>
+              </slot>
+            </div>
+          </template>
+          <div class="absolute right-0 flex items-center p-0 pr-0.5">
+            <div v-for="(button, btnIndex) in buttons" :key="btnIndex" class="p-1">
```
fabis94, line 41: button.label is probably a better key, there's never gonna be multiple buttons with the same label

packages/frontend-2/components/server-management/Table.vue
```diff
@@ -0,0 +1,106 @@
...
+  const columnClass = props.columnClasses[column]
+
+  if (colIndex === 0) {
+    return `bg-transparent py-3 pr-5 ${columnClass}`
+  }
+  return `lg:p-0 ${columnClass}`
+}
+
+const handleRowClick = (item: UserItem | ProjectItem | InviteItem) => {
+  props.onRowClick?.(item as ItemType)
```
fabis94, line 104: Unnecessary type cast, item is already `ItemType`. Additionally you can shorten the type signature for this function's only argument to just be `ItemType`

packages/frontend-2/components/server-management/UserRoleSelect.vue
```diff
@@ -0,0 +1,55 @@
+<template>
+  <FormSelectBase
+    v-model="selectedValue"
+    :items="roles"
+    :label="'Role'"
+    :name="'userRole'"
```
fabis94, line 6:
```suggestion
    label="Role"
    name="userRole"
```

fabis94: https://laracasts.com/discuss/channels/vue/when-to-use-colon-when-passing-data-to-props?page=1&replyId=294073

packages/frontend-2/components/server-management/SettingsDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+          placeholder="Owner"
+          show-label
+        />
+        <FormTextInput
+          v-model="adminContact"
+          label="Admin Email"
+          name="adminEmail"
+          placeholder="Admin Email"
+          show-label
+          :rules="emailRules"
```
fabis94, line 40: Currently this prevents you from leaving the field empty or typing in something like "n/a"

packages/frontend-2/components/server-management/SettingsDialog.vue
fabis94: You can test this out yourself. Make a change to one of the fields that aren't refetched/manually updated in the cache like "Admin contact". Then close the dialog and open it up again - it's going to show the old/stale value.

packages/frontend-2/composables/logging.ts
```diff
@@ -1,9 +1,10 @@
 import { NuxtApp } from '#app'
 import { Optional } from '@speckle/shared'
 import { buildFakePinoLogger } from '~~/lib/core/helpers/observability'
+import type pino from 'pino'
 
 export const useLogger = () => {
-  return useNuxtApp().$logger
+  return useNuxtApp().$logger as pino.Logger
```
fabis94, line 7: Check if you even need this anymore, I wouldn't be surprised if $logger was unknown for you because of the mismatched TS version used by Volar before we fixed that

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,396 @@
...
+    >
+      <template #name="{ item }">
+        <div class="flex items-center gap-2">
+          <UserAvatar v-if="isUser(item)" :user="item" />
+          {{ isUser(item) ? item.name : '' }}
+        </div>
+      </template>
+
+      <template #email="{ item }">
+        {{ isUser(item) ? item.email : '' }}
```
fabis94, line 55: This column should either be truncated or wider with padding on all sides
image

packages/frontend-2/pages/server-management/active-users.vue
```diff
@@ -0,0 +1,396 @@
...
+const isCurrentUser = (userItem: UserItem) => {
+  return userItem.id === activeUser.value?.id
+}
+
+const {
+  result: extraPagesResult,
+  fetchMore: fetchMorePages,
+  variables: resultVariables,
+  onResult
+} = useQuery(getUsers, queryVariables)
```
fabis94, line 227: Search is busted, it appears that something is setting the actual KeyboardEvent object instead of the new search string into `searchString`

image

fabis94, line 227: Search is busted, it appears that something is setting the actual KeyboardEvent object instead of the new search string into `searchString`

image
image

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,313 @@
...
+    })
+  }
+}
+
+const {
+  result: extraPagesResult,
+  fetchMore: fetchMorePages,
+  variables: resultVariables,
+  onResult
+} = useQuery(getInvites, () => ({
```
fabis94, line 275: https://github.com/specklesystems/speckle-server/pull/1764#discussion_r1294459319

validate this in other files as well

packages/frontend-2/pages/server-management/projects.vue
```diff
@@ -0,0 +1,329 @@
...
+      title: 'Failed to delete project',
+      description: errorMessage
+    })
+  }
+}
+
+const {
+  result: extraPagesResult,
+  fetchMore: fetchMorePages,
+  variables: resultVariables,
```
fabis94, line 289: Same issue with busted GQL query (check dev tools network tab) when invoking a search

fabis94 (changes requested):

packages/frontend-2/components/server-management/Card.vue
```diff
@@ -0,0 +1,23 @@
...
+        :value="info.value"
+        :cta="info.cta"
+        :icon="info.icon"
+      />
+    </div>
+  </div>
+</template>
+
+<script lang="ts" setup>
+import CardRow from './CardRow.vue'
```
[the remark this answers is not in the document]
fabis94: + its unnecessary because of nuxt component auto imports

andrewwallacespeckle:
Hi @fabis94. I've been through and made all suggested changes.

I'm having trouble with a TS error in SettingsDialog. I wonder if you could take a look? 

Let me know if there is anything else.

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+import { useGlobalToast, ToastNotificationType } from '~~/lib/common/composables/toast'
+import { adminDeleteProject } from '~~/lib/server-management/graphql/mutations'
+import { getProjects } from '~~/lib/server-management/graphql/queries'
+import { Exact, InputMaybe } from '~~/lib/common/generated/gql/graphql'
+
+const props = defineProps<{
+  title: string
+  open: boolean
+  project: ProjectItem | null
+  resultVariables:
```
fabis94, line 49: _(I originally posted this in DeleteInvitationDialog, but I've already fixed that one so I'm posting this here)_

This is a bit smelly, you shouldn't need to feed in a specific set of variables just for the mutation cache update. Additionally this isn't going to fully achieve what you're after anyway, because your cache update function will only update the cache for these variables, but any other instances of the `inviteList` query (e.g. with a different search query) will remain stale.

Here I console logged out the Apollo cache and you can see there are multiple instances of the `inviteList` query:
image

A better option is to use `cache.modify` or the util I created - `modifyObjectFields` which essentially wraps `cache.modify`, because it allows you to update multiple variations of the same field. So you can essentially update all instances of the `inviteList` query with a single call, without needing to know the variables beforehand, which is perfect for this scenario.

The complication here is that `AdminInviteList` is not uniquely identifiable - it doesn't have an ID field. And thus you can't just directly update all inviteLists like so: 

```js
cache.modify({
  id: 'AdminInviteList', // there's no ID 
  fields: {
    totalCount: (old) => old + -1
  }
})
```

What you can do, however, is identify `ROOT_QUERY` which is the root of the Query graph. And from there you can modify its `admin` field, which will in turn contain all variations of the `inviteList` field.

This is a bit complicated, so I've already done this for you in the invites page, but you'll need to do the same in other pages as well.

See the `modifyObjectFields` call in `DeleteInvitationDialog.vue` for an example. Cache updates can be complicated, so if something doesn't work right you can utilize the browser debugger to see what's going on.

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+  const result = await adminDeleteMutation(
+    {
+      ids: [projectId]
+    },
+    {
+      update: (cache, { data }) => {
+        if (data?.streamsDelete) {
+          // Remove invite from cache
+          cache.evict({
+            id: getCacheId('AdminUserListItem', projectId)
```
fabis94, line 85: This couldn't have ever worked, because the list has items of type `Project` not `AdminUserListItem`. I've noticed this issue in multiple dialogs (e.g. Delete invites dialog, which I fixed). Please check that your cache updates work properly.

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,151 @@
...
+  getCacheId,
+  getFirstErrorMessage,
+  modifyObjectFields
+} from '~~/lib/common/helpers/graphql'
+import { AdminInviteList } from '~~/lib/common/generated/gql/graphql'
+
+const props = defineProps<{
+  open: boolean
+  title: string
+  invite: InviteItem | null
```
fabis94, line 49: Notice - I removed the need to feed in query variables

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,151 @@
...
+      update: (cache, { data }) => {
+        if (data?.inviteDelete) {
+          // Remove invite from cache
+          const cacheId = getCacheId('ServerInvite', inviteId)
+          cache.evict({
+            id: cacheId
+          })
+
+          // Modify 'admin' field of ROOT_QUERY so that we can modify all `inviteList` instances
+          modifyObjectFields<undefined, { [key: string]: AdminInviteList }>(
```
fabis94, line 84: Notice - Here I fixed the cache updates

packages/frontend-2/components/server-management/DeleteProjectDialog.vue
fabis94: Make sure you fix this in all of your dialogs

packages/frontend-2/components/server-management/SettingsDialog.vue
```diff
@@ -0,0 +1,185 @@
...
+    })
+    triggerNotification({
+      type: ToastNotificationType.Success,
+      title: 'Successfully saved',
+      description: 'Your server settings have been saved.'
+    })
+    isOpen.value = false
+    emit('server-info-updated')
+  } catch (error) {
+    logger.error(error)
```
fabis94, line 165: This is unreachable code. In `updateServerInfoAndCache` you're suppressing error throws with `convertThrowIntoFetchResult`, but you're also not returning the results from that function so there's no way for you to check if the operation succeeded or not.

fabis94 (changes requested):
@andrewwallacespeckle I noticed some cache updates not working and being implemented incorrectly. I stopped there so let me know when those problems are resolved and then I'll check the rest.

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,151 @@
...
+                  totalCount: Math.max(
+                    0,
+                    (value[field]?.totalCount || 0) - removedCount
+                  )
+                }
+              }
+
+              return newVal
+            },
+            { fieldNameWhitelist: ['admin'] }
```
fabis94, line 115: This bit is the one that ensures we only update Query.admin, not some other ROOT_QUERY fields.

packages/frontend-2/components/server-management/DeleteInvitationDialog.vue
```diff
@@ -0,0 +1,151 @@
...
+
+          // Modify 'admin' field of ROOT_QUERY so that we can modify all `inviteList` instances
+          modifyObjectFields<undefined, { [key: string]: AdminInviteList }>(
+            cache,
+            ROOT_QUERY,
+            (_fieldName, _variables, value, details) => {
+              // Find all `inviteList` fields (there can be multiple due to differing variables)
+              const inviteListFields = Object.keys(value).filter(
+                (k) =>
+                  details.revolveFieldNameAndVariables(k).fieldName === 'inviteList'
```
fabis94, line 91: This needs to be done cause `k` can be a value like `inviteList:{"query":"jon"}`, so the string contains the fieldName and variables serialized together. That's why you need `resolveFieldNameAndVariables` which will separate them and you can get just the field name on its own that way.

fabis94 (changes requested):
@andrewwallacespeckle Code looks good for the most part, the remaining issues are mostly bugs  I noticed while playing around with the changes

andrewwallacespeckle:
Should all be fixed now, let me know if there is anything else.

fabis94 (approved):
Looks good! LGTM 🎆

PR #1765, opened 2023-08-15
Feature: 165 Clearer loaded version viewer

andrewwallacespeckle:
[LINK TO NOTION TICKET
](https://www.notion.so/speckle/We-Like-Tasks-36d66170974b46298bb3d9b02a241437?p=d49892f9ff4b4f7192dc77dc64ad719a&pm=s)

## Description & motivation
An improvement suggested by Fabians, this task was the make it clearer which version you were currently viewing. 

I changed the colours of the hover and active states of the "currently viewing" item to make this a bit more obvious to the user. I also added a new component <ActiveVersionCard /> which will appear below the VersionCard when a version is selected.

## Changes:

Changes are in 3 files:
- ActiveVersionCard.vue (new file)
- ModelCard.vue
- VersionCard.vue

## Screenshots:
Before:
image

After:
image

## Checklist:


packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,42 @@
...
+      <div class="italic text-foreground-2">
+        {{ timeAgo(version.createdAt) }}
+      </div>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import dayjs from 'dayjs'
+import relativeTime from 'dayjs/plugin/relativeTime'
```
fabis94, line 20: Unnecessary, we have a Nuxt plugin that already does this: `plugins/dayjs.ts`

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,42 @@
...
+defineProps({
+  version: {
+    type: Object as () => Version,
+    required: true
+  }
+})
+
+dayjs.extend(relativeTime)
+
+const timeAgo = (date: string) => {
```
fabis94, line 38: This could be a computed. That would be better from a performance perspective, because a computed is re-calculated only when any of its dependencies change (e.g. props.createdAt), but if you just use a method in your template, it's going to be invoked every single time the component gets re-rendered.

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,42 @@
...
+import relativeTime from 'dayjs/plugin/relativeTime'
+import { EyeIcon } from '@heroicons/vue/24/solid'
+
+interface Version {
+  sourceApplication?: string | null
+  message?: string | null
+  createdAt: string
+}
+
+defineProps({
```
fabis94, line 29:
```ts
defineProps<{
  version: Version
}>()
```

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,42 @@
...
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import dayjs from 'dayjs'
+import relativeTime from 'dayjs/plugin/relativeTime'
+import { EyeIcon } from '@heroicons/vue/24/solid'
+
+interface Version {
```
fabis94, line 23: This is good enough for now, but in the future I suggest looking into GQL fragment-oriented prop design: https://the-guild.dev/blog/unleash-the-power-of-fragments-with-graphql-codegen

Essentially, you can define a GQL fragment specifying the fields you're interested in, then use the auto-generated TS type for that fragment in your `defineProps` and also specify that fragment in any queries that load the data for this component, ensuring that even if you change fragment fields, both the GQL query and the component will have these changes.

fabis94: We already have many examples of this in FE2, btw, for example `ProjectModelsPageHeader_Project` in `Header.vue`

packages/frontend-2/components/viewer/resources/ModelCard.vue
```diff
@@ -40,11 +43,21 @@
         </div>
         <div
           v-else
-          class="flex flex-none items-center space-x-2 text-xs font-bold opacity-0 transition-opacity group-hover:opacity-100"
+          :class="`${
+            showVersions ? 'text-white' : ''
```
fabis94, line 47: Validate if any non-semantic colors really can't be replaced by semantic ones and if they look good in all themes

packages/frontend-2/components/viewer/resources/ModelCard.vue
```diff
@@ -40,11 +43,21 @@
...
+          } flex flex-none items-center space-x-2 text-xs font-bold opacity-80 transition-opacity group-hover:opacity-100`"
         >
           <ChevronUpIcon class="h-4 w-4" />
         </div>
       </div>
+      <ActiveVersionCard
+        v-if="loadedVersion && showVersions"
+        :version="{
```
fabis94, line 55: Why not just pass in `loadedVersion`?

packages/frontend-2/components/viewer/resources/VersionCard.vue
```diff
@@ -1,9 +1,10 @@
 <template>
   <button
     :class="`bg-foundation group relative block w-full space-y-2 rounded-md pb-2 text-left transition ${
-      clickable ? 'hover:bg-primary-muted' : 'cursor-default'
+      clickable
+        ? 'hover:bg-gray-100 dark:hover:bg-gray-700'
```
fabis94, line 5: non-semantic colors

packages/frontend-2/components/viewer/resources/ModelCard.vue
```diff
@@ -109,6 +123,7 @@ import {
   useInjectedViewerRequestedResources
 } from '~~/lib/viewer/composables/setup'
 import { useDiffUtilities } from '~~/lib/viewer/composables/ui'
+import ActiveVersionCard from './ActiveVersionCard.vue'
```
fabis94, line 126: 1. always use absolute path imports
2. don't import local components, nuxt has auto-imports for those

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,42 @@
+<template>
+  <div class="bg-primary-muted flex gap-2 py-2 px-4">
+    <EyeIcon v-tippy="'Currently viewing'" class="h-4 w-4 text-primary mt-0_5" />
```
fabis94, line 3: I noticed that this icon is focusable and gets a black outline on focus, please remove it

fabis94: Also, if the version message is very long, the icon becomes very small. You probably need to set `flex-shrink: 0` onto it.

fabis94 (changes requested):

packages/frontend-2/components/viewer/resources/VersionCard.vue
didimitrie: Non semantic colors can be okay at this stage if it looks the part (and we're missing a semantic color to do what it's needed here), and we can think of making them "official".

andrewwallacespeckle:
I have swapped to using all semantic colours, and used opacity to give some visual hierarchy to the content. This could be a good compromise?

I also made a slight change to the span that wraps timeAgoCreatedAt. I removed the semi-bold until it is selected, again I think it helps visually, but happy to swap it back if anyone disagrees. 

image

image

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
```diff
@@ -0,0 +1,40 @@
...
+  message?: string | null
+  createdAt: string
+}
+
+const props = defineProps<{
+  version: Version
+}>()
+
+const timeAgo = computed(() => {
+  if (!props.version.createdAt) return 'no message'
```
fabis94, line 37: i'm confused by this, why would a missing `createdAt` prop mean - 'no message'? and also - createdAt can't be missing, it's a non-nullable string

fabis94 (changes requested):
left 1 comment, otherwise looks good. once that bit is fixed it's more or less ready for merging

packages/frontend-2/components/viewer/resources/ActiveVersionCard.vue
andrewwallacespeckle: Thanks for catching that. You're right, the check was unnecessary. I've removed it and re-pushed

fabis94 (approved):
LGTM

PR #1792, opened 2023-09-14
Feature: FE2 - Project Webhooks Page

andrewwallacespeckle:
## Description & motivation
Adding the Webhooks Page to FE2. Accessible through an updated "settings" card on project page. Within the "Manage" dialog, there is now "Webhooks, with a cta "Manage Webhooks".

The page lists all webhooks for this project in a table. There is a "Create Webhook" button which launches a dialog window to do this. A similar Dialog is launched when the edit a Webhook by clicking on the Pencil icon. You can also delete a Webhook from here, or toggle its state.

## Changes:
- Change "Team" card to "Settings" card
- Add Webhook section to the Dialog launched from the settings card. 
- Added new Webhook Page
- Added new "Create Webhook" Dialog
- Added new "Edit Webhook" Dialog
- Added new "Delete Webhook" Dialog

#### Storybook
- Added Generic Table Component
- Added BadgeSelected Component

## Screenshots:
image
image
image

## Checklist:


andrewwallacespeckle:
There are a few issues to be addressed before this is ready for a PR.

1. I'm getting a hydration mismatch on the main webhooks page. I've done some troubleshooting without much luck.
2. I'm having issues with the cache updates on Delete, Edit and Create. 
3. There is a backend problem where the table re-orders on refresh.
4. There is a backend problem around the Webhook "Secret".

packages/frontend-2/components/project/webhooks-page/CreateWebhookDialog.vue
```diff
@@ -0,0 +1,140 @@
+<template>
```
fabis94, line 1: I was able to create a webhook without any events selected. Not only that, I was able to create a webhook without any fields filled out at all! Please add validation to all forms to prevent this from happening.

packages/frontend-2/components/project/webhooks-page/DeleteWebhookDialog.vue
```diff
@@ -0,0 +1,128 @@
...
+    },
+    {
+      update: (cache, { data }) => {
+        if (data?.webhookDelete) {
+          const cacheId = getCacheId('Webhook', webhookId)
+          cache.evict({
+            id: cacheId
+          })
+
+          modifyObjectFields<undefined, { [key: string]: WebhookCollection }>(
```
fabis94, line 73: There's no way this could ever work, because `ROOT_QUERY` has no such `webhooks` field. I suggest always validating cache mutators with a debugger to ensure they actually work.

Considering that we want to update `Project.webhooks` and `Project` can be uniquely identified (has an `id` field)_, there's no reason for us to start at the ROOT_QUERY level. You can invoke `modifyObjectFields` on `getCacheId('Project', projectId)`.

packages/frontend-2/components/project/webhooks-page/CreateWebhookDialog.vue
```diff
@@ -0,0 +1,140 @@
...
+import { createWebhookMutation } from '~~/lib/projects/graphql/mutations'
+import { WebhookCreateInput } from '~~/lib/common/generated/gql/graphql'
+import { useGlobalToast } from '~~/lib/common/composables/toast'
+
+const requiredRule = [isRequired]
+
+const props = defineProps<{
+  open: boolean
+  streamId: string
+  webhookId?: string
```
fabis94, line 71: Unused prop

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,271 @@
...
+    />
+
+    <ProjectWebhooksPageEditWebhookDialog
+      v-model:open="showEditWebhookDialog"
+      :webhook="webhookToModify"
+    />
+
+    <ProjectWebhooksPageCreateWebhookDialog
+      v-model:open="showNewWebhookDialog"
+      :url="urlValue"
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:117]: Most of these props don't even exist

fabis94 (changes requested):

packages/frontend-2/components/project/webhooks-page/CreateWebhookDialog.vue
andrewwallacespeckle: Working on triggers validation now

andrewwallacespeckle: Should be working now on origin

packages/frontend-2/lib/projects/helpers/utils.ts
```diff
@@ -0,0 +1,10 @@
+import { WebhookItem } from '~~/lib/projects/helpers/types'
+import { has } from 'lodash-es'
+import { TableItemType } from '@speckle/ui-components'
+
+export const isWebhook = (
```
fabis94, line 5: Pointless type guard

packages/frontend-2/lib/projects/helpers/types.ts
```diff
@@ -0,0 +1,13 @@
+import { Get } from 'type-fest'
+import { ProjectWebhooksQuery } from '~~/lib/common/generated/gql/graphql'
+
+export type WebhookItem = NonNullable<
+  Get<ProjectWebhooksQuery, 'project.webhooks.items[0]'>
+>
+
+export type FormValues = {
```
fabis94, line 8: Way too vague of a name considering that this is defined in a `projects` helper (there are so many different forms relating to projects)

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
+<template>
+  <div>
+    <Portal to="navigation">
+      <HeaderNavLink :to="projectRoute(projectId)" :name="projectName"></HeaderNavLink>
+      <HeaderNavLink
+        :to="`${projectRoute(projectId)}/webhooks`"
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:6]: Create and use a standard route helper for this, similar to `projectRoute`

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+
+const openDocs = () => {
+  window.open('https://speckle.guide/dev/server-webhooks.html', '_blank')
+}
+
+const handleWebhookCreated = () => {
+  refetchWebhooks()
+}
+
+const onChange = async (item: TableItemType<WebhookItem>, newValue: boolean) => {
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:222]: vague function name - on what kind of change? it appears that this is only used for `enabled` changes, not any other kind, so `onEnabledChange` would already be a more descriptive name

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+          icon: TrashIcon,
+          label: 'Delete',
+          action: openDeleteWebhookDialog,
+          class: 'text-red-500'
+        }
+      ]"
+    >
+      <template #enabled="{ item }">
+        <FormSwitch
+          :model-value="(item.enabled as boolean)"
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:55]: It's not a boolean tho, it's a `boolean | null | undefined`. You don't need to do type casting here (which should be done as rarely as possible), you can just convert it to a bool w/ `!!item.enabled`

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+              <InformationCircleIcon
+                v-if="getHistoryStatus(item) === 'noEvents'"
+                class="opacity-40"
+              />
+              <CheckCircleIcon
+                v-if="getHistoryStatus(item) === 'called'"
+                class="text-success"
+              />
+              <XCircleIcon
+                v-if="
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:73]: Can be shortened to `['error', 'alert'].includes(getHistoryStatus(item))`

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+const showDeleteWebhookDialog = ref(false)
+const showEditWebhookDialog = ref(false)
+const showNewWebhookDialog = ref(false)
+
+const projectId = computed(() => route.params.id as string)
+const projectName = computed(() => pageResult.value?.project?.name || 'Unknown Project')
+const webhooks = computed<WebhookItem[]>(() => {
+  return (
+    pageResult.value?.project?.webhooks?.items?.map(
+      (webhook) => webhook as WebhookItem
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:168]: I've adjusted the backend to no longer allow the array to contain `null` values so this should no longer be necessary

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+          <span class="text-foreground opacity-50 text-sm truncate">
+            {{ item.url }}
+          </span>
+        </div>
+      </template>
+
+      <template #triggers="{ item }">
+        <div :class="{ 'opacity-60': !item.enabled }">
+          {{
+            (item.triggers as string[])
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:93]: I've adjusted the backend to no longer allow the array to contain `null` values so this should no longer be necessary

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+      case 3:
+        return 'error'
+    }
+  } else {
+    return 'noEvents'
+  }
+}
+
+const getHistoryStatusInfo = (item: TableItemType<WebhookItem>) => {
+  if (isWebhook(item)) {
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:190]: Redundant condition (get rid of isWebhook everywhere)

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+    const recentHistory = item.history?.items?.[0]
+    if (recentHistory) {
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:192]: This can be shortened to a ternary condition: `return recentHistory ? recentHistory.statusInfo : 'No events yet'`

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+
+const openEditWebhookDialog = (item: TableItemType<WebhookItem>) => {
+  if (isWebhook(item)) {
+    webhookToModify.value = item
+    showEditWebhookDialog.value = true
+  }
+}
+
+const openDocs = () => {
+  window.open('https://speckle.guide/dev/server-webhooks.html', '_blank')
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:215]: Make `FormButton` a proper link instead with the `to` prop

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,269 @@
...
+  } else {
+    triggerNotification({
+      type: ToastNotificationType.Danger,
+      title: 'Failed to update webhook',
+      description: getFirstErrorMessage(result?.errors)
+    })
+  }
+}
+
+const { result: pageResult, refetch: refetchWebhooks } = useQuery(
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:263]: Composables like this should always be invoked at the top

packages/ui-components/src/components/form/Switch.vue
```diff
@@ -0,0 +1,45 @@
...
+    </HeadlessSwitch>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref, watch } from 'vue'
+import { Switch as HeadlessSwitch } from '@headlessui/vue'
+import { CheckIcon, XMarkIcon } from '@heroicons/vue/24/solid'
+
+const props = defineProps<{
```
fabis94, line 25: Pretty much the entire script setup block can just be simplified to a single `defineModel<boolean>()` call.

packages/ui-components/src/components/form/select/BadgeSelected.stories.ts
```diff
@@ -0,0 +1,62 @@
+import { Meta, StoryObj } from '@storybook/vue3'
+import BadgeSelected from './BadgeSelected.vue'
```
fabis94, line 2: Weird component name: FormSelectBadgeSelected. Why not `FormSelectBadges`?

packages/ui-components/src/components/form/select/BadgeSelected.vue
```diff
@@ -0,0 +1,73 @@
...
+    :multiple="multiple"
+    :items="items"
+    :label="label"
+    :name="name"
+    :help="help"
+    :rules="rules"
+  >
+    <template #something-selected="{ value }">
+      <ul class="flex flex-wrap gap-1.5 text-xs">
+        <li v-for="item in isArrayValue(value) ? value : [value]" :key="item">
```
fabis94, line 13: Item can be an object, I don't think it can be a proper key. Additionally, you aren't passing in a `by` prop to `FormSelectBase` which is used to get the actual ID of each item (see any other selector that uses FormSelectBase), so equality checks might not work at times.

The `by` field specifies which property can you find the key/id of each item in. Considering that we allow any kind of structure for items, we should make `by` a prop and allow devs to feed in their own values.

packages/ui-components/src/components/layout/Table.stories.ts
```diff
@@ -0,0 +1,222 @@
...
+
+export const WithButtons: StoryObj = {
+  ...Default,
+  args: {
+    ...Default.args,
+    buttons: [
+      {
+        icon: DocumentDuplicateIcon,
+        label: 'Duplicate',
+        action: (item: TableItemType) => console.log('Duplicate', item),
```
fabis94, line 205: Instead of console logging, use `action()` from `'@storybook/addon-actions'` so that these messages can be actually seen within storybook (google for more info)

packages/ui-components/src/components/layout/Table.vue
```diff
@@ -0,0 +1,109 @@
...
+  const columnClass = getHeaderClasses(column)
+
+  if (colIndex === 0) {
+    return `bg-transparent py-3 pr-5 px-1 ${columnClass}`
+  }
+  return `lg:p-0 px-1 ${columnClass}`
+}
+
+const handleRowClick = (item: T) => {
+  if ('id' in item) {
```
fabis94, line 105: unnecessary check, item is already of type `T` which has to extend from `{id: string}`

packages/ui-components/src/components/layout/Table.vue
```diff
@@ -0,0 +1,109 @@
...
+              />
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts" generic="T extends {id: string}, C extends string">
```
fabis94, line 57: I highly advise looking into Generics generally to make it easier for you to grasp Vue generics: https://www.typescriptlang.org/docs/handbook/2/generics.html

packages/ui-components/src/helpers/common/validation.ts
```diff
@@ -77,3 +77,17 @@ export const stringContains =
       return match.test(val) ? true : message
     }
   }
+
+export const isUrl: GenericValidateFunction<string> = (value) => {
+  if (VALID_HTTP_URL.test(value)) {
+    return true
+  }
+  return 'Invalid URL'
```
fabis94, line 85: Include the word `Value` in error messages, cause it will get replaced with the actual name/label of the input. So something like `Value is not a valid URL`. Same goes for the other function.

packages/ui-components/src/helpers/common/validation.ts
```diff
@@ -77,3 +77,17 @@ export const stringContains =
...
+}
+
+export const isItemSelected: GenericValidateFunction<[]> = (val) => {
```
fabis94, line 88: I'm confused why TS doesn't throw an error for `[]` as a type, because it's not clear what kind of array is it supposed to be. Since you explicitly don't care about the type of the individual array items you can just do `unknown[]`

https://dmitripavlutin.com/typescript-unknown-vs-any/

packages/ui-components/src/helpers/layout/components.ts
```diff
@@ -14,3 +14,7 @@ export type LayoutMenuItem<I extends string = string> = {
   disabled?: boolean
   disabledTooltip?: string
 }
+
+export type TableItemType<T = unknown> = {
```
fabis94, line 18: unnecessary

fabis94 (changes requested):

packages/frontend-2/components/project/page/team/dialog/Webhooks.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <Portal to="navigation">
+    <HeaderNavLink :to="projectRoute(project.id)" :name="project.name"></HeaderNavLink>
```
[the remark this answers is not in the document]
andrewwallacespeckle: This is an error. I must have pasted this into the wrong file.

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/projects/[id]/webhooks/index.vue:197]: Is it possible to make this target blank still? I've added external, but it doesn't open a new tab.

Comment (fabis94) [reply packages/frontend-2/pages/projects/[id]/webhooks/index.vue]: Yes, even tho it's not an explicit prop of `FormButton`, it'll work due to attribute inheritance - it'll be attached onto the final `<a>` element that is rendered

https://vuejs.org/guide/components/attrs.html

packages/frontend-2/components/project/webhooks-page/CreateWebhookDialog.vue
```diff
@@ -0,0 +1,142 @@
...
+    <form @submit="onSubmit">
+      <div class="flex flex-col gap-6">
+        <FormTextInput
+          v-model="formData.url"
+          label="URL"
+          help="A POST request will be sent to this URL when this webhook is triggered"
+          name="hookUrl"
+          show-label
+          show-required
+          :rules="(isRequired, isUrl)"
```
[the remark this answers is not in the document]
fabis94: image

It appears that it just invokes both expressions and returns the last one, so in reality you're only feeding in `isUrl`

packages/frontend-2/lib/projects/helpers/types.ts
fabis94: These naming rules might seem unimportant, but they do in fact meaningfully improve readability & maintainability of the project, especially when there's multiple people working on it.

If you ever have the time I suggest checking out a classic programming book called "Clean Code" (at least the 1st part of the book). It's oriented around Java, but the same rules apply to all programming generally. 

Dunno how legal it is, but I found this PDF on the 1st page of Google lol https://thixalongmy.haugiang.gov.vn/media/1175/clean_code.pdf

packages/ui-components/src/components/form/select/BadgeSelected.vue
andrewwallacespeckle: I think I have got this one done correctly, but I won't mark it as resolved until you have checked it just incase I did something wrong. 

I added a prop to BadgeSelected component for by and set the key of the v-for to :key="item[by]"

In FE2, I added by="id"

packages/ui-components/src/components/form/Switch.vue
andrewwallacespeckle: Another one to check before I mark as resolved. I replaced the script (except imports) with:
const enabled = defineModel<boolean>()

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
```diff
@@ -0,0 +1,252 @@
...
+const projectName = computed(() => pageResult.value?.project?.name || 'Unknown Project')
+const webhooks = computed<WebhookItem[]>(() => {
+  return pageResult.value?.project?.webhooks?.items || []
+})
+
+const getHistoryStatus = (item: WebhookItem) => {
+  const recentHistory = item.history?.items?.[0]
+  if (recentHistory) {
+    switch (recentHistory.status) {
+      case 1:
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/webhooks/index.vue:180]: 1. It's good practice to always add a `default:` case, in case a new value gets introduced that doesn't get handled by any of the existing cases. Since you don't have one, the function can theoretically also just return `undefined`
2. I suggest using enums for these kinds of cases when you need a set of named constants: https://www.typescriptlang.org/docs/handbook/enums.html

The entire function can be simplified to something like this:

```ts
enum HistoryStatus {
  NoEvents = 0,
  Alert = 1,
  Called = 2,
  Error = 3,
}

const getHistoryStatus = (item: WebhookItem) => {
  const recentHistory = item.history?.items?.[0]
  const status = recentHistory?.status as Optional<HistoryStatus> // we were careful to assign the same numeric values for each status in the enum, so it's OK to do this kind of type cast
  return status || HistoryStatus.NoEvents
}
```

packages/ui-components/src/components/form/Switch.stories.ts
```diff
@@ -0,0 +1,33 @@
...
+  render: (args) => ({
+    components: { Switch },
+    setup() {
+      return { args }
+    },
+    template: `
+      <div>
+        <Switch 
+          v-bind="args" 
+          @update:modelValue="args.modelValue = $event"
```
fabis94, line 25: This is good enough, but in the future I suggest using `ctx.updateArgs()` to update the value (see other stories), cause otherwise the Storybook internal React state won't show that updated value in the Storybook UI (e.g. in the params table)

packages/ui-components/src/components/form/select/Badges.stories.ts
```diff
@@ -0,0 +1,62 @@
...
+    setup: () => {
+      return { args }
+    },
+    template: `
+    <div class="flex justify-center h-72 w-full">
+      <Badges v-bind="args" @update:modelValue="onModelUpdate" class="w-full"/>
+    </div>
+    `,
+    methods: {
+      onModelUpdate(val: Array<SingleItem>) {
```
fabis94, line 37: This is what I meant in the other comment in Switch.stories.ts: It's kind of verbose and hacky, but unfortunately its the best way Storybook offers us to properly have bidirectional binding for Storybook args.

fabis94 (changes requested):

packages/frontend-2/components/project/webhooks-page/CreateOrEditDialog.vue
```diff
@@ -0,0 +1,230 @@
...
+
+    const result = await updateMutation(
+      {
+        webhook: {
+          id: webhookId,
+          streamId: props.webhook.streamId,
+          url: url.value,
+          description: description.value,
+          secret: secret.value?.length ? secret.value : null,
+          triggers: webhookFormValues.triggers.map((t) => t.text)
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/1792/commits/af413697da2b8ac1cba630e61773af0cb17de72e

packages/frontend-2/pages/projects/[id]/webhooks/index.vue
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/projects/[id]/webhooks/index.vue]: https://github.com/specklesystems/speckle-server/pull/1792/commits/982dec4428dca938c24b5f8e104b3560de3dc075

packages/ui-components/src/components/form/Switch.stories.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/1792/commits/0c37aa146691e9d5b46eac68742c005ac6ed1c64

packages/ui-components/src/components/form/select/Badges.stories.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/1792/commits/0c37aa146691e9d5b46eac68742c005ac6ed1c64

packages/frontend-2/components/project/webhooks-page/CreateOrEditDialog.vue
```diff
@@ -0,0 +1,230 @@
...
+      url: url.value,
+      secret: secret.value?.length ? secret.value : null,
+      streamId: props.streamId || '',
+      triggers: webhookFormValues.triggers.map((t) => t.text),
+      enabled: true
+    }
+
+    createWebhook({ webhook: webhookInput })
+      .then(() => {
+        emit('webhook-created')
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/1792/commits/21aa8598eb215cc911c223e6e17cc7f93ff4a69c

fabis94 (approved):

fabis94 (approved):

PR #1840, opened 2023-10-28
fix(helm): resolves mismatched template names in "frontend_2" templates and network policies.

thegartrell:
## Description & motivation

Running `helm upgrade` with kubernetes network policies and frontend_2 enabled causes the error below. This is happening because the names of the partials defined in `frontend_2/_helpers.tpl`  are declared in kebab case (`frontend-2.name`, `frontend-2.labels`, `frontend-2.selectorLabels`), while the names used in the networkpolicy.kubernetes.yml and networkpolicy.cilium.yml are in snake case (`frontend_2.name`, `frontend_2.labels`, `frontend_2.selectorLabels`), so the values defined in the helper aren't picked up by the downstream templates.

```
Error: UPGRADE FAILED: template: speckle-server/templates/frontend_2/networkpolicy.kubernetes.yml:6:11: executing "speckle-server/templates/frontend_2/networkpolicy.kubernetes.yml" at <include "frontend_2.name" $>: error calling include: template: no template "frontend_2.name" associated with template "gotpl"
```

It's also confusing that the values.yaml uses snake case for fe2 (`frontend_2`) but the namespaces in the subtemplates/partials in _helpers.tpl suddenly switch to kebab case (`frontend-2`). This may have contributed to the confusion that caused the first issue.

Also, a few kubernetes network policy templates are only implemented if the network policy for the fileimport_service is enabled. Each service's network policy should be implemented if it is enabled for its respective service in the values.yaml.  For example, the kubernetes network policy for frontend_2 is only implemented if `.Values.fileimport_service.networkPolicy.enabled` is true. This means that the network policy for frontend_2 can only be enabled by enabling the network policy for fileimport_service, meanwhile the networkPolicy toggle for frontend_2 doesn't actually implement anything.

I don't think this is the intended behavior. See diagram in [PR909].(https://github.com/specklesystems/speckle-server/pull/909)

## Changes:
 - Replaced references to `fileimport_service` in the kubernetes network policies of frontend, frontend_2, monitoring, preview_service, server, tests and webhook_service to reference the correct service/resource.
 - Changed namespace of all defined templates in `frontend_2` templates to use snake case (`frontend_2`) instead of kebab case (`frontend-2`) to match the values.yaml and the template names used in the existing network policies.

## Validation of changes:

The test job's network policy was also dependent on the networkPolicy for fileimport_service. This has been updated to `.Values.test.networkPolicy.enabled`.

It's not ideal, but you could add tests to `run_tests.py` that attempt to curl pods that should be unreachable from the test deployment if certain network policies are enabled. Happy to work on this or a better solution.

## Checklist:

- [x ] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?

### DIFF
--- a/utils/helm/speckle-server/templates/frontend/networkpolicy.kubernetes.yml
+++ b/utils/helm/speckle-server/templates/frontend/networkpolicy.kubernetes.yml
@@ -1,5 +1,5 @@
 {{- if not .Values.frontend_2.enabled -}}
-{{- if (and (.Values.fileimport_service.networkPolicy.enabled) (eq .Values.networkPlugin.type "kubernetes")) -}}
+{{- if (and (.Values.frontend.networkPolicy.enabled) (eq .Values.networkPlugin.type "kubernetes")) -}}
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
--- a/utils/helm/speckle-server/templates/frontend_2/_helpers.tpl
+++ b/utils/helm/speckle-server/templates/frontend_2/_helpers.tpl
@@ -1,7 +1,7 @@
 {{/*
 Expand the name of the chart.
 */}}
-{{- define "frontend-2.name" -}}
+{{- define "frontend_2.name" -}}
 {{- default "speckle-frontend-2" .Values.frontend_2.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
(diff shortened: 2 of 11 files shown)


thegartrell:
@iainsproat it looks like the CI is failing because the Apollo API key isn't being sent to test-server. I'm guessing because I don't have permissions. Should I have just left it alone for you to merge?

In trying to trigger a rebuild, I managed to sync my main with your updates and then backed them out by doing a force push on my main. It's pretty sloppy. Please let me know if you'd like me to just open a fresh PR from a fresh fork. Apologies for the confusion.

thegartrell:
Makes total sense. Thanks @iainsproat !

PR #1877, opened 2023-11-20
feat(objects server): routes /objects to dedicated server deployments

iainsproat:
## Description & motivation

At present all server traffic is routed to the server pods. This traffic varies widely in its characteristics depending on endpoint. The `/objects/*` endpoints are used to upload and download traffic via a REST endpoint. This is often 'bursty', with the 'burst' peaks coinciding with the publishing or receiving of large models.

These bursts may saturate all the database connections available to a single pod, which increases latency and causes the objects to queue in memory. Other endpoints served by the container are then starved of database connections, which can cause performance degradation - this includes the `/graphql` endpoint used by the frontend. Therefore the performance degradation is likely to be perceived by users.

## Changes:

This PR routes `/objects` to a separate deployment of speckle-server containers. All other traffic continues to be served by the original deployment.

This PR has the benefit of:
- prevents /objects traffic from saturating nodes serving other traffic.
- should help improve performance, especially perceived performance via frontend implications.

## Checklist:

### DIFF
--- a/utils/helm/speckle-server/templates/objects/deployment.yml
+++ b/utils/helm/speckle-server/templates/objects/deployment.yml
@@ -9,7 +9,7 @@ spec:
   replicas: {{ .Values.objects.replicas }}
   selector:
     matchLabels:
-      app: speckle-server
+      app: speckle-objects
       project: speckle-server
   strategy:
     type: RollingUpdate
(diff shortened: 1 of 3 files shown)


gjedlicska:
double check this @iainsproat, but i think `/api` is also serving object data, that also should be served by the objects backend.

gjedlicska (approved):

PR #1885, opened 2023-11-23
tests(auth): refactor to prevent test pollution

iainsproat:
## Description & motivation

Tests for authentication of apps are currently dependent on the order in which they are run, resulting in test pollution and brittleness (i.e. it's impossible to run a subsequent test in isolation as it requires preceding tests).

## Changes:

- each test registers its own app and does not rely on an app being registered in a prior test
- where a test requires an authorization code, it creates it itself and does not rely on it being created by a prior test
- reduces verbosity by placing default app tests within the existing forEach loop.

## Checklist:

### DIFF
--- a/packages/server/modules/auth/tests/apps.spec.js
+++ b/packages/server/modules/auth/tests/apps.spec.js
@@ -418,48 +418,6 @@ describe('Services @apps-services', () => {
       .catch((err) => expect(err.message).to.equal('Access code not found.'))
   })
 
-  it('Should receive an expiring fine-grained token in exchange for the authorization code', async () => {
-    const myTestApp = await createApp({
-      name: cryptoRandomString({ length: 10 }),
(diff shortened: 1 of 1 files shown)


packages/server/modules/auth/tests/apps.spec.js
```diff
@@ -19,6 +19,7 @@ const {
 const { Scopes } = require('@/modules/core/helpers/mainConstants')
 const { updateDefaultApp } = require('@/modules/auth/defaultApps')
 const knex = require('@/db/knex')
+const cryptoRandomString = require('crypto-random-string')
```
fabis94, line 22: Good to see more of these being cleaned up 👍 There's a lot of cross test pollution in `server` 🥲

fabis94 (approved):

PR #1889, opened 2023-11-27
FE2 Viewer - Add Measure Mode

andrewwallacespeckle:
## Description & motivation
Adding "Measure Mode" to FE2 viewer. The viewer API allows the following interactions around Measuring:
- Enable/Disable Measure Mode
- Change Units of measurement
- Change Precision of units (decimal places)
- Snap to Vertices Option
- Toggle Measurement Type from "Point to Point" to "Perpendicular"

Agi did some Figma designs for this here:
https://www.figma.com/file/W5nqvh21qxfoIru052eReb/%F0%9F%91%93-Viewer-mobile-and-desktop?type=design&node-id=650-812&mode=design&t=NLIoqO1TZ3O2k77W-0

There are a couple of issues that came up during development, but these are issues with the viewer rather than the FE. 
- We have no way of knowing when the user has selected a measurement. Because of this, the button will never be disabled. A compromise was to rename the button to "Delete Selected" to make it more obvious how this worked. 
- There is no way to cancel a measurement. If you accidentally click, you are forced to finish the measurement, and then delete it. It would be good if there was a way cancel the measurement with "esc".
- Not directly related to this PR, but the keyboard shortcuts are not working across the viewer. 

## Changes:
- Add Measure Mode Panel
- Add API interactions
- Add new Radio Group Component in ui-components

## To-do before merge:
- We have called this "Measure Mode", as whenever you open the menu, the cursor starts measuring on click. However, "alt+m" is used to trigger the Model panel. I have made the shortcut "d" for measure mode, but open to suggestions if something makes more sense. There is also the possibility of renaming this panel "Dimensions" or something so we can have a logical shortcut. To be confirmed with Dim. 
- Currently, we have labeled the checkbox "Snap to Vertices". This was deemed the most accurate. For Dim - Are you happy with this terminology or have anything more appropriate? Do you think our users will need a tooltip explaining this function?

## Screenshots:
image
image

## Checklist:


andrewwallacespeckle:
@didimitrie Couple of quick questions for you in the "To-do before merge" section if you get a minute.
If not, I'll go ahead and merge once approved. The terms can be updated in future if needed.

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -339,3 +339,30 @@ export function useThreadUtilities() {
 
   return { closeAllThreads, open, isOpenThread }
 }
+
+export function useMeasurementUtilities() {
+  const { instance } = useInjectedViewer()
```
fabis94, line 344: These utilities are interacting with the viewer directly, foregoing the viewer state. This is different from all of the other integrations between FE2 and Viewer are implemented and cause the following problems:

1. Interacting with the viewer from FE2 requires you to imperatively trigger functions onto the viewer instance. We want the integration to be declarative so that it's idiomatic to Vue apps. Instead of imperatively triggering measurements on/off, it would be more idiomatic to have a `areMeasurementsOn: boolean` reactive ref that you can read and write to. In this case you've encapsulated all of the imperative logic within your utilities, which is good, cause then devs don't have to do that manually, but all of this still exists outside of the declarative/reactive viewer state and thus can't be watched, mutated idempotently etc.
2. Since none of this is a part of the viewer state, we can't read the current state (are measurements currently enabled?) for UI purposes or for serialization purposes. So we can't accurately show that measurements are enabled in the UI or serialize this data for comments/follow mode (e.g. how we serialize and save other viewer settings with comments, so that when u open the comment u get the same viewer settings).

So, ideally, this should be implemented like the rest of the FE2<->Viewer integration:

1. Add new properties to the viewer state relevant to this integration, probably under  `state.ui`. Maybe something like: `{ui: {measurements: {areEnabled: Ref<boolean>, options: Ref<undefined | MeasurementOptions>}}}`. This will allow you to declaratively read the current state and also write new measurement settings.
2. In `postSetup.ts` actually create a composable for your integration and invoke it in `useViewerPostSetup()`. You can look at how other integrations are set up, but you would probably define `watch`ers on the new state properties, that would send the new state values to the viewer instance (instance.enableMeasurements/instance.setMeasurementOptions), only if they've changed.
4. You can leave a `useMeasurementUtilities()` that you created in this file, but it should only change those new state refs inside of them, instead of interacting with the viewer directly.

So essentially: 1. Only create the refs themselves in the viewer state, adjust TS types etc. 2. Actually implement the logic for reading/writing from/to these refs in postSetup 3. If needed, create utilities in `viewer/composables/` that simplify working with the viewer state.

fabis94: Once we have this data in the viewer state, we can also adjust the serialization logic so that these settings are stored with comments or sent to followers in spotlight mode, but I'm not sure if we need that currently

fabis94 (changes requested):
see comment i left

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -297,6 +318,9 @@ const projectionShortcut = ref(
 const sectionBoxShortcut = ref(
   `Section Box (${getKeyboardShortcutTitle([ModifierKeys.AltOrOpt, 'b'])})`
 )
+const measureShortcut = ref(
+  `Measure Mode (${getKeyboardShortcutTitle([ModifierKeys.AltOrOpt, 'd'])})`
```
andrewwallacespeckle, line 322: Open for suggestions on a better keyboard shortcut for this. I chose D for "Dimensions"

packages/server/modules/comments/services/data.ts
andrewwallacespeckle, line 1: Not too sure on what this is file is doing, but had to add this to resolve a TS error

packages/ui-components/src/components/form/select/Base.vue
andrewwallacespeckle, line 1: Change we did on a call - to fix the mount-menu-on-body bug

packages/frontend-2/lib/viewer/composables/setup.ts
```diff
@@ -75,6 +77,11 @@ export type LoadedThreadsMetadata = NonNullable<
 
 export type LoadedCommentThread = NonNullable<Get<LoadedThreadsMetadata, 'items[0]'>>
 
+interface MeasurementState {
+  enabled: boolean
```
fabis94, line 81: Ideally these 2 properties would be 2 different refs, cause now a change to `options` will mark the entire state as updated and possibly trigger code that's only interested in updates to `enabled`.

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -339,3 +339,33 @@ export function useThreadUtilities() {
 
   return { closeAllThreads, open, isOpenThread }
 }
+
+export function useMeasurementUtilities() {
+  const state = useInjectedViewerState()
+
+  const measurementSettings = state.ui.measurements.value
```
fabis94, line 346: This isn't reactive, you're just storing the unreactive settings (the thing you take out of .value on a ref is not reactive, only the ref itself is reactive) into a basic JS const. It's always going to be the same value, even if `state.ui.measurements` actually changes.

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -339,3 +339,33 @@ export function useThreadUtilities() {
...
+  }
+
+  const setMeasurementOptions = (options: MeasurementOptions) => {
+    if (measurementSettings) {
+      state.ui.measurements.value.options = options
+    }
+  }
+
+  const removeMeasurement = () => {
+    if (state.viewer.instance && state.viewer.instance.removeMeasurement) {
```
fabis94, line 361: if (state.viewer.instance?.removeMeasurement) {

packages/frontend-2/components/viewer/measurements/UnitSelect.vue
```diff
@@ -0,0 +1,54 @@
...
+import { ref, computed } from 'vue'
+
+const emit = defineEmits(['update:modelValue'])
+
+const props = defineProps<{
+  modelValue: string
+  name?: string
+}>()
+
+const UnitDisplayNames: Record<string, string> = {
```
fabis94, line 31: camelCase not PascalCase, the value isn't a class/type/component

packages/frontend-2/components/viewer/measurements/UnitSelect.vue
```diff
@@ -0,0 +1,54 @@
...
+  mi: 'Miles'
+}
+
+function getFullUnitName(unit: string): string {
+  return UnitDisplayNames[unit] || unit
+}
+
+const fullUnitName = computed(() => getFullUnitName(props.modelValue))
+
+const units = ref(['mm', 'cm', 'm', 'km', 'in', 'ft', 'yd', 'mi'])
```
fabis94, line 48:
```
const unitDisplaynames = ref(...)
const units = computed(() => Object.keys(unitDisplayNames.value))
```

Otherwise you have multiple sources of truth when it comes to what the units are. A single source of truth is always better for maintenance (more DRY)

packages/shared/src/viewer/helpers/state.ts
```diff
@@ -3,6 +3,19 @@ import type { MaybeNullOrUndefined, Nullable } from '../../core/helpers/utilityT
 import type { PartialDeep } from 'type-fest'
 import { UnformattableSerializedViewerStateError } from '../errors'
 
+declare enum MeasurementType {
```
fabis94, line 6: The `declare` is unnecessary, that keyword is usually only used in typing files when typing untyped JS code (.d.ts files)

packages/shared/src/viewer/helpers/state.ts
```diff
@@ -73,6 +86,10 @@ export type SerializedViewerState = {
     }
     explodeFactor: number
     selection: Nullable<number[]>
+    measurements: Nullable<{
```
fabis94, line 89: `enable` doesn't need to be nullable, it can be false by default. Otherwise you're introducing extra complexity - not only do you have to think about what happens when `enable` is true or false, but also what happens if its `null`

IMO `enable` should be a bool, and the options can be nullable.

fabis94: Also in the prop name `enableMeasurements`, the `Measurements` suffix is redundant, it's already clear that its about measurements from the parent property name

packages/ui-components/src/components/form/RadioGroup.vue
```diff
@@ -0,0 +1,53 @@
...
+import { ref, watchEffect } from 'vue'
+import type { ConcreteComponent } from 'vue'
+
+interface Option {
+  title: string
+  icon?: ConcreteComponent
+}
+
+const props = defineProps<{
+  label: string
```
fabis94, line 39: There's no way to pass in the current `selected` state, which is a must have for a reusable form component

packages/ui-components/src/components/form/RadioGroup.vue
```diff
@@ -0,0 +1,53 @@
...
+  options: Option[]
+}>()
+
+const emit = defineEmits(['update:selected'])
+
+const selected = ref<Option | null>(props.options[0] || null)
+
+watchEffect(() => {
```
fabis94, line 47: Why `watchEffect` instead of just `watch`? I've never had a need to use it and it's essentially a watcher that triggers whenever any of the refs used inside of its callback are updated

packages/ui-components/src/components/form/RadioGroup.vue
```diff
@@ -0,0 +1,53 @@
...
+  title: string
+  icon?: ConcreteComponent
+}
+
+const props = defineProps<{
+  label: string
+  options: Option[]
+}>()
+
+const emit = defineEmits(['update:selected'])
```
fabis94, line 43: Use `defineModel` to enable devs to pass in an existing `selected` state and also be able to emit out an update event when writing to the model ref.

fabis94: Also - instead of the model holding `Option` it should just hold the ID of the option (a string), otherwise, if you want to preselect an option, you need to have its entire `Option` object which contains irrelevant data for selection like the icon it has.

This means that you should also introduce an `id` prop in the `Option` interface

fabis94: Also - there doesn't seem to be any `vee-validate` support here, no way to set the `name` of the form field, feed in rules or ensure `vee-validate` can even read/write the selected state.

(Not sure, but you might not be able to use `defineModel` and might have to define the prop and event manually)

Read more on how to implement vee-validate support: https://vee-validate.logaretm.com/v4/examples/custom-inputs/
And feel free to look at other components like FormSelectBase to see useful props to add and how to integrate into vee-validate

packages/ui-components/src/components/form/RadioGroup.stories.ts
```diff
@@ -0,0 +1,65 @@
+import type { Meta, StoryObj } from '@storybook/vue3'
```
fabis94, line 1: 1. The `Actions` panel in storybook is empty even tho events are being fired when clicking on the options. Ensure you set this up correctly.
2. You should use the `args['update:modelValue'](val); ctx.updateArgs({ ...args, modelValue: val })` pattern to ensure that the updated property also shows up updated in the storybook controls panel
3. You should add stories for: Pre-configuring a specific option; Validating the input w/ vee-validate; Showing a required badge (see other inputs for common stories)

fabis94 (changes requested):

packages/ui-components/src/components/form/RadioGroup.vue
andrewwallacespeckle: I've removed the new RadioGroup component. I created a new component called Radio which is a lot closer to how the other items work. It is based on checkbox so should be a lot more consistent.

packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -115,7 +115,9 @@ export function useStateSerialization() {
           : null,
         lightConfig: { ...state.ui.lightConfig.value },
         explodeFactor: state.ui.explodeFactor.value,
-        selection: state.ui.selection.value?.toArray() || null
+        selection: state.ui.selection.value?.toArray() || null,
+        measurementOptions: state.ui.measurementOptions.value,
```
fabis94, line 119: Why not just?

```
measurement: {
  options: Ref<...>,
  enabled: Ref<...>
}
```

Otherwise 1) they're not grouped together in a single object and 2) they repeat the same redundant "measurement" prefix twice

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -717,6 +717,33 @@ function useDiffingIntegration() {
...
+        instance.enableMeasurements(newVal)
+      }
+    },
+    { immediate: true }
+  )
+
+  watch(
+    () => measurementOptions.value,
+    (newMeasurementState) => {
+      if (newMeasurementState) {
```
fabis94, line 739: Compare w/ previous state w/ lodash's `isEqual`

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -339,3 +339,29 @@ export function useThreadUtilities() {
...
+
+export function useMeasurementUtilities() {
+  const state = useInjectedViewerState()
+
+  const enableMeasurements = (enabled: boolean) => {
+    state.ui.measurementsEnabled.value = enabled
+  }
+
+  const setMeasurementOptions = (options: MeasurementOptions) => {
+    if (state.ui.measurementOptions.value) {
```
fabis94, line 351: Why this restriction?

fabis94 (changes requested):
Only a couple of minor issues remain, I think we can merge this tomorrow

fabis94 (dismissed):

fabis94:
Approved, but there's some TS typing issues that remain: https://app.circleci.com/pipelines/github/specklesystems/speckle-server/6541/workflows/bf51f5d5-294e-4dee-ba73-3aac11424e23/jobs/65274

fabis94 (approved):

PR #1894, opened 2023-12-01
FE2 - Title Bar Improvements

andrewwallacespeckle:
## Description & motivation
Part of the feedback received from SOM was that they placed a lot of importance on the naming of models. They had an issue where we had a max-width and truncate set on the Model name in breadcrumbs. When a model name was longer than the max-width, we obscured a lot of the model name which made it hard for them getting context on where the nested model is. This happened, even when there was plenty of screen space in the title bar. 

There will be a larger project in future to look into further improvements for the breadcrumbs. 

## Changes:
Most of the edits are in:
- HeaderNavBar
- HeaderNavLink

The model name now has no max-width, and will fill the available space. The earlier items in the breadcrumb will also truncate on smaller screens. 

There is also a change in:
- HeaderNavNotifications

I fixed an alignment issue, and scaled down the notification ping by 25%. 

## Screenshots:
Before:
image

After:
image
image

### DIFF
--- a/packages/frontend-2/components/header/NavLink.vue
+++ b/packages/frontend-2/components/header/NavLink.vue
@@ -3,12 +3,12 @@
     <NuxtLink
       :to="to"
       class="flex gap-1 items-center text-sm ml-0.5"
-      active-class="text-primary font-bold active-class"
+      active-class="text-primary font-bold group is-active"
     >
       <div v-if="separator">
         <ChevronRightIcon class="flex w-4 h-4" />
       </div>
-      <div class="maybe-truncate">
+      <div class="group-[.is-active]:truncate">
         {{ name || to }}
       </div>
     </NuxtLink>
@@ -31,11 +31,3 @@ defineProps({
   }
 })
 </script>
-
-<style>
-.active-class .maybe-truncate {
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-}
-</style>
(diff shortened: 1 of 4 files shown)


andrewwallacespeckle:
I have added a hacky fix for this. Adding shrink-0 helped, but it needed a bit more for it to only truncate the last item.

packages/frontend-2/components/header/NavLink.vue
```diff
@@ -31,3 +31,11 @@ defineProps({
   }
 })
 </script>
+
+<style>
+.active-class .maybe-truncate {
+  overflow: hidden;
```
fabis94, line 37: You can leave this as is, but in the future you can use Tailwind's `@apply XXX` directives, to construct new classes out of existing Tailwind classes.

```
.foo {
  @apply truncate font-bold text-foreground
}
```

packages/frontend-2/components/header/NavLink.vue
```diff
@@ -1,14 +1,14 @@
 <template>
-  <div class="transition text-foreground hover:text-primary-focus">
+  <div class="text-foreground hover:text-primary-focus transition last:truncate">
     <NuxtLink
       :to="to"
-      class="flex items-center text-sm"
-      active-class="text-primary font-bold"
+      class="flex gap-1 items-center text-sm ml-0.5"
+      active-class="text-primary font-bold active-class"
```
fabis94, line 6: Weird class name, the `class` suffix is def redundant. Doesn't matter that much tho

fabis94 (approved):

fabis94 (approved):

PR #1897, opened 2023-12-01
Fix unnecessary scroll

andrewwallacespeckle:
## Description & motivation
We have an absolutely positioned header bar. The main body is relatively positioned, and without adding margins, it will sit under the header bar. 

We currently have a mt-20 to manage this, but it has a side-effect where on pages where there isn't enough content to fill the browser window, the mt-20 adds a scroll, with nothing to scroll to. 

You can see this on latest.speckle.systems. Go to a model with 4 or less (or 0) versions. Hover on the model, and click the blue version button. This page should not have a scroll, but it currently does. 

## Changes:
Remove mt-20

Add div with h-20 to Header Bar component.

### DIFF
--- a/packages/frontend-2/components/header/NavBar.vue
+++ b/packages/frontend-2/components/header/NavBar.vue
@@ -32,6 +32,7 @@
     </div>
     <PopupsSignIn v-if="!activeUser" />
   </nav>
+  <div class="h-14" />
 </template>
 <script setup lang="ts">
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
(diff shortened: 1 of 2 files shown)


packages/frontend-2/components/header/NavBar.vue
```diff
@@ -32,6 +32,7 @@
     </div>
     <PopupsSignIn v-if="!activeUser" />
   </nav>
+  <div class="h-14" />
```
fabis94, line 35: You shouldn't ever have multiple root elements in a Vue component, it can break Nuxt navigation and cause other weird hard to debug issues

PR #1904, opened 2023-12-12
feat: clearer email verification flow & related error messages

fabis94:


packages/frontend-2/components/auth/PasswordResetPanel.vue
```diff
@@ -20,12 +20,16 @@
       </div>
     </template>
     <template #footer>
-      <FormButton submit full-width :disabled="loading">Send reset e-mail</FormButton>
+      <div class="flex flex-col space-y-2">
```
andrewwallacespeckle, line 23: This isn't worth changing, it won't change anything in this case, and partly preference, but I almost exclusively use gap rather than space now. 

Space will work on non flex/grid items, but it was a workaround for safari not supporting gap as far as I know. All browsers now do, so I normally stick with gap. 

The main difference is that it will only add a gap when there is more than 1 child element, whereas space will add the same padding to each child, regardless of amount of children.

andrewwallacespeckle (approved):
All looks good on the frontend

andrewwallacespeckle (approved):

PR #1905, opened 2023-12-12
FE2 UI minor improvements

andrewwallacespeckle:
## Description & motivation
Agi and I have been having weekly calls on making improvements to the FE2 UI. A lot of the changes have come from the feedback provided by SOM.

## Changes:
**Adjust Padding of Feedback Survey TopBar**
Quick fix to ensure the correct padding is applied to items in the top callout banner (checklist/invites/feedback)

**Adjust cookie expiry for hasCompletedChecklistV1**
Although the checklist dismissed cookie has an expiry of 2 years, hasCompletedChecklistV1 expires on session, meaning it keeps coming back, even when you have completed. I updated hasCompletedChecklistV1 to be 2 years. 

**Update "Create new model" dialog with labels and Capitalise letters**
Updated title. Added Labels.

**Adjust Version options dropdown**
Add Icons. Reorder based on Designs. Adjust padding. Add colour option for menu items. This has been applied to the menus for versions and models. 

**Update Toggle Component**
New look based on designs. It's now a more obvious toggle, showing both list and card options. Clicking on either toggles the selected view. 

**Move Measure mode button**
Small changes. Simply moved the Measure Mode button further down the Viewer options menu. 

**Remove disabled state from Version button**
This button showed the disabled state, but still allowed a click. This takes you to a page with no versions, but there is a CTA to add a version from here, it was decided to just remove the disabled state and allow the user to go to this page, even if they had 0 versions, so they understood better what they were seeing.

**Add Description to Create New Model and Edit Model Dialogs**
A lot of the feedback from SOM was around how they used Model names as descriptors, making it easier for them to keep track of models/versions. FE2 steps away from this for various reasons, so a compromise was bring back the "description" field. Here, users can add the human readable information on models, making keeping track of versions a lot easier. The backend already had description, but we had nowhere in the UI to set or update it. I updated the "Rename Model" dialog to become "Edit Model" and added the description field. The description field was also added to the "Create New Model" dialog. 

## Screenshots:
image
image
image
image

## Checklist:


packages/frontend-2/components/project/page/models/NewModelStructureItem.vue
andrewwallacespeckle, line 1: I'm not too sure what this page is? I had ts error which I fixed, but not sure where to test or if this is used?

fabis94: @andrewwallacespeckle I think this is the new model entry row that appears in the models list/structured view once you click on "+ New" under a specific list item

fabis94: @andrewwallacespeckle There's no description field here tho (and we might not even want one there), so you're kinda misleading devs by saying that the `useForm` submit handler will also give you a `description` value.

You should remove that bit and instead just make description null or something

packages/frontend-2/components/project/model-page/Versions.vue
```diff
@@ -119,6 +120,10 @@ graphql(`
   fragment ProjectModelPageVersionsProject on Project {
     id
     role
+    name
```
fabis94, line 123: Instead of duplicating the fields that ProjectPageHeader needs here, you should use ProjectPageHeader's `ProjectPageProjectHeader` fragment instead

packages/server/modules/core/services/branch/management.ts
```diff
@@ -45,7 +45,11 @@ export async function createBranchAndNotify(
 
   const branch = await createBranch({
     name: input.name,
-    description: isBranchCreateInput(input) ? input.description || null : null,
+    description: isBranchCreateInput(input)
```
fabis94, line 48: I think since both types have the description field you can just shorten all of this to:

```
description: input.description ?? null
```

fabis94 (changes requested):
left a few comments, looks good overall

fabis94 (approved):
lgtm

PR #1907, opened 2023-12-12
FE2 - User Feedback Request -  Menu item and Callout Banner

andrewwallacespeckle:
## Description & motivation
Give users the ability to submit feedback. This is currently a link (target blank) to a Google Form, but can be updated if this evolves.

## Changes:
**New Menu Item**
New menu style based on Agi's Designs. 

**New Menu Item**
A new menu item has been added to the main "User Menu" called "Feedback". This opens the Google form in a new tab. 

**Feedback Callout Banner**
A new callout has been added to the banner area that sometimes appears under the header bar. Here, we request the user for feedback, with the same link as above, but also a skip button. 
This is managed by cookies. We store the first time the user visits FE2 as a date, and use this date to decide if it enough time has passed to request feedback.  

**Instructional Animation Component**
A new component in ui-components. This allows you to quickly put together example layouts in tailwind, using the background slot. You then can simulate various use interactions with the mouse, including moving and clicking. There are 5 empty slots for use in triggering extra content in the animation.

## Screenshots:
image
image
image


packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -256,12 +265,26 @@ watch(search, (newVal) => {
 
 watch(areQueriesLoading, (newVal) => (showLoadingBar.value = newVal))
 
-const twoYearsFromNow = new Date()
-twoYearsFromNow.setFullYear(twoYearsFromNow.getFullYear() + 2)
+function getFutureDateByDays(daysToAdd: number) {
+  const futureDate = new Date()
+  futureDate.setDate(futureDate.getDate() + daysToAdd)
```
fabis94, line 270: What if getDate() returns 31? If you add to it, is it gonna roll over back to 1?

Generally I advise using dayjs() for any date manipulation, cause it takes care of all kinds of details like this. The base JS Date() object and its related functions are generally way too low level to easily work with.

packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -296,6 +324,29 @@ const showChecklist = computed(() => {
   return false
 })
 
+const showFeedbackRequest = computed(() => {
+  let storedDateString = onboardingOrFeedbackDate.value
+
+  const currentDate = new Date()
+
+  if (!storedDateString) {
+    const formattedDate = currentDate.toISOString().split('T')[0]
```
fabis94, line 333: Why do this? These manual string manipulations are brittle and you can just store the full timestamp and then easily read it any way you like with dayjs()

packages/frontend-2/components/projects/FeedbackRequest/Banner.vue
```diff
@@ -0,0 +1,46 @@
...
+          external
+          target="_blank"
+          @click="emitDismissedOrOpened"
+        >
+          Open Survey
+        </FormButton>
+      </div>
+    </div>
+    <ProjectsFeedbackRequestSkipDialog
+      :open="showSkipDialog"
```
fabis94, line 29: What if the dialog is closed by the user pressing Esc or clicking outside of it? That is - what if the value of `open` is changed from inside the SkipDialog component and emited out as an event? You've only got one way binding here (no v-model:open and thus not listening for updates to that value), so you're likely to get your local `showSkipDialog` value out of sync from the actual `open` state

packages/frontend-2/components/projects/FeedbackRequest/SkipDialog.vue
```diff
@@ -0,0 +1,117 @@
...
+    v-model:open="isOpen"
+    max-width="sm"
+    :buttons="dialogButtons"
+    prevent-close-on-click-outside
+    title="Feedback Option"
+    hide-closer
+  >
+    <CommonAnimationInstructional
+      :initial-position="{ top: 56, left: 71 }"
+      :actions="[
```
fabis94, line 12: Too much logic for the template IMO, move this to the script section

packages/ui-components/src/components/common/animation/Instructional.vue
```diff
@@ -0,0 +1,140 @@
...
+    >
+      <ClickIcon
+        class="absolute -top-5 -left-4 h-12 w-12 -rotate-12 text-foreground"
+        :class="[{ hidden: !isClicked }]"
+      />
+      <MouseIcon class="absolute inset-0 h-11 w-11" />
+    </div>
+    <div class="w-full h-full overflow-hidden">
+      <slot name="background"></slot>
+      <template v-if="slotVisibility.slot1">
```
fabis94, line 22: 5 slots seems like an arbitrary amount, remember that slots can be dynamically defined (like you once did for the Table component) and we can use that to our advantage here. Or maybe have 1 slot with the slot index being passed through slot props. And the dev can ask for how many slots he likes through a prop, or something

packages/ui-components/src/components/common/animation/Instructional.vue
```diff
@@ -0,0 +1,140 @@
...
+      <template v-if="slotVisibility.slot5">
+        <slot name="slot5"></slot>
+      </template>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref, onMounted, defineProps, type PropType, onBeforeUnmount } from 'vue'
+import MouseIcon from './MouseIcon.vue'
```
fabis94, line 43: Switch both of these to absolute path imports

fabis94 (changes requested):
left a few comments

packages/frontend-2/components/projects/FeedbackRequest/Banner.vue
andrewwallacespeckle: I added prevent-close-on-click-outside and hide-closer to prevent this. Is this enough, or should I handle additional scenarios where the dialog might be closed unexpectedly?

andrewwallacespeckle:
This is ready for re-review @fabis94 

Thanks!

packages/ui-components/src/components/common/animation/Instructional.vue
```diff
@@ -0,0 +1,125 @@
...
+})
+
+const isAnimating = ref(true)
+const mousePosition = ref({ ...props.initialPosition })
+const isClicked = ref(false)
+const animationDuration = ref(500)
+const isMouseVisible = ref(true)
+const dynamicSlots = ref(props.slotsConfig || [])
+
+function delay(action: DelayAction) {
```
fabis94, line 74: You can use `wait` from `@speckle/shared` for this

```ts
async function delay(action: DelayAction) {
  await wait(action.duration)
}
```

packages/ui-components/src/components/common/animation/Instructional.vue
```diff
@@ -0,0 +1,125 @@
...
+}
+
+function toggleSlotVisibility(action: SlotAction) {
+  const slotToToggle = dynamicSlots.value.find((slot) => slot.name === action.slot)
+  if (slotToToggle) {
+    slotToToggle.visible = !slotToToggle.visible
+  }
+}
+
+async function handleAction(action: Action) {
```
fabis94, line 89: Doesn't look like this has to be async

packages/ui-components/src/components/common/animation/Instructional.vue
```diff
@@ -0,0 +1,125 @@
...
+      return delay(action)
+    case 'slot':
+      toggleSlotVisibility(action as SlotAction)
+      break
+  }
+}
+
+onMounted(() => {
+  const loopActions = async () => {
+    while (isAnimating.value) {
```
fabis94, line 109: Does this need throttling/delays maybe? I'm concerned about the performance of this as it will run indefinitely and potentially run through a lot of CPU cycles.

fabis94 (changes requested):
Left a few minor comments, looks good overall

packages/ui-components/src/components/common/animation/Instructional.vue
andrewwallacespeckle: I have changed it so there is now a delay before and after each animation run

fabis94 (approved):

PR #1960, opened 2024-01-11
fix(fileimport-service): logging was not writing valid log messages directly to std(out|err)

iainsproat:
## Description & motivation

The file import service was wrapping valid log messages from the parser into another pino log message. It should have written the parser's log message directly to stdout/stderr.

This resulted in the following log message as viewed in seq:

Screenshot 2024-01-11 at 12 54 04

## Checklist:

### DIFF
--- a/packages/fileimport-service/src/daemon.js
+++ b/packages/fileimport-service/src/daemon.js
@@ -244,7 +244,7 @@ function runProcessWithTimeout(processLogger, cmd, cmdArgs, extraEnv, timeoutMs)
         JSON.parse(data) // validate that data is already in JSON format
         process.stdout.write(data)
       } catch {
-        boundLogger.info('Parser: %s', JSON.stringify(data))
+        boundLogger.info({ parserLog: data }, 'ParserLog: {parserLog}')
       }
     })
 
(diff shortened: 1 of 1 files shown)


packages/fileimport-service/src/daemon.js
```diff
@@ -240,19 +240,21 @@ function runProcessWithTimeout(processLogger, cmd, cmdArgs, extraEnv, timeoutMs)
...
     childProc.stderr.on('data', (data) => {
       try {
-        JSON.parse(data.toString()) // data is already in JSON format
-        process.stderr.write(data.string())
+        if (typeof data !== 'string') throw new Error('not a string')
+        JSON.parse(data) // validate that data is already in JSON format
+        process.stderr.write(data)
       } catch {
-        boundLogger.info('Parser: %s', data.toString())
+        boundLogger.info('Parser: %s', JSON.stringify(data))
```
gjedlicska, line 257: these should be seq message templates and context objects

gjedlicska (changes requested):

gjedlicska:
We prob need to split the data by line and re-log it line by line. The data line should be a clef message json stringified. So we can attach the parser type as a prop and parse out the clef message from the line

PR #1972, opened 2024-01-15
fix(server): locking to avoid postgres notification listeners processing the same message multiple times

fabis94:


packages/server/modules/core/utils/dbNotificationListener.ts
```diff
@@ -1,33 +1,78 @@
...
+      'Unexpected failure! Attempting to get task lock before redis client is initialized'
+    )
+  }
+
+  const lockKey = `${lockName}:${taskId}`
+  const lock = await redisClient.set(lockKey, '1', 'EX', 60 * 60, 'NX')
+  const releaseLock = async () => {
+    if (!redisClient) {
+      throw new LogicError(
+        'Unexpected failure! Attempting to get task lock before redis client is initialized'
```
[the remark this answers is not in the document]
fabis94: yep, fixed

packages/server/modules/core/utils/dbNotificationListener.ts
```diff
@@ -1,33 +1,78 @@
...
+  const unlock = await getTaskLock(messageId)
+  if (unlock) {
+    dbNotificationLogger.info(
+      logPayload,
+      'Message #{messageId} of channel {channel} starting processing...'
+    )
+    await Promise.resolve(listener.listener(msg))
+    await unlock()
+  } else {
+    dbNotificationLogger.info(
```
gjedlicska, line 71: Lets lower this to debug.

packages/server/modules/core/utils/dbNotificationListener.ts
```diff
@@ -1,33 +1,78 @@
...
-    },
-    'Message received'
+  const messageId = getMessageId(msg)
+
+  const logPayload = {
+    ...msg,
+    listenerRegistered: !!listener,
+    messageId
+  }
+  dbNotificationLogger.debug(
```
gjedlicska, line 55: This is redundant, the next if else block log lines should be sufficient.

packages/server/modules/core/utils/dbNotificationListener.ts
```diff
@@ -1,33 +1,78 @@
...
+    logPayload,
+    'Message #{messageId} received in channel {channel}'
   )
   if (!listener) return
 
-  return listener.listener(msg)
+  // Only process if lock acquired
+  const unlock = await getTaskLock(messageId)
+  if (unlock) {
+    dbNotificationLogger.info(
```
gjedlicska, line 64: Lets lower this to debug.

gjedlicska (changes requested):

packages/server/modules/core/utils/dbNotificationListener.ts
fabis94: do we really want all of those to be debug level? if we're capturing requests w/ INFO, I would think that we would want the same for incoming DB events, at least the ones that are actually processed, not skipped

gjedlicska (approved):

PR #1976, opened 2024-01-19
feat(server): new automation result reporting scope

fabis94:
Based on my `fabians/token-resource-limits` branch


gjedlicska (approved):

gjedlicska:
Nice, we're going to need to sign ppl out of automate, otherwise automations are going to break.
Cause we're using the already existing speckle app token, which doesn't have this scope, to create a new project scoped token.

PR #1979, opened 2024-01-22
FE2 - Embedding

andrewwallacespeckle:
Adding embed functionality to FE2. This is done through the ProjectModelPageDialogEmbed component. 

This can be accessed from:
- Project Page - Card Actions
- Version Page - Card Actions
- Viewer - Share Menu in Header

The dialog has an iframe preview of the embed, depending on options selected. 

These options are:
- Transparent background
- Hide viewer controls
- Hide the Selection Info panel
- Prevent scrolling (zooming)
 -Load model manually
- Comment slideshow mode

image


andrewwallacespeckle:
@fabis94 From our call, a reminder of an issue that currently exists on this branch. This now builds locally fine (but with warnings), but it still fails on circleci at the Link step:
https://app.circleci.com/pipelines/github/specklesystems/speckle-server/6927/workflows/f31f5b96-3dd2-4018-b36b-8b26a9dd1b35/jobs/72675

There are a lot of different failures on circleci, but i struggled to find the root cause.

The issue we encountered with pre-commit hooks has now been fixed.

andrewwallacespeckle:
> @fabis94 From our call, a reminder of an issue that currently exists on this branch. This now builds locally fine (but with warnings), but it still fails on circleci at the Link step: https://app.circleci.com/pipelines/github/specklesystems/speckle-server/6927/workflows/f31f5b96-3dd2-4018-b36b-8b26a9dd1b35/jobs/72675
> 
> There are a lot of different failures on circleci, but i struggled to find the root cause.
> 
> The issue we encountered with pre-commit hooks has now been fixed.

Resolved

andrewwallacespeckle:
@fabis94 Outstanding issues on this task:
PreSetupWrapper - When clicking an object and opening selection info, it will use a portal for mobile view. When the selection info is moved to this portal, the ui is there, but the model info is not. There are also console errors about undefined content. 

Another I forgot to mention - Comment Slideshow mode - When the user checks comment slideshow option, it should go through comments in chronological order. Am I right in thinking the best way to achieve this is to check for the newest comment, and append it to the url with focusedThreadID? Is there a way of getting the latest comment?


Other outstanding issues I will look at today:
- Base URL Env - confirm with Iain & Gergo
- Preview URL Builder
- Embed Loading State
- Turn off Selection info on click of comment on mobile
- Build iframe src with url method

fabis94:
@andrewwallacespeckle Regarding comment slideshow mode - `state.ui.threads.items` contains all of the threads already sorted in correct order, you just need to walk up or down the array depending on which arrow was pressed and set `focusedThreadId` to the correct thread ID value (you can use `useThreadUtilities().open()` for this

packages/frontend-2/components/project/page/models/Actions.vue
```diff
@@ -23,11 +23,20 @@
       :project-id="projectId"
       @deleted="$emit('model-updated')"
     />
+    <ProjectModelPageDialogEmbed
+      v-model:open="embedDialogOpen"
+      :project-id="projectId"
+      :model-id="model.id"
+      :visibility="props.visibility"
```
fabis94, line 30: props prefix redundant in Vue templates (check everywhere)

packages/frontend-2/components/project/page/latest-items/comments/IntroCard.vue
```diff
@@ -61,4 +77,7 @@ withDefaults(
 )
 
 const { isDarkTheme } = useTheme()
+const { embedOptions } = useEmbedState()
+
+const router = useRouter()
```
fabis94, line 82: `useRoute()`

fabis94 (changes requested):
looks good for the most part, mostly various minor issues

packages/ui-components/src/components/form/ClipboardInput.vue
```diff
@@ -48,10 +51,29 @@ const emit = defineEmits<{ (e: 'copy', val: string): void }>()
...
 
+const copied = ref(false)
+
 const handleCopy = async () => {
   if (props.value) {
     await copy(props.value)
+    copied.value = true
     emit('copy', props.value)
+
+    setTimeout(() => {
```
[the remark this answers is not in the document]
andrewwallacespeckle: This updates the icon to the clipboard-document-check from clipboard-document instantly on click. It reverts back to original after 2 seconds

packages/frontend-2/pages/projects/[id]/models/[modelId]/index.vue
```diff
@@ -1,80 +1,28 @@
...
-      </ClientOnly>
-    </div>
-  </ViewerPostSetupWrapper>
-  <div
-    v-if="tourState.showViewerControls"
-    class="sm:hidden shadow-t fixed bottom-0 left-0 max-h-[65vh] overflow-hidden w-screen z-50 transition-all duration-300 empty:-bottom-[65vh]"
-  >
-    <PortalTarget name="bottomPanel"></PortalTarget>
-    <PortalTarget name="mobileComments"></PortalTarget>
+  <div>
```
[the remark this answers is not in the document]
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/projects/[id]/models/[modelId]/index.vue:2]: I thought I was allowed to do this in Vue 3. I added the eslint rule back in to temporarily check this, and I have fixed it in any file that I have modified. I have removed the rule again as there was some other files it was flagging.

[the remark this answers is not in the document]
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/projects/[id]/models/[modelId]/index.vue:2]: It didn't fix this error however, I've had not luck on finding a solution yet

packages/frontend-2/lib/viewer/composables/setup/core.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Keeping some core Viewer state code here so that we can import it without
+ * importing the entire Viewer state related codebase. Useful in embed mode where
+ * we don't want to load all of the Viewer JS before the Play button is pressed.
+ */
+
+/**
+ * Vue injection key for the Injectable Viewer State
+ */
+export const InjectableViewerStateKey: InjectionKey<InjectableViewerState> = Symbol(
```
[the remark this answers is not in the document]
andrewwallacespeckle: Thanks for the explanation

packages/frontend-2/components/viewer/selection/Sidebar.vue
```diff
@@ -73,17 +75,24 @@ import { containsAll } from '~~/lib/common/helpers/utils'
...
     metadata: { filteringState }
   },
+  urlHashState: { focusedThreadId },
   ui: { diff }
 } = useInjectedViewerState()
 const { objects, clearSelection } = useSelectionUtilities()
 const { hideObjects, showObjects, isolateObjects, unIsolateObjects } =
   useFilterUtilities()
 
+const { embedOptions } = useEmbedState()
```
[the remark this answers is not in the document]
andrewwallacespeckle: I have removed it from Sidebar, and anywhere else Viewer was available. 

It's still in the following, Please confirm the below are need:
- Footer.vue (Used in manual mode with no viewer)
- PresetupWrapper.vue - Pre Viewer Setup
- SignIn.vue - Pre Viewer Setup
- viewer.vue - layout file

andrewwallacespeckle:
@fabis94 I've also noticed this in the console, without much success in finding what is causing it:

image

packages/frontend-2/pages/projects/[id]/models/[modelId]/index.vue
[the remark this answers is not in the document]
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/projects/[id]/models/[modelId]/index.vue:2]: @fabis94 I still can't find anything on this, can you have a look?

packages/frontend-2/components/project/model-page/dialog/embed/Embed.vue
```diff
@@ -0,0 +1,248 @@
...
+const hideSelectionInfo = ref(false)
+const preventScrolling = ref(false)
+const manuallyLoadModel = ref(false)
+const commentSlideshowMode = ref(false)
+
+const updateOption = (optionRef: Ref<boolean>, newValue: unknown) => {
+  optionRef.value = newValue === undefined ? false : !!newValue
+}
+
+const embedDialogOptions = [
```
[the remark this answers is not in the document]
andrewwallacespeckle: I made this reactive, but it no longer updated the url. My approach was to keep the array as a plain structure, relying on the reactivity of the individual ref properties within the array's objects to trigger updates. Is this enough here or should I look for another solution?

packages/frontend-2/components/header/LogoBlock.vue
```diff
@@ -1,28 +1,29 @@
...
-  },
-  active: {
-    type: Boolean,
-    default: true
-  }
-})
+defineProps<{
+  minimal?: boolean
+  active?: boolean
+  to?: string
```
fabis94, line 25: Shouldn't it default to `/` so that you don't need to specify it everywhere? You can use vue's `withDefaults()` for this

packages/frontend-2/components/header/NavShare.vue
```diff
@@ -0,0 +1,124 @@
...
+  if (SpeckleViewer.ViewerRoute.isModelResource(firstResource.value)) {
+    return firstResource.value.modelId
+  }
+  return ''
+})
+
+const isFederated = computed(() => parsedResourceIds.value.length > 1)
+
+const handleCopyId = () => {
+  copy(modelId.value, { successMessage: 'Model ID copied to clipboard' })
```
fabis94, line 114: I'm not sure about the value of such a function. Surely the user would be interested in the version ID as well?

packages/frontend-2/components/header/NavShare.vue
```diff
@@ -0,0 +1,124 @@
...
+}
+
+const handleCopyLink = () => {
+  copyModelLink(props.projectId, props.resourceIdString)
```
fabis94, line 118: That's not how `copyModelLink` works, look at the parameters it expects. It expects a modelId & versionId separately.

packages/frontend-2/components/project/model-page/dialog/embed/Embed.vue
```diff
@@ -0,0 +1,263 @@
...
+const router = useRouter()
+const route = useRoute()
+const { copy } = useClipboard()
+const {
+  public: { baseUrl }
+} = useRuntimeConfig()
+
+const { isSmallerOrEqualSm } = useIsSmallerOrEqualThanBreakpoint()
+
+const projectVisibility = ref(ProjectVisibility)
```
fabis94, line 130: You don't need to do this. As long as you have ProjectVisibility imported in the setup block, you can also use it in the template directly.

packages/frontend-2/components/project/model-page/dialog/embed/Embed.vue
```diff
@@ -0,0 +1,263 @@
...
+
+const handleEmbedCodeCopy = async (value: string) => {
+  await copy(value, {
+    successMessage: 'Embed code copied to clipboard',
+    failureMessage: 'Failed to copy embed code to clipboard'
+  })
+}
+
+const updateOption = (optionRef: Ref<boolean>, newValue: unknown) => {
+  optionRef.value = newValue === undefined ? false : !!newValue
```
fabis94, line 233:
```suggestion
  optionRef.value = !!newValue
```

packages/frontend-2/components/project/page/latest-items/comments/IntroCard.vue
```diff
@@ -30,22 +30,38 @@
...
         Head over to a model and start coordinating right away!
       </div>
-      <div v-else class="mt-2">
-        <FormButton :icon-left="PlusIcon" @click="() => $emit('new-discussion')">
+      <div v-else class="mt-3">
+        <FormButton
+          v-if="isEnabled"
+          size="sm"
+          :icon-right="ArrowTopRightOnSquareIcon"
+          :to="route.path"
```
fabis94, line 43: Why?

packages/frontend-2/components/project/page/team/Dialog.vue
```diff
@@ -21,6 +25,8 @@ import type { ProjectPageTeamDialogFragment } from '~~/lib/common/generated/gql/
 import { graphql } from '~~/lib/common/generated/gql'
 import { useTeamDialogInternals } from '~~/lib/projects/composables/team'
 
+type OpenSectionType = 'invite' | 'access' | null
```
fabis94, line 28: Duplicate type (I saw the same one in Settings.vue), export this out from some kind of helper. Also, it might be easier to work with if you define it as an enum instead.

packages/frontend-2/components/viewer/GlobalFilterReset.vue
```diff
@@ -25,6 +30,10 @@ const {
   filters: { hasAnyFiltersApplied }
 } = useFilterUtilities()
 
+defineProps<{
+  embed?: boolean
```
fabis94, line 34: Why not just read this value from the embed state? `useEmbed()`

packages/frontend-2/components/viewer/PreSetupWrapper.vue
```diff
@@ -0,0 +1,153 @@
...
+  setup: [InjectableViewerState]
+}>()
+
+const route = useRoute()
+const { showTour, showControls } = useViewerTour()
+
+const projectId = computed(() => route.params.id as string)
+const modelId = computed(() => route.params.modelId as string)
+const versionId = computed(() => {
+  const parts = modelId.value.split('@')
```
fabis94, line 106: 1. Use resourceIdString builder utils for this
2. There could easily be multiple versionIds in the url (multiple models). How is this supposed to work?

packages/frontend-2/components/viewer/anchored-point/Thread.vue
```diff
@@ -147,10 +144,29 @@
               </div>
             </div>
             <ViewerAnchoredPointThreadNewReply
-              v-if="!modelValue.archived && canReply"
+              v-if="
+                !modelValue.archived &&
```
fabis94, line 148: Getting a bit too long/complicated for the template. Probably a good idea to move this to a computed.

packages/frontend-2/lib/viewer/composables/activity.ts
```diff
@@ -174,7 +176,7 @@ export function useViewerUserActivityTracking(params: {
     const incomingSessionId = event.sessionId
 
     if (sessionId.value === incomingSessionId) return
-    if (status === ViewerUserActivityStatus.Disconnected) {
+    if (!isEmbedEnabled && status === ViewerUserActivityStatus.Disconnected) {
```
fabis94, line 179: This feels very inefficient - why not just disable the subscription altogether if we're in embed mode? Instead of still receiving and processing all updates, just doing nothing if `isEmbedEnabled`

packages/ui-components/src/components/layout/DialogSection.vue
```diff
@@ -109,7 +126,10 @@ const props = defineProps({
         onClick?: () => void
       }
     | undefined,
-  alwaysOpen: Boolean
+  alwaysOpen: Boolean,
+  guidedOpen: Boolean,
+  lazyLoad: Boolean,
```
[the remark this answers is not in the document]
fabis94: @andrewwallacespeckle This conversation was resolved without any explanation and/or changes

fabis94 (changes requested):
haven't tested it, but code looks fine for the most part. some minor issues remain

packages/frontend-2/components/viewer/PreSetupWrapper.vue
andrewwallacespeckle: This was an unused prop, so i removed

andrewwallacespeckle:
@fabis94 These should all be done now. 

There is 1 remaining issue that I'm struggling with. It is around the Dialog Section changes. I reverted the lazy load changes, and used always-open instead of guided-open. `?settings=true` works, as does `?settings=access`. The problem comes when you close the dialog, it reopens as an unclickable dialog. I'm not sure whats happening. I added a watch to remove the settings parameter on close, but it still has the problem. Could you take a look?

One last thing is there is a merge conflict in a file that you worked on, could you resolve the conflict? I'm not sure what is correct.

packages/frontend-2/components/header/LogoBlock.vue
```diff
@@ -1,28 +1,34 @@
...
+withDefaults(
+  defineProps<{
+    minimal?: boolean
+    active?: boolean
+    to?: string
+    showTextOnMobile?: boolean
+    target?: string
+  }>(),
+  {
+    to: '/'
```
fabis94, line 31: `active` previously defaulted to true

packages/frontend-2/components/header/NavShare.vue
```diff
@@ -0,0 +1,133 @@
...
+  if (SpeckleViewer.ViewerRoute.isModelResource(firstResource.value)) {
+    return firstResource.value.modelId // Assuming your firstResource object has a modelId property
+  }
+  return ''
+})
+
+const isFederated = computed(() => parsedResourceIds.value.length > 1)
+
+const handleCopyId = () => {
+  copy(props.resourceIdString, { successMessage: 'ID copied to clipboard' })
```
fabis94, line 121: In practice this can be multiple IDs, if there are multiple models loaded in the viewer. I'm still not sure if this won't be confusing to the user, tho, cause the only place where we accept IDs with this format is in the viewer URLs. And maybe people get confused if there are multiple comma delimited IDs?

I dunno, I would probably discuss this feature with Benjamin more thoroughly to better make sure this is actually something valuable to our users

packages/frontend-2/components/project/page/stats-block/Settings.vue
```diff
@@ -24,14 +24,20 @@
       </div>
     </template>
     <template #default>
-      <ProjectPageTeamDialog v-model:open="dialogOpen" :project="project" />
+      <ProjectPageTeamDialog
+        v-model:open="dialogOpen"
+        :project="project"
+        :open-section="openSection || undefined"
```
fabis94, line 30: Why not just make the ref `ref<OpenSectionType | undefined>()` so that you don't need to do this? The introduction of `null` into the mix doesn't seem to bring any benefit

packages/frontend-2/lib/viewer/composables/activity.ts
```diff
@@ -81,7 +82,8 @@ export function useViewerUserActivityBroadcasting(
   const apollo = useApolloClient().client
 
   const invokeMutation = async (message: ViewerUserActivityMessageInput) => {
-    if (!isLoggedIn.value) return false
+    const { isEnabled: isEmbedEnabled } = useEmbed()
```
fabis94, line 85: Remember that composables can only be invoked 1) in other composables 2) in vue setup functions 3) in nuxt plugins/middlewares.

You're invoking the composable in an arbitrary function - `invokeMutation`. `invokeMutation` can be invoked at any point in time (its triggered manually), so there's no guarantee at all that the `useEmbed()` composable will work.

Move this line a few lines up so that it actually gets invoked at the correct time with the rest of the composables inside `useViewerUserActivityBroadcasting`

fabis94 (changes requested):

packages/frontend-2/components/header/NavShare.vue
andrewwallacespeckle: There is a v-if on the "Copy ID" button, where it won't be shown if there are federated models. handleCopyId can only be called when there is a single model.

andrewwallacespeckle: image

packages/frontend-2/components/project/page/stats-block/Settings.vue
```diff
@@ -61,6 +67,49 @@ const props = defineProps<{
...
+}
+
+onMounted(() => {
+  readDialogStateFromQuery()
+})
+
+watch(
+  () => route.query.settings,
+  () => {
+    readDialogStateFromQuery()
```
fabis94, line 112: Only invoke if new settings value differs from the old one and is set

fabis94 (dismissed):

fabis94 (approved):

PR #1982, opened 2024-01-22
Edit ProjectPageHeader from Versions page

andrewwallacespeckle:
Quick bugfix removing header from version page


packages/frontend-2/components/common/EditableTitleDescription.vue
```diff
@@ -0,0 +1,194 @@
...
+import { debounce } from 'lodash-es'
+
+const props = defineProps({
+  initialTitle: String,
+  initialDescription: String,
+  canEdit: Boolean,
+  isDisabled: Boolean
+})
+
+const emit = defineEmits(['update:title', 'update:description'])
```
fabis94, line 87: We don't have to use defineModel, but the prop & event names should at least be connected. Rename initialTitle and initialDescription to title & description respectively.

packages/frontend-2/components/common/EditableTitleDescription.vue
```diff
@@ -0,0 +1,194 @@
...
+  'normal placeholder:text-foreground-2 text-foreground-2 focus:text-foreground',
+  'border-0 border-b-2 focus:border-outline-3',
+  'p-0 bg-transparent border-transparent focus:outline-none focus:ring-0'
+])
+
+const onInputKeydown = (e: KeyboardEvent) => {
+  if (e.code === 'Enter' && e.target instanceof HTMLElement) {
+    e.preventDefault()
+    e.target.blur()
+    if (e.target.dataset.type === 'title' && lastTitleValue.value !== title.value) {
```
fabis94, line 116: How are these dataset values set up? I don't see any data- props defining these values in the template.

packages/frontend-2/components/common/EditableTitleDescription.vue
```diff
@@ -0,0 +1,194 @@
...
+])
+
+const descriptionInputClasses = computed(() => [
+  'normal placeholder:text-foreground-2 text-foreground-2 focus:text-foreground',
+  'border-0 border-b-2 focus:border-outline-3',
+  'p-0 bg-transparent border-transparent focus:outline-none focus:ring-0'
+])
+
+const onInputKeydown = (e: KeyboardEvent) => {
+  if (e.code === 'Enter' && e.target instanceof HTMLElement) {
```
fabis94, line 113: We only want Enter to submit for 'title'. 'description' allows multiple lines and thus should not submit on enter.

packages/frontend-2/components/project/model-page/Header.vue
```diff
@@ -33,5 +49,87 @@ const props = defineProps<{
...
+watch(
+  () => props.project.model.name,
+  () => {
+    resetTitle()
+  },
+  { immediate: true }
+)
+
+const save = async () => {
+  if (!anythingToUpdate.value || anyMutationsLoading.value) return
```
fabis94, line 104: I see a duplicate save request when changing the title in the model page, not in the project page tho. So I suggest putting a few breakpoints there to see why it gets triggered twice with the same value, even with all of these new duplication checks.

fabis94 (changes requested):
there's a few minor issues I found, see comments

packages/frontend-2/components/common/EditableTitleDescription.vue
andrewwallacespeckle: Not sure how they dropped off. Re-added.

andrewwallacespeckle:
Thanks. Should be all done.

fabis94 (changes requested):
it appears that in EditableTitleDescription you don't cancel the debounced emits onBlur (I mentioned how to do this in the CR, essentially - debouncedFn.cancel())
so when a user clicks outside from the input field, onBlur triggers and emits an update
and then after the debounce period, the debounced fn also triggers and emits an update
on top of that, you only make the check if last value !== new value in onBlur
while you should be doing that as close to the final emit as possible, to ensure that its always checked
whether onBlur or when the debounced fn triggers
so probably in emitTitle and emitDescription
on top of that, in ModelPageHeader's save(), you only resetInputs() at the very end, so if save() gets triggered twice, the 2nd invocation still thinks theres an update to be made, even if the first invocation is already processing and has sent out the changes
but if you fix the first issue, then this will probably become irrelevant

fabis94 (changes requested):
found bug (reported in dms)

fabis94 (approved):
Seems to work now

PR #1994, opened 2024-01-26
Remove `findAll` when matching ids

AlexandruPopovici:
## Description & motivation
In FE2 when clicking on any of the items highlighted in the image below

finaAll1

Which would make all the objects of the selected type to become 'selected', for a large number of objects the main thread would get blocked for tens of seconds. 

60% of the execution time was taken up by using `findAll` when going over very many string ids, to get the tree node from the viewer. In API 2.0, this is no longer needed. Instead you use `findId` which is incomparably faster since it's just one or two lookups. After making this change, the viewer overhead was completely gone. However there is still some FE2 related overhead, accounting for approx 30% of the total inital overhead. 

I'm leaving two profiler recordings, one before the change, one after the change. The FE2 related overhead is visible in both, and it seems to be the same one mentioned [here](https://www.notion.so/speckle/Update-to-the-Viewer-API-2-0-76d56381e6c448fbbeea1d8e88f00ac2?pvs=4#ad9f9a6da5c24681b54050591bb8613d)

## Changes:
- No more using `findAll` when trying to match object ids. Replaced with `findId`

## Screenshots:
**Before:**

**After:**

## Checklist:

### DIFF
--- a/packages/frontend-2/lib/viewer/composables/ui.ts
+++ b/packages/frontend-2/lib/viewer/composables/ui.ts
@@ -228,20 +228,15 @@ export function useSelectionUtilities() {
   } = useInjectedViewer()
 
   const setSelectionFromObjectIds = (objectIds: string[]) => {
-    const res = worldTree.value
-      ? worldTree.value.findAll((node: TreeNode) => {
-          const t = node.model as Record<string, unknown>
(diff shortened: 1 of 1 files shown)


fabis94 (approved):
I guess so? haha

didimitrie (approved):

PR #2009, opened 2024-02-02
[WBX-31] Make account modal less annoying in the viewer

andrewwallacespeckle:
## Description & motivation
[Link to JIRA Ticket
](https://spockle.atlassian.net/jira/software/c/projects/WBX/issues/WBX-31?filter=myopenissues)

## Changes:
- Change settings to Team
- Speckle Logo to Marketing Site
- Add Login button to Header on Desktop
- Add “Reply” Button below conversations
- Launch Login Modal on click of “+” to start new conversation

### DIFF
--- a/packages/frontend-2/components/header/LogoBlock.vue
+++ b/packages/frontend-2/components/header/LogoBlock.vue
@@ -1,28 +1,34 @@
 <template>
-  <NuxtLink class="flex items-center shrink-0" to="/">
+  <NuxtLink class="flex items-center shrink-0" :to="to" :target="target">
     <img
-      class="block h-6 w-6"
-      :class="{ 'mr-2': !minimal, grayscale: active }"
+      class="h-8 w-8 block"
+      :class="{
+        grayscale: active
+      }"
       src="~~/assets/images/speckle_logo_big.png"
       alt="Speckle"
     />
+
     <div
       v-if="!minimal"
-      class="text-primary h6 mt-0 hidden font-bold leading-7 md:flex"
+      class="text-primary text-base mt-0 font-bold leading-7"
+      :class="showTextOnMobile ? '' : 'hidden md:flex'"
     >
       Speckle
     </div>
   </NuxtLink>
 </template>
 <script setup lang="ts">
-defineProps({
-  minimal: {
-    type: Boolean,
-    default: false
-  },
-  active: {
-    type: Boolean,
-    default: true
+withDefaults(
+  defineProps<{
+    minimal?: boolean
+    active?: boolean
+    to?: string
+    showTextOnMobile?: boolean
+    target?: string
+  }>(),
+  {
+    to: '/'
   }
-})
+)
 </script>
--- a/packages/frontend-2/components/header/NavBar.vue
+++ b/packages/frontend-2/components/header/NavBar.vue
@@ -5,8 +5,8 @@
         <div class="flex items-center truncate">
           <HeaderLogoBlock
             :active="false"
-            :to="isLoggedIn ? '/' : 'https://speckle.systems/'"
-            :target="isLoggedIn ? undefined : '_blank'"
+            :to="activeUser ? '/' : 'https://speckle.systems/'"
+            :target="activeUser ? undefined : '_blank'"
           />
           <HeaderNavLink
             to="/"
@@ -32,5 +32,5 @@
 </template>
 <script setup lang="ts">
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
-const { activeUser, isLoggedIn } = useActiveUser()
+const { activeUser } = useActiveUser()
 </script>
--- a/packages/frontend-2/components/header/NavBar.vue
+++ b/packages/frontend-2/components/header/NavBar.vue
@@ -3,7 +3,11 @@
     <nav class="fixed z-20 top-0 h-14 bg-foundation shadow hover:shadow-md transition">
       <div class="flex items-center justify-between h-full w-screen px-4">
         <div class="flex items-center truncate">
-          <HeaderLogoBlock :active="false" />
+          <HeaderLogoBlock
+            :active="false"
+            :to="isLoggedIn ? '/' : 'https://speckle.systems/'"
+            :target="isLoggedIn ? undefined : '_blank'"
+          />
           <HeaderNavLink
             to="/"
             name="Dashboard"
@@ -28,5 +32,5 @@
 </template>
 <script setup lang="ts">
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
-const { activeUser } = useActiveUser()
+const { activeUser, isLoggedIn } = useActiveUser()
 </script>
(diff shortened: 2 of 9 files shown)


packages/frontend-2/components/header/NavBar.vue
```diff
@@ -3,7 +3,11 @@
     <nav class="fixed z-20 top-0 h-14 bg-foundation shadow hover:shadow-md transition">
       <div class="flex items-center justify-between h-full w-screen px-4">
         <div class="flex items-center truncate">
-          <HeaderLogoBlock :active="false" />
+          <HeaderLogoBlock
+            :active="false"
+            :to="activeUser ? '/' : 'https://speckle.systems/'"
+            :target="activeUser ? undefined : '_blank'"
```
benjaminvo, line 9: I think we can just open in the same window logged out too

packages/frontend-2/components/viewer/anchored-point/NewThread.vue
```diff
@@ -69,6 +69,21 @@
...
+          <FormButton size="sm" @click="showLoginDialog = true">
+            Sign in to Speckle
+          </FormButton>
+          <AuthLoginPanel
+            v-model:open="showLoginDialog"
+            dialog-mode
+            max-width="sm"
+            subtitle="Create a free account to keep using Speckle!"
+          />
+        </div>
```
benjaminvo, line 86: I think this one is unnecessary. Let's just open the sign in dialog as soon as you click the new thread button above.

Screenshot 2024-02-02 at 15 51 46

packages/frontend-2/components/viewer/anchored-point/Thread.vue
```diff
@@ -151,6 +151,21 @@
...
+              <FormButton size="sm" @click="showLoginDialog = true">
+                Sign in to Speckle
+              </FormButton>
+              <AuthLoginPanel
+                v-model:open="showLoginDialog"
+                dialog-mode
+                max-width="sm"
+                subtitle="Create a free account to keep using Speckle!"
+              />
+            </div>
```
benjaminvo, line 168: I like your "Join the conversation" copy but I suggest we just have a blue "Reply" button and clicking that opens the sign in dialog.

Screenshot 2024-02-02 at 23 11 33

benjaminvo (changes requested):

packages/frontend-2/components/viewer/anchored-point/Thread.vue
andrewwallacespeckle: image

benjaminvo: @andrewwallacespeckle Can you update it to what's in my image instead? No "Join the conversation" text and just a big Reply button (full-width button and a bit taller than what's in your image). Now that I'm thinking about it we could then use the "Join the conversation" copy as the subtitle in the auth modal.

andrewwallacespeckle: Ah sorry, I misread your comment. Will update now

andrewwallacespeckle: image

andrewwallacespeckle: I've made this change and pushed. There is a bit more work in modifying the Dialog to include "join the conversation", but it can be done in about 10 mins. Please confirm if you want me to do this

packages/frontend-2/components/header/LogoBlock.vue
```diff
@@ -1,28 +1,34 @@
...
-    type: Boolean,
-    default: false
-  },
-  active: {
-    type: Boolean,
-    default: true
+withDefaults(
+  defineProps<{
+    minimal?: boolean
+    active?: boolean
```
fabis94, line 25: `active` defaulted to `true` before

packages/frontend-2/components/header/NavBar.vue
```diff
@@ -3,7 +3,10 @@
     <nav class="fixed z-20 top-0 h-14 bg-foundation shadow hover:shadow-md transition">
       <div class="flex items-center justify-between h-full w-screen px-4">
         <div class="flex items-center truncate">
-          <HeaderLogoBlock :active="false" />
+          <HeaderLogoBlock
+            :active="false"
+            :to="activeUser ? '/' : 'https://speckle.systems/'"
```
fabis94, line 8: Wouldn't the 2nd URL be dependant on the environment? Do we always want to send guests to speckle.systems even if they're on a completely different deployment like latest.speckle.systems or even an academy one?

packages/frontend-2/components/header/NavUserMenu.vue
```diff
@@ -143,23 +143,24 @@ import {
...
-import type { Optional } from '@speckle/shared'
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
 import { useAuthManager } from '~~/lib/auth/composables/auth'
-import { loginRoute } from '~~/lib/common/helpers/route'
 import { useTheme, AppTheme } from '~~/lib/core/composables/theme'
 import { useServerInfo } from '~/lib/core/composables/server'
+import type { RouteLocation } from 'vue-router'
+
+defineProps<{
+  loginUrl?: RouteLocation & { href: string }
```
fabis94, line 153: What's the point requiring this prop to be so complicated if you only use it as a bool in the template? It could easily just be `shouldShowSignIn: boolean`

fabis94 (changes requested):

packages/frontend-2/components/header/NavUserMenu.vue
andrewwallacespeckle: It's also used as the 'to' prop in the NuxtLink

`            <NuxtLink
              :class="[
                active ? 'bg-foundation-focus' : '',
                'flex gap-3.5 items-center px-3 py-2.5 text-sm text-primary cursor-pointer transition'
              ]"
              :to="loginUrl"
            >
              <ArrowRightOnRectangleIcon class="w-5 h-5" />
              Sign In
            </NuxtLink>`

packages/frontend-2/components/header/NavBar.vue
andrewwallacespeckle: Good point, @benjaminvo what are your thoughts?

packages/frontend-2/components/header/NavUserMenu.vue
fabis94: @andrewwallacespeckle Right, my bad. In that case, it still doesn't necessarily have to be RouteLocation it can be `RouteLocationRaw`, which also supports basic `strings` and thus is way less complicated to feed in

fabis94: image
image

packages/frontend-2/components/header/NavBar.vue
benjaminvo: Yeah good question. Yes it should be dependent on the environment

fabis94 (approved):

PR #2014, opened 2024-02-06
Headerbar Consistency - Post merge

andrewwallacespeckle:
After merging in Embed, Benjamin noticed we can potentially have 1 buttons in the headerbar. One is sized smaller, and looked like a bug. I have readjust the size of the Share button to sm to match the Sign in button. 

Also increased gap on items in headerbar on mobile. 

Also removed canPostComment condition in v-if on ViewerAnchoredPointNewThread to ensure user can still try to comment on embed, but will be asked to login.

### DIFF
--- a/packages/frontend-2/components/header/NavBar.vue
+++ b/packages/frontend-2/components/header/NavBar.vue
@@ -12,7 +12,7 @@
           />
           <PortalTarget name="navigation"></PortalTarget>
         </div>
-        <div class="flex items-center gap-1.5 sm:gap-2">
+        <div class="flex items-center gap-2.5 sm:gap-2">
           <PortalTarget name="secondary-actions"></PortalTarget>
           <PortalTarget name="primary-actions"></PortalTarget>
           <!-- Notifications dropdown -->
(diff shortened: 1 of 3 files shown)


fabis94 (approved):

packages/frontend-2/components/header/NavBar.vue
```diff
@@ -12,7 +12,7 @@
           />
           <PortalTarget name="navigation"></PortalTarget>
         </div>
-        <div class="flex items-center gap-1.5 sm:gap-2">
+        <div class="flex items-center gap-2.5 sm:gap-2">
```
benjaminvo, line 15: Let's increase to gap-4 instead of gap-2.5 to make it easier to tap with a fat finger

andrewwallacespeckle: Just bear in mind that there can be other items in the headerbar. I personally think we are losing too much space to in the headerbar here, which is why I kept it a lower figure, but happy to let you make the decision. 
image

benjaminvo: good point, I only checked logged out where we don't have a notification bell. Let's just go with what you have here then

benjaminvo (approved):

PR #2015, opened 2024-02-06
feat(fe2): /profile route for user profile edit [WBX-66]

fabis94:


packages/ui-components/package.json
```diff
@@ -38,7 +38,7 @@
     "vue": "^3.3.0"
   },
   "dependencies": {
-    "@headlessui/vue": "^1.7.13",
+    "@headlessui/vue": "^1.7.18",
```
fabis94, line 41: resolved dialog SSR hydration mismatch error by updating headlessui

andrewwallacespeckle (approved):

PR #2029, opened 2024-02-09
FE2 Hotfix - Viewer Comments Permission Issue

andrewwallacespeckle:
Currently, when a user who is logged in, but they lack permission to comment, they are given a login dialog. This is incorrect as they are already logged in.

I updated the condition to check for activeUser, rather than canReply. 

There was also a small bug with the header of the panel. Removed unnecessary class. 

Edit: Confirmed during standup - Users with "View" access can still comment.


packages/frontend-2/components/viewer/anchored-point/Thread.vue
```diff
@@ -163,7 +163,7 @@
               </FormButton>
             </div>
             <div
-              v-if="!canReply && !isEmbedEnabled"
+              v-if="!activeUser && !isEmbedEnabled"
```
fabis94, line 166: Instead of removing the canReply check altogether, you should update the `useCheckViewerCommentingAccess()` composable to check that either: the user has a project role OR the project has `allowPublicComments` turned on.

The way you've configured it now is not correct, because you're allowing every logged in user to reply, even if the project doesn't have allowPublicComments turned on and the user does not have a project role.

fabis94 (approved):

PR #2032, opened 2024-02-09
[WBX-153] Persist Comment Visibility

andrewwallacespeckle:
[JIRA TICKET
](https://spockle.atlassian.net/browse/WBX-153)

## Changes:
- Add new cookie using `useSynchronizedCookie` called `discussionLoadedVersionOnly`
- Defaults to true, updates via existing toggle in discussions panel
- Changed `v-if` to `v-show` to make sure this is initialised on mount


didimitrie (dismissed):

packages/frontend-2/components/viewer/comments/Comments.vue
```diff
@@ -118,6 +119,11 @@ graphql(`
   }
 `)
 
+const discussionLoadedVersionOnly = useSynchronizedCookie<boolean>(
```
fabis94, line 122: Would be cleaner to put all of this together with where threadFilters actually get set up. In setup.ts `setupResourceRequest()` we define the empty `threadFilters` ref.

fabis94 (changes requested):
minor comment, looks fine otherwise

fabis94 (changes requested):
i'm confused about how this is supposed to work - i don't see any cookie ref anywhere anymore
is it actually persisted if you reload the page?
from the PR it just looks like you're syncing 2 refs to be in sync, but there's no logic for persisting any of those values to cookies or any other storage

fabis94 (approved):

PR #2033, opened 2024-02-10
[WBX-33] Add Keystrokes to Measure Mode

andrewwallacespeckle:
[JIRA TICKET
](https://spockle.atlassian.net/browse/WBX-33)

## Changes:
- Add onKeyStroke (Escape) to Controls.vue. This will escape the active selected item. In the case of measure mode, this also cancels the in progress measurement
- When Measure mode is active, pressing delete or backspace, will delete the in selected measurement


didimitrie (dismissed):

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -441,4 +442,12 @@ onMounted(() => {
 watch(isSmallerOrEqualSm, (newVal) => {
   activeControl.value = newVal ? 'none' : 'models'
 })
+
+onKeyStroke('Escape', () => {
```
fabis94, line 446: Preferably we want to keep all of these key stroke composables together with the rest of the code that sets up the relevant integration. There's all of those composables in postSetup that set up various integrations, I bet there's one for measurements and you can put all of these onKeyStroke invocations there with the rest of the setup.

fabis94 (changes requested):
see comment

packages/frontend-2/components/viewer/Controls.vue
andrewwallacespeckle: I have moved all measurement logic to postSetup, but this onKeyStroke is still needed here to close the activeControl (for any panel, not just measurements)

fabis94 (approved):

PR #2072, opened 2024-02-22
Viewer API Improvements

AlexandruPopovici:
## Description & motivation
A bunch of small changes that came out after the documentation process mostly:
- Improved typing
- Removed unused members
- Other small non-functional changes

## Checklist:


packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -50,7 +50,7 @@ export function useStateSerialization() {
   ): SerializedViewerState => {
     const { concreteResourceIdString } = options || {}
 
-    const camControls = state.viewer.instance.getExtension(CameraController).controls
+    const camControls = state.viewer.instance.getExtension(CameraController)!.controls
```
fabis94, line 53: @AlexandruPopovici 

Every time you use `!` or `xx as SomeType` you're essentially overriding TypeScript's typing system and potentially introducing a location in code where an error can occur (just like they've been happening so far) due to the object not existing or being the wrong type.

You should use these as sparingly as possible, and instead change your code to explicitly check for the correct type through type guards or for existence of properties.

I'd even go as far as to mark `!` usages as an error through ESLint (we have that on the server side)

packages/viewer/package.json
```diff
@@ -54,6 +54,7 @@
   "dependencies": {
     "@speckle/objectloader": "workspace:^",
     "@types/flat": "^5.0.2",
+    "@types/underscore": "1.11.15",
```
fabis94, line 57: Are we really using underscore somewhere? 🤔

fabis94 (changes requested):
There's way too many usages of the non-null assertion operator (`!`), which essentially is the same as just being in non-strict mode (not having to check if the thing actually exists).

You should remove the majority (99%) of these, if not all of them. If you have questions about any specific usage, I can help you with making it work without `!` while still retaining good DX.

As an example: If you're accessing an object that you're absolutely sure will exist in 99% of cases, and might not only exist very rarely (e.g. before full initialization), you may do something like this:

```js
const getExtensionStrict = () => {
  const extension = this.viewer.getExtension(...)
  if (!extension) {
    throw new MissingExtensionError(...)
  }

  return extension
}
```

That way the return type will no longer be nullable, but if for some reason something does invoke the function before it's properly initialized, you at least get a properly typed error explaining what happened.

That said - in many cases it's just better to check for the existence of the thing that you're trying to access, especially if there's real reason to believe that it might not exist in some cases

packages/viewer/tsconfig.json
```diff
@@ -4,13 +4,15 @@
     "lib": ["DOM"],
```
fabis94, line 4: Add this rule to your eslint config

```
    '@typescript-eslint/no-non-null-assertion': 'error'

```

AlexandruPopovici:
@fabis94 I'm afraid I partly disagree with you on 
"There's way too many usages of the non-null assertion operator (!), which essentially is the same as just being in non-strict mode (not having to check if the thing actually exists)."

All the `!` I've used because typescript is incapable of detecting if something really is in danger of being null when accessed. And this happens because I like to initialize a lot of things with null, but in order to do that I need to add `null` to it's type at which point every time that member is accessed typescript will bitch about it possibly being null, **even if due to how three.js, the bvh library or the viewer itself work** the supposition that the member is false is impossible. Also, I get constantly hit with `Property X has no initializer and is not definitely assigned in the constructor` which is absolute garbage. The idea that you **must** members initialized or assigned in the constructor is absolute horseshit

 I'll happily make any changes that actually add value, but I'm afraid that catering to typescript's ridiculous notions of "type-safety" does not.

packages/viewer/package.json
AlexandruPopovici: Yeah, it's used once, but it's probably not strictly necessary

AlexandruPopovici:
@fabis94 

I've made quite a few changes that address the issue you've raised in the previous review
- Added '@typescript-eslint/no-non-null-assertion': 'error' so there are definitely no more non-null-assertions
- Possible null references are treated either by guarding against them, in the situations where those references can actually be `null`. There are some situations where there is no possibility if the references being null, but the compiler cannot determine that, In those scenarios, the library will throw if it encounters nulls. In these impossible scenarios, nulls indicate an unrecoverable fault and the execution cannot continue.  
- I've added `"strictPropertyInitialization": false` in tsconfig. Like I previously said the rule of having all members strictly initialized is ridiculous and I see no need for enforcing it. **However** if you feel it's critical I would ask you ignore it for now and we can have a separate discussion about it in the next iteration
- Removed the `underscore` as a dependency
- A lot of other stuff that I can't specifically remember right now, some of it was a welcomed addition, some was typescript catering

fabis94 (dismissed):
didn't go through everything again, but looks good, good job

PR #2079, opened 2024-02-26
FE2 Viewer UX Improvements

andrewwallacespeckle:
## Description & motivation
Various FE2 Viewer UX Improvements. A lot around the "Selection Info", but other general fixes included too. Please see changes list below for full list. 

## Changes:
- Added new FE2 Component called ViewerSidebar. This is the reusable part of the sidebar, without any content. 
- Updated ViewerSelectionSidebar to use new ViewerSidebar
- Change Selection Info Pagination count from 42 to 20
- Tidied up "Toast" when only title is displayed. 
- Added a copy button to the data items in selection info
- Changed "Delete Measurement" to "Clear Measurements" and make it clear all


packages/frontend-2/components/viewer/AnchoredPoints.vue
andrewwallacespeckle, line 1: I removed the unnecessary .value here

packages/frontend-2/components/viewer/Sidebar.vue
```diff
@@ -0,0 +1,150 @@
...
+  const element = resizableElement.value as HTMLElement | null
+  if (element) {
+    element.classList.remove('transition-all')
+  }
+
+  startX = event.clientX
+  startWidth = width.value
+
+  document.addEventListener('mousemove', resizing)
+  document.addEventListener('mouseup', stopResizing)
```
andrewwallacespeckle, line 117: Only 1 event listener setup on mounted. Use vueUse

No need to use document here, just use the element that's needed. 

When you start resizing, set a ref that we are in resize mode. Then use mouseMove to check if in resize mode, if so change resize values.

andrewwallacespeckle: Create composable for mouse move events if no vueUse function exists

fabis94 (changes requested):
discussed necessary changes in call

PR #2094, opened 2024-02-28
feat: make automate app registration optional based on the existence of the url config

gjedlicska:
## Checklist:


packages/server/.env-example
```diff
@@ -125,5 +125,3 @@ STRATEGY_LOCAL=true
 ############################################################
 # FRONTEND_HOST=127.0.0.1
 # FRONTEND_PORT=8081
-
-SPECKLE_AUTOMATE_URL="http://127.0.0.1:3030"
```
fabis94, line 129: Why remove it? There was still use in it being there for local envs

packages/server/modules/auth/defaultApps.js
```diff
@@ -22,7 +22,7 @@ module.exports = async () => {
   await registerOrUpdateApp({ ...SpeckleConnectorApp })
   await registerOrUpdateApp({ ...SpeckleExcel })
   await registerOrUpdateApp({ ...SpecklePowerBi })
-  await registerOrUpdateApp({ ...SpeckleAutomate })
+  if (speckleAutomateUrl()) await registerOrUpdateApp({ ...SpeckleAutomate })
```
fabis94, line 25: Seems like a very localized fix that doesn't resolve the root issue with .env vars overwriting redirectUrls on startup. I'd just have made it so that whenever apps are retrieved on the repo/service level, they're overwritten with whatever's in defaultApps.js (if there are relevant apps there).

gjedlicska: I tried to do that. It turned out to be a lot more complicated than it should be. When we query the apps back, we add a lot of info around scopes authors etc. That doesn't exist the same way on the default app definitions. It can be done, but its a complex fix for a very specific small problem, that only affects our 'latest' setup.

packages/server/.env-example
gjedlicska: This is the safeguard, to not enable overriding the automate url by default.

packages/server/modules/auth/defaultApps.js
fabis94: Isn't the same issue gonna happen with other default apps? The web app & API explorer both have redirect URLs dependant on env vars

packages/server/modules/auth/services/apps.js
```diff
@@ -30,6 +31,9 @@ module.exports = {
       .select('id', 'name', 'avatar')
       .where({ id: app.authorId })
       .first()
+
+    const defaultApp = getDefaultApp({ id })
```
fabis94, line 35: These 2 lines are repeated 3 times, I'd move this out to a reusable function, even if its just at the top of this file and not exported out

fabis94 (approved):
A couple of minor style problems, but it does look like it does the job

fabis94 (approved):

PR #2107, opened 2024-03-07
Feat (dui3): CNX-9156 Mixpanel plugin

oguzhankoral:
Composable to send HTTP requests to the mixpanel endpoint.


teocomi (commented):
I could only look at the mp data, so this is not a code review.
Please double check the server ID is correct, I only see `AEAA6ACC771AF86F1C97940EF7F8A545` which does not match anynknown ones.
And that the host app names are sent as lowercase.

oguzhankoral:
> I could only look at the mp data, so this is not a code review. Please double check the server ID is correct, I only see `AEAA6ACC771AF86F1C97940EF7F8A545` which does not match anynknown ones. And that the host app names are sent as lowercase.

Good catch! You were right, I should have use `hostname` only from `serverUrl`. Tested and validated but would be greate if you can double check @teocomi

PR #2108, opened 2024-03-08
Update Grid/List view toggle component to make the active item clearer

benjaminvo:
I've updated the Grid/List view toggle component to make it clearer which item is the active item. The old design made it appear as if the inactive item was the active item.

### Changes

- The active item has the light background color, some spacing around it, and a shadow.
- Updated tippy to have one for each of the two toggle items. "Grid View" / "List View"
- Note: I've used a few `dark:` Tailwind classes to tweak the colors in dark mode. I've heard from Andrew we don't want to do this but it didn't work well in dark mode without it.

#### Before
The item with the darker background below is the active item even though you'd think it's the white one.

CleanShot 2024-03-08 at 08 10 35@2x

CleanShot 2024-03-08 at 08 12 30@2x

#### After

CleanShot 2024-03-08 at 08 10 18@2x

CleanShot 2024-03-08 at 08 12 00@2x

### DIFF
--- a/packages/frontend-2/components/project/page/LatestItems.vue
+++ b/packages/frontend-2/components/project/page/LatestItems.vue
@@ -25,7 +25,7 @@
       <slot v-if="!hideFilters" name="filters">
         <div class="flex space-x-4 items-center grow justify-end flex-wrap">
           <slot name="filters" />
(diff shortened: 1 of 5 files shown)


andrewwallacespeckle (approved):
Huge improvement, a lot clearer now!

PR #2109, opened 2024-03-08
Make Following clearer

benjaminvo:
Make it clearer that you are following someone or being followed.

### Changes
- Dark blue outline and top centred. The top centred element can go behind menus on smaller screens but I don't think it matters that much. The current "Stop Following [...]" button is also positioned on top of the Selection Info panel (bug) so we have to find a new place for it.
- Dedicated "Stop" button 
- Not using Vue Portal

### Before

CleanShot 2024-03-08 at 14 17 19@2x

CleanShot 2024-03-08 at 14 17 28@2x

CleanShot 2024-03-08 at 14 17 40@2x

### After

CleanShot 2024-03-08 at 14 13 21@2x

CleanShot 2024-03-08 at 14 13 48@2x

CleanShot 2024-03-08 at 14 13 12@2x

CleanShot 2024-03-08 at 14 14 02@2x

### DIFF
--- a/packages/frontend-2/components/viewer/AnchoredPoints.vue
+++ b/packages/frontend-2/components/viewer/AnchoredPoints.vue
@@ -86,30 +86,33 @@
       "
     >
       <div
-        class="w-full h-full outline -outline-offset-0 outline-8 rounded-md outline-blue-500/40"
+        class="w-full h-full outline -outline-offset-0 outline-8 rounded-md outline-blue-500"
       >
-        <div class="absolute bottom-4 right-4 p-2 pointer-events-auto">
-          <Portal to="pocket-right">
+        <div
+          class="absolute top-0 left-[50%] -translate-x-1/2 pointer-events-auto bg-primary text-white text-xs px-3 h-8 flex items-center rounded-b-md cursor-default"
+        >
+          <div v-if="spotlightUserSessionId && spotlightUser">
+            Following {{ spotlightUser?.userName.split(' ')[0] }}
             <FormButton
-              v-if="spotlightUserSessionId && spotlightUser"
+              color="secondary"
               size="xs"
-              class="truncate"
+              class="ml-1 -mr-1.5"
               @click="() => (spotlightUserSessionId = null)"
             >
-              <span>Stop Following {{ spotlightUser?.userName.split(' ')[0] }}</span>
+              <span>Stop</span>
             </FormButton>
-            <div
-              v-else
-              v-tippy="followers.map((u) => u.name).join(', ')"
-              class="text-xs p-2 font-bold text-primary"
-            >
-              Followed by {{ followers[0].name.split(' ')[0] }}
-              <span v-if="followers.length > 1">
-                & {{ followers.length - 1 }}
-                {{ followers.length - 1 === 1 ? 'other' : 'others' }}
-              </span>
-            </div>
-          </Portal>
+          </div>
+          <div
+            v-else
+            v-tippy="{ placement: 'bottom' }"
+            :content="followers.map((u) => u.name).join(', ')"
+          >
+            Followed by {{ followers[0].name.split(' ')[0] }}
+            <span v-if="followers.length > 1">
+              & {{ followers.length - 1 }}
+              {{ followers.length - 1 === 1 ? 'other' : 'others' }}
+            </span>
+          </div>
         </div>
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


benjaminvo:
The last commit adds rounded borders on each side of the follow element through SVGs:  

CleanShot 2024-03-11 at 09 38 39@2x

packages/frontend-2/components/viewer/AnchoredPoints.vue
```diff
@@ -86,30 +86,53 @@
...
               Followed by {{ followers[0].name.split(' ')[0] }}
               <span v-if="followers.length > 1">
                 & {{ followers.length - 1 }}
                 {{ followers.length - 1 === 1 ? 'other' : 'others' }}
               </span>
             </div>
-          </Portal>
+          </div>
+          <svg
+            class="mt-1"
```
andrewwallacespeckle, line 128:
```suggestion
            class="absolute right-[0.5px] top-1"
```

packages/frontend-2/components/viewer/AnchoredPoints.vue
```diff
@@ -86,30 +86,53 @@
...
-              v-if="spotlightUserSessionId && spotlightUser"
-              size="xs"
-              class="truncate"
-              @click="() => (spotlightUserSessionId = null)"
-            >
-              <span>Stop Following {{ spotlightUser?.userName.split(' ')[0] }}</span>
-            </FormButton>
+        <div class="absolute top-0 left-[50%] -translate-x-1/2 flex">
+          <svg
+            class="mt-1"
```
andrewwallacespeckle, line 93:
```suggestion
            class="absolute left-[0.5px] top-1"
```

packages/frontend-2/components/viewer/AnchoredPoints.vue
```diff
@@ -86,30 +86,53 @@
...
-          <Portal to="pocket-right">
-            <FormButton
-              v-if="spotlightUserSessionId && spotlightUser"
-              size="xs"
-              class="truncate"
-              @click="() => (spotlightUserSessionId = null)"
-            >
-              <span>Stop Following {{ spotlightUser?.userName.split(' ')[0] }}</span>
-            </FormButton>
+        <div class="absolute top-0 left-[50%] -translate-x-1/2 flex">
```
andrewwallacespeckle, line 91: It's not worth changing here, but I'd try and avoid these positioning hacks, trying to use flex where possible. Something like this does the same using only flex
image

packages/frontend-2/components/viewer/AnchoredPoints.vue
benjaminvo: @andrewwallacespeckle These changes break the layout for me. Why do this instead of the top margin?

andrewwallacespeckle (changes requested):
Looks good to me! I like moving the button here, it makes a lot more sense. 

I've commented a quick fix for the alignment issue you mentioned around the SVG corners. 

There is also another comment that isn't worth fixing, but worth thinking about in future changes.

packages/frontend-2/components/viewer/AnchoredPoints.vue
benjaminvo: Ah, yes much cleaner. I've been used to centering absolute positioned elements like this. I've updated it.

packages/frontend-2/components/viewer/AnchoredPoints.vue
andrewwallacespeckle: Sorry, you will need to add `px-2` to the parent to add padding. 

Margin works, but when the item is being absolutely positioned (as in my suggestion) it makes more sense to position it using top, left etc. If not we would have to say "top-0 mt-1". 

This is how this looks for me:
image

benjaminvo:
@andrewwallacespeckle Thanks for the review! I updated the alignment. The SVG corner problem isn't fixed with what you suggested – it actually just breaks it completely for me :D.

andrewwallacespeckle:
> @andrewwallacespeckle Thanks for the review! I updated the alignment. The SVG corner problem isn't fixed with what you suggested – it actually just breaks it completely for me :D.

This seems to have fixed the svg alignment too. I'm no longer getting the slight gap, even without the other chanegs

benjaminvo:
Same! Nice

andrewwallacespeckle (approved):

PR #2116, opened 2024-03-12
DRAFT: Gergo/new automate module

fabis94:


fabis94:
@gjedlicska No point in pushing anything in here, I'm only using this as a base & a reference. My branch is fabians/fe2-automate-integration

gjedlicska:
> @gjedlicska No point in pushing anything in here, I'm only using this as a base & a reference. My branch is fabians/fe2-automate-integration

I'm still in a need to figure out some backend stuff around triggers and events. If i push to this branch would it conflict with your FE efforts @fabis94 ?
Ideally you would merge stuff from my branch into yours.

fabis94:
@gjedlicska Probably, because my branch is not actually downstream from yours. I squashed all of your changes, omited the GQL schema bit and created a new branch from that. So if I'd try to merge this in, I'd have a lot of conflicts between duplicate versions of the same code (I think)

You could push your changes to my branch directly, or I can also go over your subsequent commits later on and manually get them into my branch

PR #2123, opened 2024-03-13
Add breadcrumbs to discussions and models. Temp change to routing

andrewwallacespeckle:
## Changes:
- Update projectDiscussionsRoute & allProjectModelsRoute so they go to project page. This is temporary until the Tab routing is done. 
- Add breadcrumbs to both models and discussions page, incase someone has a bookmark saved, they can still use breadcrumbs to navigate.


benjaminvo (approved):
Approved. But I wonder if we could just redirect the All models and All discussions pages to go to the `projectDiscussionsRoute` and `allProjectModelsRoute` on load?

Can be a follow up improvement if you want to merge this. This is also fine for now.

PR #2140, opened 2024-03-15
Hidden Object Outlines

AlexandruPopovici:
## Description & motivation
Handles https://spockle.atlassian.net/browse/WBX-383. When hiding selected objects, they will be shown as outlines only

## Changes:
- `stencilOutlines` from `MaterialOptions` has changed from a `boolean` to 
  ```
  enum StencilOutlineType {
    NONE,
    OVERLAY,
    OUTLINE_ONLY
  }
  ```
- All `Batch` implementations now have a dedicated `getDepth` function that returns the visibility ranges required for the `DepthPass`, instead of using the opaque visibility ranges
- The material's `colorWrite` property is now being factored in when sorting the draw ranges
- `stencilOutlines` and `pointSize` propertie are being factored in when computing the material's hash

## Checklist:


benjaminvo:
@AlexandruPopovici It works great! Much better UX. It only works though if hiding from the selection info panel, not if you hide an object from the Scene Explorer. But I think it's because when hiding from the Scene Explorer we also instantly deselect the object. This is a pure frontend issue, right? == Is a separate frontend ticket after this is merged.

AlexandruPopovici:
> you
Yes, it does seem that hiding from the scene explorer, the object gets de-selected. If that's the case, then the outlines can't be visible, since the object is not selected anymore

benjaminvo (dismissed):
Approving this from a testing perspective. Would like a quick code check from Fabs too

fabis94 (approved):

PR #2145, opened 2024-03-19
feat(fe2): upgrade to nuxt 3.11

fabis94:
Tested out a dev & prod build, also tested out various pages locally


packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -33,7 +33,7 @@
                 widthClasses
               ]"
               :as="isForm ? 'form' : 'div'"
-              @submit.prevent="onSubmit"
+              @submit.prevent="onSubmit || noop"
```
fabis94, line 36: @andrewwallacespeckle An undefined event handler (onSubmit is optional) breaks the app fyi

PR #2154, opened 2024-03-22
feat(fe2): add survicate nps survey

andrewwallacespeckle:
## Description & motivation
We are integrating Survicate into the NWA. This is a survey tool for collecting user feedback. 

Currently, we are only testing this, and using it for 1 survey - the Net Promoter Score. This should be displayed when:
- The user has used the NWA for at least 3 days. We are using the existing `onboardingOrFeedbackDate` cookie for this. If it doesn't exist, it will be created and set as todays date. 
- The user has 5 Versions across all projects

https://developers.survicate.com/javascript/installation/

## Changes:
- Added 2 new env - `NUXT_PUBLIC_SURVICATE_WORKSPACE_KEY` & `NUXT_PUBLIC_SURVICATE_NPS_SURVEY_ID`
- Added `survicate.ts` plugin
- Removed old Feedback Request banner/dialog, although we are keeping the submit feedback link in the menu, and it should keep linking to the Google Form.

## To-do before merge:
Some settings for this, lives in the Survicate admin panel. Things like how often this should be shown to users. 

For the purpose of testing, a lot of settings were turned off, so the Survey loads on every load. 

Before merge - we need to revert these settings.


packages/frontend-2/lib/auth/composables/activeUser.ts
andrewwallacespeckle, line 1: @fabis94 The original brief here was that we count the total versions across all projects, and only show if there are 5 versions. I have added a project.totalCount here for now, but is there a good way of getting the total version count instead?

fabis94: I added limit: 0, cause that makes it a lighter query on the server-side

packages/frontend-2/nuxt.config.ts
```diff
@@ -50,6 +50,8 @@ export default defineNuxtConfig({
       baseUrl: '',
       mixpanelApiHost: 'UNDEFINED',
       mixpanelTokenId: 'UNDEFINED',
+      survicateWorkspaceKey: 'UNDEFINED',
```
fabis94, line 53: 1. You should put these in .env.example as well, otherwise devs won't know of these keys and won't be able to set them up locally
2. Don't do 'UNDEFINED' as a default value, it will cause survicate to always be loaded even if the env vars aren't set, but with invalid values ('UNDEFINED') which will just cause erroring requests to survicate

packages/frontend-2/plugins/survicate.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  )
+
+  const onboardingOrFeedbackDate = onboardingOrFeedbackDateString.value
+    ? new Date(onboardingOrFeedbackDateString.value)
+    : new Date()
+
+  const shouldShowSurvey = checkSurveyDisplayConditions(onboardingOrFeedbackDate)
+
+  if (!process.client || !isLoggedIn.value || !shouldShowSurvey) {
```
fabis94, line 26: if you don't need this to run on the serverside, add a .client.ts prefix to the plugin

packages/frontend-2/plugins/survicate.ts
```diff
@@ -0,0 +1,86 @@
...
+    const survicateInstance = getSurvicateInstance()
+
+    if (!survicateInstance) {
+      throw new Error('Survicate instance is not available after initialization.')
+    }
+
+    const { distinctId } = await initUser()
+
+    if (distinctId) {
+      survicateInstance.setVisitorTraits({ distinctId })
```
fabis94, line 52: Just set onAuthStateChange immediate: true so that you don't have to do this

packages/frontend-2/plugins/survicate.ts
```diff
@@ -0,0 +1,86 @@
...
+      (user, { resolveDistinctId }) => {
+        const distinctId = resolveDistinctId(user)
+        if (distinctId) {
+          survicateInstance.setVisitorTraits({ distinctId })
+        }
+      },
+      { immediate: false }
+    )
+
+    survicateInstance.showSurvey(survicateNpsSurveyId as string, {
```
fabis94, line 66: unnecessary type cast

packages/frontend-2/plugins/survicate.ts
```diff
@@ -0,0 +1,86 @@
...
+      forceDisplay: true
+    })
+
+    app.provide('survicate', survicateInstance)
```
fabis94, line 70: The new syntax for providing stuff is to just return from the plugin:

```js
return {
  provide: {
    survicate: survicateInstance
  }
}
```

fabis94 (changes requested):
left some comments

packages/frontend-2/plugins/survicate.client.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const logger = useLogger()
+  const onAuthStateChange = useOnAuthStateChange()
+  const prepareSurvey = useInitMainSurvey()
+
+  try {
+    const { initSurvicate, getSurvicateInstance } = await import(
+      '@survicate/survicate-web-surveys-wrapper/widget_wrapper'
+    )
+    await initSurvicate({ workspaceKey: survicateWorkspaceKey })
```
fabis94, line 32: We should probably skip all this if `!survicateWorkspaceKey?.length`, I dunno what's gonna happen if u try to initialize it w/ an empty key

fabis94 (approved):

PR #2155, opened 2024-03-22
feat(fe2): vimeo video wrapper component and update links from youtube

andrewwallacespeckle:
## Description & motivation
We have swapper to Vimeo from YouTube. I have added a new component to make this consistent. 

## Changes:
- Added a new VideoWrapper component into UI-Components
- Swapped "Install Manager" Video
- Swapped "Log In" Video
- Swapped "First Model Upload" Video
- Used same video wrapper for consistency

## Screenshots:
image
image
image


packages/ui-components/src/components/common/VideoWrapper.stories.ts
```diff
@@ -0,0 +1,61 @@
...
+  vimeoId: string
+}>
+
+export default {
+  title: 'Components/Common/VideoWrapper',
+  component: CommonVideoWrapper,
+  argTypes: {
+    autoplay: {
+      control: { type: 'boolean' },
+      defaultValue: false
```
fabis94, line 17: defaultValue is redundant if you set up all args in the Default story (and use that as a base for all other stories)

packages/ui-components/src/components/common/VideoWrapper.vue
```diff
@@ -0,0 +1,44 @@
+<template>
```
fabis94, line 1: The name is kinda weird
1. CommonVideoEmbed is more self explanatory
2. Better yet - CommonVimeoEmbed, since this doesn't handle any other kind of videos (and we might theoretically still want to embed YT videos in FE2 or other apps that might use ui-components)

fabis94 (changes requested):

fabis94 (approved):

PR #2193, opened 2024-04-04
Extension Access Changes

AlexandruPopovici:
## Description & motivation
This was a loose end from switching to API 2.0. The extension creation and access approach was over complicated by using `Providers` as interfaces that some extensions could implement and were used to decide what gets injected where. With this PR, the concept of Providers is completely removed and instead we're using explicit Extension types and subtypes. This simplifies working with Extensions a lot while keeping the core idea for their creation and access intact

## Checklist:


packages/viewer/src/modules/Viewer.ts
```diff
@@ -107,7 +108,7 @@ export class Viewer extends EventEmitter implements IViewer {
     return extension as T
   }
 
-  public getExtension<T extends Extension>(type: typeof Extension): T {
+  public getExtension<T extends Extension>(type: Constructor<T>): T {
```
fabis94, line 111: @AlexandruPopovici  The way this function was typed was off, which is why you had to manually add the <CameraController> generic argument everywhere (`getExtension<CameraController>(CameraController)`). There were usages in FE2 that did not have that generic argument so obviously those were still broken.

I fixed the function type to correctly resolve the class type from the generic argument (previously you had `typeof Extension` which was in no way related to the generic T arg, so no inference was made), so now it should work even without having to manually specify the generic everywhere.

You should go through all of those usages where you added the arg and remove them

PR #2207, opened 2024-04-12
feat(fe2): move settings to tab on projects page

andrewwallacespeckle:
## Description & motivation
[LINEAR TICKET](https://linear.app/speckle/issue/WEB-684/move-settings-to-a-tab)

Project settings have been moved from the Team block, and into it's own tab in the new tabbed layout. 

Within settings, there are 3 subpages:
- General
- Collaborators
- Webhooks

Various items are disabled when the user does not have the correct role. 

## Changes:
- Moved Settings to new tab
- Removed Settings from Team Block
- Added new input style - "foundation", and replaced all input fields with this style in the project page. 
- Added new RadioGroup component to ui-components
- Added new SettingsBlock component to fe2

## Screenshots:
image
image
image


packages/frontend-2/components/project/page/InviteDialog.vue
```diff
@@ -48,34 +51,37 @@
...
 import { useServerInfo } from '~~/lib/core/composables/server'
 
 type InvitableUser = UserSearchItem | string
 
 const props = defineProps<{
-  project: ProjectPageTeamDialogFragment
-  defaultOpen: boolean
+  project?: ProjectPageSettingsCollaboratorsQuery
+  disabled?: boolean
+  disabledMessage?: string
```
fabis94, line 78: What's the point of this prop, if the message never gets used? We do check for the existence of this prop in the template, but then it's not clear why this exists and it probably has to be renamed to something more intuitive.

From the name I would think that this allows feeding in a custom disabled message

packages/frontend-2/components/project/page/InviteDialog.vue
```diff
@@ -88,9 +94,22 @@ const { userSearch, searchVariables } = useUserSearch({
     limit: 5
   }))
 })
-const { collaboratorListItems } = useTeamDialogInternals({
-  props: toRefs(props)
-})
+
+const projectId = computed(() => route.params.id as string)
```
fabis94, line 98: Not ideal to rely on route params in a non-page component, because it could also be used in different pages (and the Invite Dialog is already on 2 different pages).  Preferably we want to keep the projectId prop here

packages/frontend-2/components/project/page/settings/collaborators/Collaborators.vue
```diff
@@ -0,0 +1,188 @@
+<template>
+  <ProjectPageSettingsBlock title="Collaborators" :icon="UsersIcon">
+    <template #introduction>
+      <p>Invite new members or edit roles for existing ones.</p>
+    </template>
+    <template #topButtons>
```
fabis94, line 6: Slot names should be kebab-cased because of https://vuejs.org/style-guide/rules-strongly-recommended.html#prop-name-casing

(Slot names are specified similarly to props in the template)

packages/frontend-2/components/project/page/team/Dialog.vue
```diff
@@ -1,32 +1,12 @@
...
+    <div class="flex flex-col text-foreground"></div>
   </LayoutDialog>
 </template>
 <script setup lang="ts">
 import type { ProjectPageTeamDialogFragment } from '~~/lib/common/generated/gql/graphql'
 import { graphql } from '~~/lib/common/generated/gql'
-import { useTeamDialogInternals } from '~~/lib/projects/composables/team'
 import { OpenSectionType } from '~~/lib/projects/helpers/components'
 
 graphql(`
```
fabis94, line 12: The GQL fragment for this dialog probably needs less data now, no?

packages/frontend-2/components/project/page/settings/collaborators/Collaborators.vue
```diff
@@ -0,0 +1,188 @@
...
+              "
+            >
+              Cancel Invite
+            </FormButton>
+          </div>
+        </template>
+      </div>
+    </div>
+
+    <ProjectPageInviteDialog
```
fabis94, line 67: Preferably this would have its own fragment

packages/frontend-2/components/project/page/settings/general/General.vue
```diff
@@ -0,0 +1,135 @@
...
+import { useQuery } from '@vue/apollo-composable'
+import type { ProjectUpdateInput } from '~~/lib/common/generated/gql/graphql'
+import { useUpdateProject } from '~~/lib/projects/composables/projectManagement'
+import { graphql } from '~~/lib/common/generated/gql'
+
+const projectPageSettingsGeneralQuery = graphql(`
+  query ProjectPageSettingsGeneral($projectId: String!) {
+    project(id: $projectId) {
+      id
+      team {
```
fabis94, line 62: This bit only seems to be relevant in the Leave block, so it should be moved into its fragment and then "canLeaveProject" should be a computed inside of there

packages/frontend-2/components/project/page/settings/general/block/Access.vue
```diff
@@ -0,0 +1,80 @@
...
+  fragment ProjectPageSettingsGeneralBlockAccess_Project on Project {
+    id
+    visibility
+  }
+`)
+
+const props = defineProps<{
+  project: ProjectPageSettingsGeneralBlockAccess_ProjectFragment
+  disabled?: boolean
+  disabledMessage?: string
```
fabis94, line 42: Also not clear why this prop is there if it isn't really used (as a string, at least)

packages/frontend-2/components/project/page/settings/general/block/Access.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <ProjectPageSettingsBlock
+    background
+    title="Access"
+    :icon="LockClosedIcon"
+    :disabled-message="
+      disabled && disabledMessage ? 'You must be a project owner' : undefined
```
fabis94, line 7: The message will never be rendered unless you pass in the 'disabledMessage' prop, is that expected?

packages/frontend-2/components/project/page/settings/general/block/Access.vue
```diff
@@ -0,0 +1,80 @@
...
+
+const props = defineProps<{
+  project: ProjectPageSettingsGeneralBlockAccess_ProjectFragment
+  disabled?: boolean
+  disabledMessage?: string
+}>()
+
+const emit = defineEmits(['update-visibility'])
+
+const selectedOption = ref(props.project.visibility ?? ProjectVisibility.Private)
```
fabis94, line 47: Not a big deal, but `||` is in most cases better than `??`, because surely you'd want it to fall back to that value even if in the case of a bug `props.project.visibility` is `false` or `''` (empty string) or `0` or any other kind of falsy value

packages/frontend-2/components/project/page/settings/general/block/Delete/Delete.vue
```diff
@@ -0,0 +1,52 @@
...
+import { TrashIcon } from '@heroicons/vue/24/outline'
+import { graphql } from '~~/lib/common/generated/gql'
+import type { ProjectPageSettingsGeneralBlockDelete_ProjectFragment } from '~~/lib/common/generated/gql/graphql'
+
+graphql(`
+  fragment ProjectPageSettingsGeneralBlockDelete_Project on Project {
+    id
+    name
+    role
+    models {
```
fabis94, line 38: You should definitely use limit: 0 for both models & commentThreads, because
1. That way you access the same cache thet is used in the project page query, where these same params are used
2. The server does way less work, because it wont actually try to resolve any items

packages/frontend-2/components/project/page/settings/general/block/Discussions.vue
```diff
@@ -0,0 +1,81 @@
...
+  }
+`)
+
+const props = defineProps<{
+  project: ProjectPageSettingsGeneralBlockDiscussions_ProjectFragment
+  disabled?: boolean
+  disabledMessage?: string
+}>()
+
+const emit = defineEmits(['update-comments-permission'])
```
fabis94, line 47: Untyped payload

packages/frontend-2/components/project/page/settings/general/block/Discussions.vue
```diff
@@ -0,0 +1,81 @@
...
+
+const selectedOption = ref(props.project.allowPublicComments ? 'anyone' : 'teamMembers')
```
fabis94, line 49: These available values could be an enum for better DX

packages/frontend-2/components/project/page/team/Block.vue
```diff
@@ -3,50 +3,44 @@
...
     </template>
     <template #default>
-      <ProjectPageTeamDialog
+      <ProjectPageInviteDialog
         v-model:open="dialogOpen"
-        :project="project"
-        :open-section="openSection"
+        :disabled="!isOwner"
+        :disabled-message="
+          !isOwner ? 'Sorry, you must be a project owner to invite users' : undefined
```
fabis94, line 31: Misleading API - the disabledMessage string is never actually used. It's just used like a boolean

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -268,24 +268,26 @@ export function useUpdateProject() {
 
   return async (
     update: ProjectUpdateInput,
+    customSuccessMessage?: string,
```
fabis94, line 271: Why not put this inside of `options`? That's what that's for. With this kind of backwards incompatible change (adding a new param before pre-existing params) you have to go through all usages of this function and make sure `options` is now the third parameter

packages/frontend-2/lib/projects/composables/team.ts
```diff
@@ -27,22 +27,17 @@ export function useTeamManagePermissionsInternals(params: {
...
-export function useTeamDialogInternals(params: {
-  props: {
-    project: Ref<ProjectPageTeamDialogFragment>
-  }
-}) {
-  const {
-    props: { project }
-  } = params
+export function useTeamInternals(
+  projectData: ComputedRef<ProjectPageSettingsCollaboratorsQuery | undefined>
```
fabis94, line 31: Ideally this should only ask for a fragment of what it needs. That way it can be re-used in more pages/components without having to request the entire collaborators query.

I already saw duplicate definitions of `isOwner` checks in various components

packages/frontend-2/lib/projects/graphql/fragments.ts
```diff
@@ -81,3 +81,47 @@ export const projectUpdatableMetadataFragment = graphql(`
     allowPublicComments
   }
 `)
+
+export const projectPageLatestItemsModelsFragment = graphql(`
```
fabis94, line 85: I moved these over from a couple of unnecessary Vue components I deleted

packages/frontend-2/pages/projects/[id]/index/settings.vue
```diff
@@ -0,0 +1,86 @@
...
+
+const activeSettingsPageTab = computed({
+  get: () => {
+    const path = route.path
+    if (path.includes('/settings/collaborators')) return settingsTabItems.value[1]
+    if (path.includes('/settings/webhooks')) return settingsTabItems.value[2]
+    return settingsTabItems.value[0]
+  },
+  set: (val: LayoutPageTabItem) => {
+    switch (val.id) {
```
Comment (fabis94) [inline packages/frontend-2/pages/projects/[id]/index/settings.vue:71]:
```ts
switch (val.id) {
      case 'collaborators':
        router.push(projectCollaboratorsRoute(projectId.value))
        break
      case 'webhooks':
        router.push(projectWebhooksRoute(projectId.value))
        break
      case 'general':
      default:
        router.push(projectSettingsRoute(projectId.value))
        break
    }
```

packages/ui-components/src/components/form/RadioGroup.stories.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const Default: RadioGroupStoryType = {
+  render: (args) => ({
+    components: { RadioGroup },
+    setup() {
+      return { args }
+    },
+    template: `
+      <div>
+        <RadioGroup v-bind="args" @update:modelValue="args['update:modelValue']" />
```
fabis94, line 44: There's no two-way binding between the Storybook arg and the model value. If I change the value by clicking on the button, I won't see the new value pop up in the controls panel. I suggest using the `useStorybookVmodel` composable I created to simplify doing so

packages/ui-components/src/components/form/RadioGroup.vue
```diff
@@ -0,0 +1,110 @@
...
+  value: string
+  title: string
+  introduction?: string
+  icon: ConcreteComponent
+  help?: string
+}
+
+const props = defineProps<{
+  options: OptionType[]
+  modelValue: string
```
fabis94, line 91: This could probably be shortened w/ `defineModel`

packages/ui-components/src/components/form/TextInput.vue
```diff
@@ -305,13 +306,15 @@ const sizeClasses = computed((): string => {
   switch (props.size) {
     case 'sm':
       return 'h-6 text-sm'
+    case 'md':
+      return 'h-8 text-sm'
```
fabis94, line 310: The difference between md and base is negligible, do we really want to add extra complexity here by adding this new size?

packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -72,7 +72,7 @@
                     v-for="(button, index) in buttons"
                     :key="index"
                     v-bind="button.props"
-                    :disabled="button.disabled"
+                    :is-disabled="button.disabled"
```
fabis94, line 75: There's no such prop on `FormButton`

packages/ui-components/src/components/layout/PageTabs.vue
```diff
@@ -101,38 +177,82 @@ const activeItemRef = computed(() => {
...
     }
-  )
-}
+    checkArrowsVisibility()
+  }
+)
 </script>
+<style>
+/* Hide scrollbar for Chrome, Safari and Opera */
+.hide-scrollbar::-webkit-scrollbar {
```
fabis94, line 249: Weird, why not just disable overflow? And if we do in fact want scrolling, then surely hiding the scrollbar is bad accessibility?

fabis94 (changes requested):
Mostly various minor issues

packages/ui-components/src/components/layout/PageTabs.vue
andrewwallacespeckle: We don't want to disable overflow, as it does still scroll. The issue with the scrollbar is that it looked ugly sitting below the menu items. It also obscured the sliding underline which made it hard to see the active item. 

As an alternative to the scrollbar, I added arrows and a fade to each side of the menu, when there was scroll available:

https://github.com/specklesystems/speckle-server/assets/139135120/e83fe5ea-3612-4562-91d1-884b21347870

This is based on how Dribble handle this problem, which Benjamin suggested.

packages/ui-components/src/components/form/TextInput.vue
andrewwallacespeckle: This was added so the heights would be equal here:
image

Previously, there was a different Select style to TextInput. One used borders and the other outlines and shadows. 

I added a new button style called "foundation" to match the select style. This has an unwanted side effect where it now was a different height than the select inputs, due to the difference in border and outline. 

I added this new size to match the select inputs height. I couldn't just update base as non "foundation" inputs would be affected.

packages/ui-components/src/components/layout/PageTabs.vue
fabis94: @andrewwallacespeckle Ok, good enough. But please mention this as a comment, cause other devs might have the same question

packages/ui-components/src/components/form/TextInput.vue
fabis94: @andrewwallacespeckle Maybe it's better to just ensure the existing button & select styles are the same height? Sounds like we're adding extra complexity not even cause there's a design need, but to deal with existing styles being inconsistent

packages/frontend-2/components/project/page/settings/general/block/ProjectInfo.vue
```diff
@@ -0,0 +1,155 @@
...
+    localProjectDescription.value !== props.project.description
+  )
+})
+
+const emitUpdate = () => {
+  emit('update-project', {
+    name: localProjectName.value,
+    description: localProjectDescription.value
+  })
+  if (targetRoute.value) {
```
fabis94, line 96: Feels off that this component handles the redirect even tho it doesn't handle the saving. What if the saving fails with an error in the parent component? It's still going to redirect.

IMO the redirect is a post-save activity not an "about to save" activity and thus must be handled by the component that actually handles the saving

packages/frontend-2/components/project/page/settings/webhooks/Webhooks.vue
```diff
@@ -0,0 +1,243 @@
...
+            icon: PencilIcon,
+            label: 'Edit',
+            action: openEditWebhookDialog,
+            class: 'text-primary'
+          },
+          {
+            icon: TrashIcon,
+            label: 'Delete',
+            action: openDeleteWebhookDialog,
+            class: 'text-red-500'
```
fabis94, line 46: would prefer to use our semantic colors instead. what about the 'danger' ones?

packages/frontend-2/components/project/page/team/Block.vue
```diff
@@ -3,50 +3,41 @@
...
+import { UserPlusIcon } from '@heroicons/vue/24/outline'
+import { Roles, type Optional } from '@speckle/shared'
 import { OpenSectionType } from '~/lib/projects/helpers/components'
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
 import { graphql } from '~~/lib/common/generated/gql'
 import type { ProjectPageStatsBlockTeamFragment } from '~~/lib/common/generated/gql/graphql'
-const { activeUser } = useActiveUser()
 
 graphql(`
   fragment ProjectPageStatsBlockTeam on Project {
```
fabis94, line 43: Fragment should ask for invite dialog's fragment. Also, I notice `project` isn't even being passed in to the dialog, is that expected? Seems off

packages/frontend-2/components/project/page/team/Block.vue
```diff
@@ -66,6 +57,10 @@ const props = defineProps<{
   project: ProjectPageStatsBlockTeamFragment
 }>()
 
+const { activeUser } = useActiveUser()
+
+const isOwner = computed(() => props.project?.role === Roles.Stream.Owner)
```
fabis94, line 62: useTeamInternals composable

packages/frontend-2/components/project/page/team/Block.vue
fabis94: + useTeamInternals fragment, if you end up using that

packages/frontend-2/lib/projects/composables/team.ts
```diff
@@ -27,22 +52,17 @@ export function useTeamManagePermissionsInternals(params: {
...
-    props: { project }
-  } = params
+export function useTeamInternals(
+  projectData: ComputedRef<ProjectPageTeamInternals_ProjectFragment | undefined>
+) {
+  const { isGuest: isServerGuest, activeUser } = useActiveUser()
 
-  const { isOwner, isServerGuest, activeUser } =
-    useTeamManagePermissionsInternals(params)
+  const isOwner = computed(() => projectData.value?.role === Roles.Stream.Owner)
```
fabis94, line 60: Why duplicate the same logic `useTeamManagePermissionsInternals` already did? You should add back the `useTeamManagePermissionsInternals` invocation so that the code is more DRY

packages/ui-components/src/components/form/Button.vue
```diff
@@ -32,7 +32,7 @@ import type { Nullable, Optional } from '@speckle/shared'
 import { ArrowPathIcon } from '@heroicons/vue/24/solid'
 import type { FormButtonColor, FormButtonTextColor } from '~~/src/helpers/form/button'
 
-type FormButtonSize = 'xs' | 'sm' | 'base' | 'lg' | 'xl'
+type FormButtonSize = 'xs' | 'md' | 'sm' | 'base' | 'lg' | 'xl'
```
fabis94, line 35: Do we really need this? Doesn't appear to be used

packages/ui-components/src/components/form/TextInput.vue
```diff
@@ -92,7 +93,7 @@ import { useTextInputCore } from '~~/src/composables/form/textInput'
 import type { PropAnyComponent } from '~~/src/helpers/common/components'
 
 type InputType = 'text' | 'email' | 'password' | 'url' | 'search' | 'number' | string
-type InputSize = 'sm' | 'base' | 'lg' | 'xl'
+type InputSize = 'sm' | 'md' | 'base' | 'lg' | 'xl'
```
fabis94, line 96: Doesn't appear to be used

packages/ui-components/src/components/form/TextInput.stories.ts
```diff
@@ -31,6 +31,10 @@ export default {
     'update:modelValue': {
       type: 'function',
       action: 'v-model'
+    },
+    size: {
+      control: { type: 'select' },
+      options: ['sm', 'md', 'lg', 'base', 'xl']
```
fabis94, line 37: md needs removing

fabis94 (changes requested):
nearly there, mostly various minor issues

packages/frontend-2/components/project/page/settings/general/General.vue
```diff
@@ -0,0 +1,121 @@
...
+
+  const updatePayload: ProjectUpdateInput = {
+    id: project.value.id,
+    ...updates
+  }
+
+  const options = customSuccessMessage ? { customSuccessMessage } : {}
+
+  updateProject(updatePayload, options).then(() => {
+    if (onComplete) {
```
fabis94, line 116: Do we really want to do this regardless of whether the project update succeeded or failed?

packages/frontend-2/components/project/page/settings/general/block/ProjectInfo.vue
```diff
@@ -0,0 +1,159 @@
...
+  emit('update-project', {
+    name: localProjectName.value,
+    description: localProjectDescription.value,
+    onComplete: handleRedirection
+  })
+}
+
+const handleRedirection = () => {
+  if (targetRoute.value) {
+    router.push(targetRoute.value)
```
fabis94, line 101: Shouldn't targetRoute.value be cleaned up after using it?

packages/frontend-2/lib/projects/helpers/components.ts
```diff
@@ -50,7 +50,7 @@ export const commentPermissionsSelectItems: Record<
   },
   [CommentPermissions.TeamMembersOnly]: {
     id: CommentPermissions.TeamMembersOnly,
-    title: 'Only team members can comment'
+    title: 'Only invited collaborators can comment'
```
fabis94, line 53: ditto

packages/ui-components/src/components/layout/tabs/Horizontal.vue
```diff
@@ -0,0 +1,238 @@
...
+  checkArrowsVisibility()
+}
+
+const scrollRight = () => {
+  scrollContainer.value?.scrollBy({ left: 100, behavior: 'smooth' }) // Adjust the scroll amount as needed
+  checkArrowsVisibility()
+}
+
+const handleScroll = () => {
+  checkArrowsVisibility()
```
fabis94, line 191: Scrolling & window resizing should **always** be throttled or debounced because of the large number of events fired. In this case throttle probably makes the most sense.

packages/ui-components/src/components/layout/tabs/Horizontal.vue
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+onMounted(() => {
+  if (isClient) {
+    if (props.items.length && !activeItem.value) {
+      setActiveItem(props.items[0])
+    }
+    checkArrowsVisibility()
+    nextTick(() => {
```
fabis94, line 211: nextTick/setTimeouts are sketchy and unreliable, i'd evaluate if there's a way to avoid this

packages/ui-components/src/components/layout/tabs/Horizontal.vue
```diff
@@ -0,0 +1,238 @@
...
+      ensureActiveItemVisible()
+    })
+  }
+})
+
+watch(
+  () => [props.items, activeItem.value] as const,
+  ([newItems]) => {
+    if (Array.isArray(newItems) && newItems.length && !activeItem.value) {
+      setActiveItem(newItems[0] as LayoutPageTabItem)
```
fabis94, line 221: unnecessary type cast

packages/ui-components/src/components/layout/tabs/Vertical.vue
```diff
@@ -0,0 +1,107 @@
...
+      setActiveItem(props.items[0])
+    }
+  }
+})
+
+watch(
+  () => [props.items, activeItem.value] as const,
+  ([newItems]) => {
+    if (Array.isArray(newItems) && newItems.length && !activeItem.value) {
+      setActiveItem(newItems[0] as LayoutPageTabItem)
```
fabis94, line 103: unnecessary type cast

fabis94 (changes requested):
didn't test it, but looks good for the most part. some minor questions/issues remain

packages/frontend-2/components/project/VisibilitySelect.vue
```diff
@@ -66,7 +66,7 @@ const items = ref<
   },
   [ProjectVisibility.Private]: {
     id: ProjectVisibility.Private,
-    description: 'Only team members will have access',
+    description: 'Only invited collaborators can access',
```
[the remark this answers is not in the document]
benjaminvo: Yeeah, good point. I just wanted to emphasize that collaborators are "invited". But no need for it so I'll change it.

packages/frontend-2/components/project/page/settings/general/General.vue
fabis94: Also - why the .then() syntax? Just make the fn an async one and use await

packages/frontend-2/components/project/page/settings/general/General.vue
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  const updatePayload: ProjectUpdateInput = {
+    id: project.value.id,
+    ...updates
+  }
+
+  const options = customSuccessMessage ? { customSuccessMessage } : {}
+
+  updateProject(updatePayload, options)
```
fabis94, line 117: 1. This doesn't resolve the comment I left about using .then/.catch syntax over async/await
2. updateProject is never gonna throw, you need to actually check if the result of updateProject is truthy

packages/frontend-2/components/project/page/settings/general/block/ProjectInfo.vue
```diff
@@ -0,0 +1,160 @@
...
+}
+
+const dialogButtons = computed(() => [
+  {
+    text: 'Discard Changes',
+    props: { color: 'secondary', fullWidth: true, outline: true },
+    onClick: () => {
+      showConfirmDialog.value = false
+      resetLocalState()
+      if (targetRoute.value) {
```
fabis94, line 119: 1. Duplicates implementation of handleRedirection
2. 2. Doesn't handle targetRoute cleanup

fabis94 (changes requested):
,

packages/frontend-2/components/project/page/settings/general/General.vue
andrewwallacespeckle: Sorry, I missed the second comment on async. Hopefully this fixes everything, but let me know if I am misunderstanding.

fabis94 (approved):

fabis94:
@andrewwallacespeckle i noticed there's a pre-commit error in the pipeline

PR #2228, opened 2024-04-24
Fix (dui3): refetch projects models

oguzhankoral:


didimitrie (approved):
Leave a comment why we do that refetch and then merge away!

PR #2229, opened 2024-04-25
New URL formats in URLHelper 

AlexandruPopovici:
## Description & motivation
Handles https://linear.app/speckle/issue/WEB-835/enable-new-url-formats-in-the-sandbox

## Changes:
- `URLHelper` moved to `viewer` so it can be optionally used
- `URLHelper` now supports new and old urls
- `object-loader` no longer attempts to read from the `AuthToken` local storage key

## Checklist:


packages/viewer/src/modules/UrlHelper.ts
```diff
@@ -0,0 +1,276 @@
...
+  projectId: string
+}
+
+interface CommitReferencedObjectUrl {
+  origin: string
+  streamId: string
+  commitId: string
+}
+
+export default class UrlHelper {
```
fabis94, line 14: This isn't a problem that necessarily needs fixing, but I don't see the point of this being a class, especially if all of its functions are static. These could just easily be functions directly exported out from the module.

packages/viewer/src/modules/UrlHelper.ts
```diff
@@ -0,0 +1,276 @@
...
+        }
+        `,
+        variables: {
+          projectId: ref.projectId
+        }
+      })
+    })
+    const { data } = await res.json()
+    const urls: string[] = []
+    data.project.models.items.forEach(
```
fabis94, line 267: `data` can easily be undefined if there's an error during the execution of the query. Check for this in all queries.

fabis94 (changes requested):
.

packages/viewer/src/modules/UrlHelper.ts
fabis94: `project` can be null too, if its not found

packages/viewer/src/modules/UrlHelper.ts
```diff
@@ -0,0 +1,310 @@
...
+  const ref: ReferencedObjectUrl = {
+    origin: parsed.origin,
+    projectId
+  }
+
+  const resources = SpeckleViewer.ViewerRoute.parseUrlParameters(
+    decodeURIComponent(params[0])
+  )
+
+  for (let k = 0; k < resources.length; k++) {
```
fabis94, line 130: Instead of awaiting each of those async functions serially, you should add all of the resulting promises to an array and then `await Promise.all()` them, to ensure they're all loaded in parallel

fabis94 (changes requested):
looks ok for the most part, there's 1 remaining performance optimization opportunity I noticed

fabis94 (approved):

PR #2230, opened 2024-04-25
fix: model list view query bug w/ names w/ parentheses

fabis94:


packages/frontend-2/components/project/page/models/StructureItem.vue
```diff
@@ -17,6 +17,9 @@
         <template v-else-if="pendingModel">
```
fabis94, line 17: 1. Added loading bar when children items are loading
2. Fixed up spacings a bit for children items, where they didn't have any padding and looked broken

benjaminvo (approved):

PR #2232, opened 2024-04-29
fix(dui3/ui-components): optinal css fiddles for ancient chrome versions

didimitrie:
## Description & motivation

DUI3 runs sometimes in an ancient chrome version (65), which does not support the following: 
- `inset` classes
- `dvh` units

This results in unscrollable dialogs in aforementioned ancient chrome version. 

## Changes:

Adds a tiny optional off prop to the layout dialog component, that then does the following: 

- `inset-0` becomes `top-0 bottom-0 right-0 left-0` 
- `dvh` units become `vh` units

cc @andrewwallacespeckle as this is something i'd like to merge in main too!

### DIFF
--- a/packages/ui-components/src/components/layout/Dialog.vue
+++ b/packages/ui-components/src/components/layout/Dialog.vue
@@ -11,11 +11,11 @@
         leave-to="opacity-0"
       >
         <div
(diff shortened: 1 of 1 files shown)


andrewwallacespeckle:
Looks good to me, let me know if you are still having issues and I can take a look

PR #2237, opened 2024-05-01
feat(fe1): WEB-734 - dissuade fe1 signups

andrewwallacespeckle:
## Description & motivation
[Linear Ticket
](https://linear.app/speckle/issue/WEB-734/only-allow-account-creation-in-fe2)

We were still getting a lot of fe1 signups, so it was decided to disable this when the env for fe2messaging is true. 

Now, when you go to the login page, and click register, you will be redirected to the new register page. 

If you go directly to the register page, you will be redirected to the new login page.

Also, the login blurb messaging is now based on the env, rather than always showing. 

## To do before merge:
@fabis94 Could you check over the registration token changes here? Just to make sure that this won't break the registration flow when a token is present, it should pass through to the new FE, but just checking this is the same in FE1 and FE2 and I am passing it correctly. This was hard to test.


fabis94 (approved):
Looks OK. It shouldn't be that hard to test tho, not sure what you mean by that 🤔

andrewwallacespeckle:
> Looks OK. It shouldn't be that hard to test tho, not sure what you mean by that 🤔

I can't get the local email server working, so invites don't work for me. I haven't had time to troubleshoot it properly, will hopefully get a look next week.

PR #2242, opened 2024-05-03
feat(fe2): promo banner support

andrewwallacespeckle:
## Description & motivation
Adding a mechanism to manage Promo Banners. This is 2 components - PromoBannersWrapper and PromoBannersBanner.

The wrapper is passed an object containing banner details. It will set a cookie for each banner, defaulting to false. Once that banner has been dismissed, the cookie will update to true, hiding this from the user until the cookie is reset. 

There is a priority field, if multiple banners are available, it will display the one with the highest priority. 

There is a function to not show another promo banner for 24 hours after dismissing the last one. 

## To-do before merge:
- This relies on absolute image references, this may be ok for the first draft to get the Speckleverse promo up, or Fabians may know a better way of handling this.


packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -116,6 +120,17 @@ const debouncedSearch = ref('')
 const openNewProject = ref(false)
 const showLoadingBar = ref(false)
 
+const promoBanners = ref<PromoBanner[]>([
+  {
+    id: 'speckleverse',
+    primaryText: 'Join our online hackathon!',
+    secondaryText: 'June 7 - 9, 2024',
+    url: 'https://speckle.systems/blog/hackathon/',
+    imageSrc: '~~/assets/images/banners/speckleverse.svg',
```
fabis94, line 129: clean up unused properties

packages/frontend-2/components/promo-banners/Banner.vue
```diff
@@ -0,0 +1,56 @@
...
+    </div>
+  </div>
+</template>
+<script setup lang="ts">
+const props = defineProps<{
+  id: string
+  primaryText: string
+  secondaryText?: string
+  url: string
+  imageSrc: string
```
fabis94, line 46: unused

packages/frontend-2/components/promo-banners/Wrapper.vue
```diff
@@ -0,0 +1,61 @@
...
+import { useSynchronizedCookie } from '~/lib/common/composables/reactiveCookie'
+import dayjs from 'dayjs'
+import type { PromoBanner } from '~/lib/promo-banners/types'
+
+const props = defineProps<{
+  banners: PromoBanner[]
+}>()
+
+const hideAllPromoBanners = useSynchronizedCookie('hide-all-promo-banners', {
+  default: () => false,
```
fabis94, line 25: You don't necessarily always have to set defaults for cookies - set them when you actually want them to have a value. In this case it's enough for the cookie to have an undefined value, and only it being created when you set it to `true`

packages/frontend-2/components/promo-banners/Wrapper.vue
```diff
@@ -0,0 +1,61 @@
...
+const activeBannerId = ref<string | null>(null)
+
+const sortedBanners = computed(() =>
+  [...props.banners].sort((a, b) => a.priority - b.priority)
+)
+
+// Set active banner initially based on cookies
+sortedBanners.value.forEach((banner) => {
+  const cookie = useSynchronizedCookie(`banner-dismissed-${banner.id}`, {
+    default: () => false
```
fabis94, line 38: unnecessary default value, just creates an unnecessary cookie

packages/frontend-2/components/promo-banners/Wrapper.vue
```diff
@@ -0,0 +1,61 @@
...
+)
+
+function handleDismissed(id: string) {
+  hideAllPromoBanners.value = true
+
+  // Hide the dismissed banner
+  activeBannerId.value = null
+
+  // Set dismissed cookie
+  const dismissedCookie = useSynchronizedCookie<boolean>(`banner-dismissed-${id}`)
```
fabis94, line 58: composable invoked outside of acceptable scope

fabis94 (changes requested):
@benjaminvo @andrewwallacespeckle 
If it's really important we can launch this already, but keep in mind that there's no expiration mechanism in the banner so for self-hosters if they don't update they'll stay around even after the hackathon is already over

benjaminvo:
> @benjaminvo @andrewwallacespeckle If it's really important we can launch this already, but keep in mind that there's no expiration mechanism in the banner so for self-hosters if they don't update they'll stay around even after the hackathon is already over

That's a good point. @andrewwallacespeckle I think we should add an expiration date to these and set it to when the hackathon starts.

andrewwallacespeckle:
Good point Fabs, I've added the expiry date mechanism.

fabis94 (approved):

PR #2246, opened 2024-05-07
bugfix(fe2): Fix conditioning around posting comments in viewer

andrewwallacespeckle:
## Description & motivation
There was a bug in the viewer where a user who was logged in, was still presented the login screen when trying to post a comment. This occurred when the user was on a public project, that they were not a collaborator in. 

I adjusted the conditions around canPostComment, so it fits better in every scenario. I also hid the 2 "New Comment" buttons when commenting was not possible, but the user was logged in.

### DIFF
--- a/packages/frontend-2/components/viewer/anchored-point/Thread.vue
+++ b/packages/frontend-2/components/viewer/anchored-point/Thread.vue
@@ -162,7 +162,7 @@
               </FormButton>
             </div>
             <div
-              v-if="!canReply && !isEmbedEnabled"
+              v-if="!canReply && !isEmbedEnabled && !isLoggedIn"
               class="p-3 flex flex-col items-center justify-center bg-foundation-2"
             >
               <FormButton full-width @click="$emit('login')">Reply</FormButton>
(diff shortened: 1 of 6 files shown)


packages/frontend-2/components/viewer/anchored-point/Thread.vue
```diff
@@ -162,7 +162,7 @@
               </FormButton>
             </div>
             <div
-              v-if="!canReply && !isEmbedEnabled"
+              v-if="!canReply && !isEmbedEnabled && !isLoggedIn"
```
fabis94, line 165: This bit i'm not sure about - what if the user is logged in, and yet still can't reply due to not having the rights to do so? Do we still show something to him?

fabis94 (approved):

packages/frontend-2/components/viewer/anchored-point/Thread.vue
andrewwallacespeckle: This will hide the "+" in the viewer for these users. So they are logged in, but can't comment, they won't see the plus.

PR #2248, opened 2024-05-08
Fix (dui3): close dialogs on doc switch

oguzhankoral:
This is annoying when user switch doc and dialogs stays on open state. We should close all of them. Not sure we can do it with another way by using some callback function from Dialog component.


didimitrie (changes requested):
I'm hesitant to

didimitrie (approved):

PR #2250, opened 2024-05-09
bugfix(ui-components): Scroll jumps to top on Dialog open

andrewwallacespeckle:


benjaminvo:
@andrewwallacespeckle I tested locally. Still happens for me 🤔 I tested the invite and delete project dialogs.

I'm on mac chome 124

andrewwallacespeckle:
> @andrewwallacespeckle I tested locally. Still happens for me 🤔 I tested the invite and delete project dialogs.
> 
> I'm on mac chome 124

I'm on the same version on mac, and it's working. 

Can you make sure you run `yarn build:public`? 

If that doesn't work, can you open a dialog and check in inspector. It should add a `dialog-open` class to the HTML tag. Is this happening?

benjaminvo:
@andrewwallacespeckle Ah, that was it. Works great now, will review.

benjaminvo (approved):
Works great! A bit annoying that we need to overwrite headlessui with these CSS classes but that's how it is.

PR #2253, opened 2024-05-13
feat(fe2) - Replace CommonEditableTitleDescription with new CommonTitleDescription

andrewwallacespeckle:
## Description & motivation
Currently, the user can edit the Project/Model Title and Description from the Project or Model page. This has been raised as an issue, as normally the project title is very important, and integrated heavily in workflows. Changing it should be a task that is thought out, not a quick change from the project page. 

Now, users cannot edit the title or description from anywhere other than the project settings page.  

## Changes:
- Add CommonTitleDescription
- Update 2 uses of CommonEditableTitleDescription in Project and Model page headers


packages/frontend-2/components/common/TitleDescription.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <div class="flex flex-col gap-2 sm:pr-8">
```
benjaminvo, line 2: I think we can get rid of this right padding now that the edit icon is gone?

benjaminvo:
Tested locally and works great No other comment than the one above

packages/frontend-2/components/common/TitleDescription.vue
andrewwallacespeckle: I added this padding to increase the spacing between the team panel and the title/description.

I can see how this was unclear, so I have removed the padding and instead increased the gap in the parent item, by removing the responsive helper `sm:gap-4`

benjaminvo (approved):

benjaminvo (changes requested):
I just thought of something: Let's write "No description" in italic if no description. It feels a bit empty without it and it might prompt users to add it.

CleanShot 2024-05-15 at 10 22 13@2x

benjaminvo (approved):
Looks good!

PR #2255, opened 2024-05-14
bugfix(fe2): Coloring function disappears if parameter title is too long

andrewwallacespeckle:
## Description & motivation
If there was a longer filter name, the colouring function is pushed off the filter menu, and becomes unusable. 

## Changes:
- Remove -mt-1 which wasn't needed
- Add truncate to the button label, and a max width for mobile and desktop

### DIFF
--- a/packages/frontend-2/components/viewer/explorer/Filters.vue
+++ b/packages/frontend-2/components/viewer/explorer/Filters.vue
@@ -2,19 +2,19 @@
   <ViewerLayoutPanel class="mt-2" hide-close>
     <template #title>Filtering</template>
     <template #actions>
-      <div class="flex justify-between items-center w-full">
-        <div class="-mt-1">
-          <FormButton
-            v-tippy="'Change Filter'"
-            text
-            size="xs"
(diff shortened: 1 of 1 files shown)


packages/frontend-2/components/viewer/explorer/Filters.vue
```diff
@@ -2,19 +2,19 @@
...
-            class="capitalize"
-            @click="showAllFilters = !showAllFilters"
-          >
+      <div class="flex justify-between items-center w-full gap-2">
+        <FormButton
+          v-tippy="'Change Filter'"
+          text
+          size="xs"
+          :icon-right="showAllFilters ? ChevronUpIcon : ChevronDownIcon"
+          class="capitalize"
```
benjaminvo, line 11: I think you also need a `truncate` class on the button. The reset + coloring buttons do a small jump to the right when the button text is too long. Gif:
CleanShot 2024-05-14 at 13 25 45

benjaminvo (changes requested):

benjaminvo (approved):

PR #2256, opened 2024-05-14
feat(fe2): Use shortened header in Scene Explorer

andrewwallacespeckle:
## Description & motivation
In scene explorer, we show the full model name, including nestings. This means the actual model name is often cut off. 

We show the shortened model name in the Models panel. We should shorten it in the same was in scene explorer. 

## Changes:
- Add getNestedModelHeader function
- User getNestedModelHeader with result of getHeaderAndSubheaderForSpeckleObject


benjaminvo (approved):
Nice. Tested locally. And I checked that it's the same method used in the Model tab.

PR #2257, opened 2024-05-14
fix(fe2): Uncomment code around model duplication

andrewwallacespeckle:
### Problem:
Encountered an issue with the viewer where multiple resources for the same model were being loaded, causing duplication and potential conflicts.

### Solution:
Found the problem in the following block of code:
`// In case we want to go back to 1 resource per model:
// if (modelId && encounteredModels.has(modelId)) continue
if (encounteredObjects.has(objectId)) continue`

Uncommented the line to ensure only one resource per model is loaded:
`if (modelId && encounteredModels.has(modelId)) continue`

### Question:
Does anyone see any reason why this line was originally commented out? Is there any specific use case I'm missing, or is it safe to keep this uncommented?


benjaminvo:
Has to be a question for @didimitrie

didimitrie:
It might be diffing, where we need to load two versions of the same model at the same time

andrewwallacespeckle:
> It might be diffing, where we need to load two versions of the same model at the same time

I thought this, but I tested diffing and it still works. It uses this format of url:
http://localhost:8081/projects/fd8cab6317/models/6f3c730d7f@d57b0cd578,all#diff={%22diffs%22:[{%22versionA%22:%226f3c730d7f@d57b0cd578%22,%22versionB%22:%226f3c730d7f@a991fed98e%22}]}

Where the diff versions are separate to loaded versions. The change shouldn't affect this. 

To play it safe we wait a week and ask Fabians his intention in commenting this out, or merge and make a note to discuss with him on return.

andrewwallacespeckle:
@fabis94 Thanks for the call on this one. I've updated the comment around this based on our call.

fabis94 (approved):

PR #2260, opened 2024-05-15
fix(fe2): "Manage" Button causes authentication error for user without edit permissions

andrewwallacespeckle:
## Description & motivation
It was noticed that when a project contributor clicks the "Manage" button in the team block, they are taken to a Contributor page with no contributors, and an authentication error in console. A reload fixed this, suggesting it is not an authentication problem, but more likely to be some race condition. 

I have spent a bit of time troubleshooting this, but have not found the underlying issue. Some input from Fabians would be good here. 

In the meantime, I have changed the conditioning around the manage button, so only users with edit permissions can see the Manage button. Users without this permission can still go to settings>contributors, but they won't see the Manage button from the project page. 

## Changes:
- Change condition on "Manage" button from "activeUser" to "canEditProject".

### DIFF
--- a/packages/frontend-2/components/project/page/team/Block.vue
+++ b/packages/frontend-2/components/project/page/team/Block.vue
@@ -15,7 +15,7 @@
     <template #bottom>
       <div class="flex items-center justify-between mt-1">
         <UserAvatarGroup :users="teamUsers" class="max-w-[104px]" />
-        <div v-if="activeUser">
+        <div v-if="canEdit">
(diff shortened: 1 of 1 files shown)


benjaminvo (approved):
Looks good. I couldn't test the contributor state locally because I'm an Owner on all projects, but the code looks good.

Let's make a ticket to revisit this so we can show the Manage button to Contributiors too. But this fix is fine short term!

PR #2261, opened 2024-05-15
feat(fe2): Hide settings tab for logged out users

andrewwallacespeckle:
## Description & motivation
Settings tab should be hidden for users not logged in. We now only push settings to the PageTabItems when the user is logged in. 

We also had to add a new middleware called "can-view-settings". This will check that the user is logged in, if not, it will redirect them to the project route.

Fabians had raised a similar problem in webhooks. So I made a new middleware for this and applied it similarly. If a user who is logged out, or does not have edit permissions, they will be redirected to the project page when they try to go to the webhooks url. 

## Changes:
- Only add settings to PageTabItems when logged in. 
- Add can-view-settings middleware
- Add can-view-webhooks middleware


benjaminvo (changes requested):
Two things:
1. Actually the Settings tab should be hidden to all non collaborators. So there's also a logged in case where you don't want to show it to users. Sorry, didn't think of this when writing the issue.
2. Clicking on Webhooks or visiting `settings/webhooks` directly sends me to the Models tab.

benjaminvo (approved):

PR #2264, opened 2024-05-16
fix(fe2): trim spaces around slashes in model names

andrewwallacespeckle:
## Description & motivation
There was a bug where having a space before or after the "/" would cause moving/creating models to fail. It created a folder for the model, but the folder had no models. 

## Changes:
- Added new helper function called sanitizeModelName
- Anywhere the user can edit the Model Name (or add a new one), this function is called to remove any spaces around the slash


benjaminvo (approved):
Very nice

PR #2268, opened 2024-05-17
fix(fe2): comment button persists after deselecting

andrewwallacespeckle:
## Description & motivation
The comment button doesn't disappear after deselect an object either by pressing esc or closing the Selection Info panel.

## Changes:
- Check that there are objects selected before showing New Comment button.


packages/frontend-2/components/viewer/anchored-point/NewThread.vue
```diff
@@ -72,6 +72,7 @@
       </ViewerCommentsPortalOrDiv>
     </div>
   </div>
+  <div v-else></div>
```
benjaminvo, line 75: Why is this needed?

andrewwallacespeckle: If shouldShowThreadBubble returns false, the component will still be mounted, but there will be nothing inside the , this can cause side effects, and has done elsewhere in the app. Now, when I see this, I add an empty div so something will still be rendered and avoid those side effects. There was no side effect that I noticed here, but it's just good practice.

PR #2269, opened 2024-05-17
Alessandro/web 893 refactor proposal for the automate module

alemagio:
### Repository

In general there are 2 approaches we could use to implement repositories:
1. repositories as data access objects, so we would have one repository for every table in the DB
2. repositories only for aggregates (entities we use as top-level in the sense of business logic)

In this PR I decided to use approach 1.

- In this PR, we assume Automation as an aggregate, so all the operations should go through it.
- The `automate` module only has one repository, which is the `AutomationRepository`
- the `AutomationRepository` receives as a dependency a `db` object, which is the object used to access the database 
- the `db` object has a field for every entity the `AutomationRepository` has access to

### Missing points
-  we should decide how to implement `dataloaders`

## Checklist:


packages/server/modules/automate/graph/resolvers/automate.ts
```diff
@@ -451,11 +452,20 @@ export = {
       })
     },
     async trigger(parent, { automationId }, ctx) {
+      const automationRepository = createAutomationRepository({
+        db: {
+          AutomationTriggers: AutomationTriggers.knex,
```
gjedlicska, line 457: I think these `.knex` references are not working well enough now.

alemagio: I did what I needed to avoid changing all the db setup we already have in place
as long as this is scoped in a single place I think we can deal with it

packages/server/modules/shared/authz.ts
```diff
@@ -321,22 +330,37 @@ export const authPipelineCreator = (
   return pipeline
 }
 
-export const streamWritePermissions: AuthPipelineFunction[] = [
+export const streamWritePermissions: ({
+  db
+}: {
+  db: Knex
+}) => AuthPipelineFunction[] = ({ db }: { db: Knex }) => [
```
gjedlicska, line 337: why do we need the `{ db } ` argument in the created function?

gjedlicska:
@alemagio prob you meant to close this one :P

alemagio:
@gjedlicska This is why I shouldn't do stuff while in a meeting

PR #2278, opened 2024-05-22
feat(fe2): change submit logic for access settings before embedding

andrewwallacespeckle:
## Description & motivation
Previously, when you tried to embed a Private project, you would be greeted with a Dialog warning that this was not possible, and a select which allowed you to change the project access (if you are an owner). 

This updated instantly, so as soon as you selected "Public", the project was updated. 

This has been changed, and a save button introduced. Now, when you change Status the Save button will be enabled, allowing the user to save.


benjaminvo:
Everything works fine! Some design feedback: Let's only show the copy "To change this setting [...]" if you're not the owner + let's output it below the select input which I assume will be disabled.

I don't have a private project where I'm not an owner so I can't test this.

CleanShot 2024-05-23 at 13 08 41@2x

benjaminvo (changes requested):
Posted feedback in separate comment

benjaminvo:
@andrewwallacespeckle Maybe I wasn't clear.

I want the blue alert to always say "Model embedding does not work [...]". And then under the select input write the text "To change this setting [...]" if user is not Owner. And btw that copy can be shortened to "You must be an Owner of this project to change this setting".

andrewwallacespeckle:
Sorry, I misread your comment. 

Now, when you are not an owner:
image

And when you are an owner:
image

benjaminvo (approved):
nice!

PR #2279, opened 2024-05-22
feat: ensure consistent ordering of nodes to match ViewerLayoutPanel

andrewwallacespeckle:
## Description & motivation
Currently, in the viewer, the Model panel has a different ordering to the Scene Explorer. This task aims to correct this by ordering Scene Explorer as Model panel currently is:
Model w/ version > Model > Folder name > Object ID

## Changes
Correctly classify and order nodes into versionNodes, modelNodes, folderNodes, allModelNodes, and objectNodes


andrewwallacespeckle:
> @andrewwallacespeckle I don't completely understand this sorting: "Model w/ version > Model > Folder name > Object ID". Can you explain?

I'm not entirely sure why this was order was chosen, but it is very intentional. There is even a comment about it:
` /**
     * Flatten results into an array of items that are properly ordered according to resource identifier priority.
     * Loading priority: Model w/ version > Model > Folder name > Object ID
* /`

We can easily update this, but I'd assume that order was chosen for a reason. This PR uses the same ordering method in Scene Explorer to replicate Models order.

packages/frontend-2/components/viewer/explorer/Explorer.vue
```diff
@@ -89,7 +90,14 @@ const rootNodes = computed(() => {
 
   if (!worldTree.value) return []
   expandLevel.value = -1
-  const nodes = []
+
+  // Arrays to hold categorized nodes
+  const versionNodes = []
```
fabis94, line 95: @andrewwallacespeckle I'm wondering if there's a way to avoid duplicating this sorting logic. Since `resourceItems` is already sorted correctly, maybe we can just colocate each node of rootNodes with its resourceItem and use its idx in that array to set up the order?

```ts
const results: Record<number, ExplorerNode> = {}
for (const node of rootNodes) {
  const objectId = ...
  const order = resourceItems.value.findIndex(ri => ri.objectId === objectId)
  results[order] = node
}

return sortBy(Object.values(results), i => i[0]).map(i => i[1])
```

fabis94 (changes requested):

andrewwallacespeckle:
Latest changes made on pairing call with @fabis94

fabis94 (approved):

fabis94 (approved):

PR #2280, opened 2024-05-22
fix(fe2): Use better input names to avoid triggering autocomplete. Stop "a few seconds ago" from breaking line

andrewwallacespeckle:
## Description & motivation
2 Small changes to fe2. 

The first was reported by Matteo in discord. Because we call various inputs "name", autocomplete wrongly thinks these are the user’s name, rather than model/project/function etc. I updated these to be more accurate.

The second is something I noticed when working on another task. If you push a version through a connector, quite often you will see "A few seconds ago" as the time in the viewer. This copy is too long, and breaks onto 2 lines and looks poor. 

## Changes:
- Add "autocomplete='off'" to inputs with their name as name, even though it is not the users name. Autocomplete checks the name attribute, and ours cannot be easily changed without breaking changes. Instead, we will just turn off autocomplete for these inputs. 
- Stop "a few seconds ago" from breaking line

### DIFF
--- a/packages/frontend-2/components/automate/function/create-dialog/DetailsStep.vue
+++ b/packages/frontend-2/components/automate/function/create-dialog/DetailsStep.vue
@@ -12,7 +12,7 @@
       <div class="sm:w-7/12">
         <FormTextInput
           size="lg"
-          name="name"
+          name="functionName"
           label="Name"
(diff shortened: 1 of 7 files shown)


benjaminvo (approved):
Nice!

PR #2282, opened 2024-05-23
feat(fe2): Cancel ongoing measurement with esc

andrewwallacespeckle:
## Description & motivation
If a measurement has started by the user clicking the first point, we know listen for escape. If the user hits escape, it will cancel the ongoing measurement, without disabling Measure mode entirely.


benjaminvo (changes requested):
You can no longer exit the Measure tool with the esc button. While you're not measuring.

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -379,11 +379,23 @@ export function useMeasurementUtilities() {
     state.viewer.instance.getExtension(MeasurementsExtension).clearMeasurements()
   }
 
+  const activeMeasurement = computed(() => {
+    const measurementsExtension =
```
fabis94, line 383: I don't think this is actually reactive, cause the viewer certainly isn't. Does it really update in real-time? You can validate this by setting up a watcher on this that emits the value in the console.

If it isn't reactive, better to be clear about that and expose this as a function instead (getActiveMeasurement(): boolean)

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -391,6 +391,8 @@ watch(isSmallerOrEqualSm, (newVal) => {
 })
 
 onKeyStroke('Escape', () => {
-  activeControl.value = 'none'
+  if (activeControl.value !== 'measurements') {
```
fabis94, line 394: Not ideal that the logic for what should happen on Esc is split apart in multiple files (this and postSetup.ts), but it can be left as is for now

fabis94 (dismissed):
I asked 1 question, but outside of that everything looks ok

packages/frontend-2/components/viewer/Controls.vue
andrewwallacespeckle: Now in 1 file - Controls.vue

fabis94 (approved):

andrewwallacespeckle:
@benjaminvo For some reason, I can't merge this until you have approved the changes, even though Fabs already has. Can you re-approve and I'll merge.

benjaminvo (approved):

benjaminvo:
@andrewwallacespeckle There you go!

PR #2292, opened 2024-05-28
feat(server): exec engine API GET call retry + better handling of network outages

fabis94:
## Checklist:


gjedlicska:
Why are we not using something like https://www.npmjs.com/package/fetch-retry ?

gjedlicska (commented):
Why are we not using something like https://www.npmjs.com/package/fetch-retry ?

fabis94:
> Why are we not using something like https://www.npmjs.com/package/fetch-retry ?

@gjedlicska 1) It does more or less the same thing, so I can achieve the same without adding a dependency 2) It seems to patch fetch globally, which I want to avoid

gjedlicska (approved):

PR #2294, opened 2024-05-28
fix: extra safety for when automate is disabled

fabis94:
tl;dr; Even if a module is disabled, its GQL components (typedefs, resolvers) will still be loaded to avoid complexity on the client-side. So we must always ensure resolvers can handle the module being disabled.


gjedlicska (changes requested):
We cannot have the resolvers working. They give access to the API. So if a server has the module disabled, but the resolvers still work, someone can just use the graphql API to utilize the functionality of the disabled module.

gjedlicska (approved):

PR #2298, opened 2024-05-29
fix(fe2): View menu not outputting correct names

andrewwallacespeckle:
## Description & motivation
It was reported by Nikos that Views were not named correctly in the View menu. They were outputing the ID of the view, rather than the name. 

We were outputting `view.name` which should have been `view.view.name`.

I have tested this with a Rhino model, and a Revit model. 

## Changes:
- Change `view.name` to `view.view.name` to match the correct shape of views
- Stop close on selection. Had to remove headlessui


packages/frontend-2/components/viewer/views/Menu.vue
```diff
@@ -78,7 +64,11 @@ const {
 } = useInjectedViewerState()
 const { setView: setViewRaw } = useCameraUtilities()
 const mp = useMixpanel()
-const menuButtonId = useId()
+
+const open = ref(false)
+const active = ref(false)
```
fabis94, line 69: Doesn't appear that this value ever gets mutated

packages/frontend-2/components/viewer/views/Menu.vue
```diff
@@ -1,10 +1,8 @@
 <template>
-  <Menu as="div" class="relative z-30">
-    <MenuButton :id="menuButtonId" v-slot="{ open }" as="template">
-      <ViewerControlsButtonToggle flat secondary :active="open">
-        <IconViews class="w-5 h-5" />
-      </ViewerControlsButtonToggle>
-    </MenuButton>
+  <div ref="menuWrapper" class="relative z-30">
```
fabis94, line 2: Instead of removing Menu, you could've just used its API to manually manage its open/closed state. The Menu still has accessability foundations baked in, which I'm not sure if you've re-implemented. See MenuItems static prop

fabis94 (changes requested):

packages/frontend-2/components/viewer/views/Menu.vue
andrewwallacespeckle: I did try using the static prop with MenuItems, but it still required wrapping them in a v-show, which caused the open state to toggle when a MenuItem was clicked. They do provide a close() slot prop, but there's no way to keep the menu open on selection.

I also experimented with managing a second state to control the visibility of MenuItems using v-show, but it ended up being a bit of a mess. The accessibility features of Headless UI are tightly integrated with their internal open state. Trying to override this with our own state just didn't work well.

It felt like I was fighting against the library rather than working with it.

Maybe there is something I'm missing that is an easy solution?

fabis94: Nah, if it was a big PITA then skip it. Do ensure its still keyboard-navigatable tho

packages/viewer/src/IViewer.ts
```diff
@@ -91,7 +91,7 @@ export interface ViewerEventPayload {
 export type SpeckleView = {
   name: string
   id: string
-  view: { origin: Vector3Like; target: Vector3Like }
+  view: { name: string; origin: Vector3Like; target: Vector3Like }
```
fabis94, line 94: @andrewwallacespeckle Confirm w/ @AlexandruPopovici. Even if there is such a prop, it might be optional or nullable

fabis94 (changes requested):

packages/viewer/src/IViewer.ts
AlexandruPopovici: Seems like it does have a `name` https://github.com/specklesystems/speckle-sharp/blob/b2f9f49cde910b080a5aa118a66c3a40452f8693/Objects/Objects/BuiltElements/View.cs#L11

andrewwallacespeckle: Thanks for checking @AlexandruPopovici. Can you let me know when you have had a look? It would be good to get this out soon as views are currently broken without this.

packages/viewer/src/IViewer.ts
```diff
@@ -88,10 +88,12 @@ export interface ViewerEventPayload {
   [ViewerEvent.LightConfigUpdated]: LightConfiguration
 }
 
-export type SpeckleView = {
-  name: string
-  id: string
-  view: { name: string; origin: Vector3Like; target: Vector3Like }
+export type SpeckleView = SpeckleObject & {
```
AlexandruPopovici, line 91: @andrewwallacespeckle 
The previous `SpeckleView` structure was legacy stuff. A `SpeckleView` is just another `SpeckleObject` with extra stuff, some of which may not be filled out by connectors so I made them optional. 
For the `name` prop, there is a very good change that it will always exist, however I am not 100% sure so I put it down as optional. I've asked connectors and maybe we will get a definite answer on this. If not, I guess you could also fallback to the displaying the `id` which is mandatory to exist

AlexandruPopovici:
@andrewwallacespeckle 
I've changed the `SpeckleView` model so that there is no more unneeded legacy stuff. I've updated the viewer library to use the new model, but I haven't done so to the frontend, that's why the CI is failing.

andrewwallacespeckle:
> @andrewwallacespeckle I've changed the `SpeckleView` model so that there is no more unneeded legacy stuff. I've updated the viewer library to use the new model, but I haven't done so to the frontend, that's why the CI is failing.

I can sort the frontend, thanks for checking this for me, Alex!

fabis94 (approved):

PR #2302, opened 2024-05-30
WEB-1052 Create test automations

cdriesler:
## Description & motivation

Ok I think we have it.

We'd like to create "test automations" that enable special pathways that enable testing an automate function locally. This requires:

- The ability to create these automations
- A separate path on the frontend to do so
- A new, simplified, endpoint for creating these

## Changes:

This PR gets as far as "make the test automation" from the frontend. They are not yet presented any differently, but the creation flow looks different:

Screen Shot 2024-05-30 at 12 19 06

## Notes:

Seeing this new pathway requires setting `NUXT_PUBLIC_FF_TEST_AUTOMATIONS_ENABLED='true'` in frontend-2 .env


gjedlicska (approved):
lets go

PR #2316, opened 2024-06-03
feat: authenticate user as function author

fabis94:


fabis94:
https://linear.app/speckle/issue/WEB-1038/gh-auth-link-redirect-to-the-automate-gh-auth-flow-url

PR #2322, opened 2024-06-04
feat(fe2): Highlight objects on hover

andrewwallacespeckle:
## Description & motivation
Highlight models and objects when hovering them in the Models tab, Scene Explorer tab, and Selection Info panel.

## Changes:
- Add Hover Highlight to Models in Model Panel
- Add Hover Highlight to Objects in Scene Explorer
- Add Hover Highlight to Objects in Selection Info


packages/frontend-2/components/viewer/explorer/StringFilterItem.vue
```diff
@@ -122,6 +128,18 @@ const color = computed(() => {
     ?.color
 })
 
+const highlightObject = () => {
```
fabis94, line 131: A bit too low level for a Vue component IMO, these should be moved to a `useHighlightedObjectsUtilities` composable in `lib/viewer/composables/ui.ts`, so that you can easily highlight/un-highlight objects without having to do low level array manipulation. Not only that, but you're duplicating these same array manipulations in each Vue component.

Additionally, there could be a shit ton of IDs in these arrays, so we have to be careful with how we do the highlighting/un-highlighting and we have to ensure its performant. So test all of this highlighting out w/ large models where you're potentially highlighting large amounts of IDs at the same time.

fabis94 (changes requested):

packages/frontend-2/components/viewer/explorer/StringFilterItem.vue
andrewwallacespeckle: I tested the highlighting with a large Rhino file on my computer. There's a slight slowdown, but the highlighting itself doesn’t add much to the overall lag. The viewer is generally less responsive due to the model size, but the highlighting function works well even with many IDs.

benjaminvo:
1. There's a bug where selections are stuck. Elements keep being selected even though they aren't. [Loom here](https://www.loom.com/share/ac351e7f1c3148e0bbd981fcbbedb53f).
2. I wonder if we should trigger the outline instead of coloring the object? Would like to try that if it's easy
3. I don't think we should trigger the hover effect on elements that are already selected. So this includes also removing the effect from the selection info panel for now. I'd like to come up with a hover effect that will also work well if it's a selected element, but I don't have any ideas right now.
4. Let's remove all hover transitions to make it more snappy.
5. I am a bit worried about the performance 🤔 I am experience a bit of a lag when quickly hovering a lot of elements,

andrewwallacespeckle:
@fabis94 @AlexandruPopovici I have removed the function from Filtering. In each use case, I logged the object, confirming that only a single object ID is ever passed. I did some profiling and found that highlighting a complex object takes approximately 150ms. I'm not certain what the threshold for being considered too heavy is in this context, but I'd appreciate any insights or recommendations.

andrewwallacespeckle:
@fabis94 This should be good to go for a final review. Thanks

fabis94 (approved):

PR #2327, opened 2024-06-05
feat(fe2): disable New Automation CTA if no models available

fabis94:
Plus some extra fixes:
* Model count being off by 1 in Models tab header (might also fix WEB-657)
* App crashing if models/versions used by automations are deleted


packages/server/assets/automate/typedefs/automate.graphql
```diff
@@ -24,7 +24,7 @@ type AutomationRevisionFunction {
 
 type VersionCreatedTriggerDefinition {
   type: AutomateRunTriggerType!
-  model: Model!
+  model: Model
```
gjedlicska, line 27: Why do we need to change these to nullable?

fabis94: @gjedlicska Cause you can delete them

gjedlicska (approved):

PR #2337, opened 2024-06-06
feat(fe2): Allow user to decide viewer zoom on addition of new model

andrewwallacespeckle:
## Description & motivation
A user requested that the camera does not move when he adds new models. Although he wants this, it may be annoying for other users. 

## Changes:
- Add a Toast when the user successfully adds a new Model
- Disable the zoom when a new model is added
- Add a CTA in the Toast, giving the user the option to zoom


andrewwallacespeckle:
@benjaminvo I've added a new toggle in the "New Model" dialog in the viewer. Are you happy with the design of this or are there any changes you would like me to make? Once you have signed off design, I will submit this PR to Fabians for CR.

benjaminvo (approved):
I like it!

fabis94:
@andrewwallacespeckle @benjaminvo shouldn't this be a setting on the viewer page? it doesn't really change anything on the project page

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -139,7 +139,7 @@ function useViewerObjectAutoLoading() {
 
       await Promise.all(removableObjectIds.map((i) => loadObject(i, true)))
       await Promise.all(
-        addableObjectIds.map((i) => loadObject(i, false, { zoomToObject }))
+        addableObjectIds.map((i) => loadObject(i, false, { zoomToObject: false }))
```
fabis94, line 142: This PR seems incomplete. Where's the vue template changes? Where's the cookie? Where's the part where you read the cookie and use it in this fn call?

fabis94 (changes requested):

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
andrewwallacespeckle: Sorry Fabs, the brief for this one changed, I should have updated the ticket with the new brief. I've updated the PR description now. 

It was decided to drop the Toggle, and instead add a Call to action to the new Toast that appears on a new model being added. 

We also don't zoom when a new model is added to viewer now, only when the user clicks the call to action.

andrewwallacespeckle: Uploading Screen Recording 2024-06-07 at 11.38.09.mov…

Here's a quick video to save you loading this up

fabis94 (approved):

PR #2351, opened 2024-06-10
Adjust table so run details button is visible

benjaminvo:
The button to view the run details isn't visible on the lg breakpoint on the automation page. I've decreased the table's general min width to fix this, but we need to follow up soon with a better responsive table behaviour. I've checked other tables in the product after this change and it doesn't introduce any problems.

### Before
CleanShot 2024-06-10 at 12 52 44@2x

### After
CleanShot 2024-06-10 at 12 48 16@2x

### DIFF
--- a/packages/ui-components/src/components/layout/Table.vue
+++ b/packages/ui-components/src/components/layout/Table.vue
@@ -5,7 +5,7 @@
     >
       <div
         v-if="items.length > 0"
-        class="grid z-10 grid-cols-12 items-center gap-6 font-semibold bg-foundation-page rounded-t-lg w-full border-b border-outline-3 pb-2 pt-4 px-4 min-w-[900px]"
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/layout/Table.vue
```diff
@@ -118,7 +118,7 @@ const paddingRightStyle = computed(() => {
 
 const rowsWrapperClasses = computed(() => {
   const classParts = [
-    'relative grid grid-cols-12 items-center gap-6 px-4 py-1 min-w-[900px] bg-foundation'
```
fabis94, line 121: @benjaminvo  This changes the base Table element, meaning all tables everywhere will have this change. Have you validated that this reduced width doesn't break any Table layouts?

benjaminvo: Yep, from the initial PR description:

> I've checked other tables in the product after this change and it doesn't introduce any problems.

fabis94 (approved):

PR #2362, opened 2024-06-12
feat: optionally add userId to automate backend tracking

gjedlicska:
## Checklist:

### DIFF
--- a/packages/server/modules/automate/services/tracking.ts
+++ b/packages/server/modules/automate/services/tracking.ts
@@ -10,8 +10,11 @@ import {
 } from '@/modules/automate/helpers/types'
 import {
   InsertableAutomationRun,
-  getFullAutomationRevisionMetadata
+  getFullAutomationRevisionMetadata,
+  getFullAutomationRunById
 } from '@/modules/automate/repositories/automations'
+import { getCommit } from '@/modules/core/repositories/commits'
+import { getUserById } from '@/modules/core/services/users'
 import { mixpanel } from '@/modules/shared/utils/mixpanel'
 import { throwUncoveredError } from '@speckle/shared'
 import dayjs from 'dayjs'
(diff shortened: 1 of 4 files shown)


packages/server/modules/automate/services/tracking.ts
```diff
@@ -73,43 +87,77 @@ const onAutomationRunStatusUpdated =
...
-      const mp = mixpanel({ userEmail: undefined })
-      await mp.track('Automation Run Triggered', {
-        automationId: automation.id,
-        automationName: automation.name,
-        automationRunId: automationRun.id,
-        projectId: automation.projectId,
-        source
-      })
-      break
+const getUserEmailFromAutomationRun =
```
fabis94, line 90: @gjedlicska Tbh I hate that we're doing this haha - are we basically gonna have to handle all ways a run could be created (which involves all kinds of DB queries, even re-pulling the same run again just to get extra metadata) in this tracking service just to get a user email?

At that point, if it's such an absolute necessity for Benjamin to get this tracking info, I'd rather just have a `triggeredById` column which is a lot easier to fill on run creation, cause you already have all of the relevant metadata at hand.

packages/server/modules/automate/tests/trigger.spec.ts
```diff
@@ -1112,6 +1113,16 @@ const { FF_AUTOMATE_MODULE_ENABLED } = getFeatureFlags()
...
+        const testVersion = {
+          id: cryptoRandomString({ length: 10 }),
+          authorId: testUser.id,
+          streamId: testUserStream.id,
+          branchName: testUserStreamModel.name,
+          objectId: null
+        }
+
+        // @ts-expect-error force setting the objectId to null
+        await createTestCommit(testVersion)
```
fabis94, line 1125: better to leave the objectId an empty string as instructed in the type, than to override a TS error

fabis94 (changes requested):

packages/server/modules/automate/services/tracking.ts
gjedlicska: I am not fond of it either, but this value def doesn't belong to our object model. On a conceptual basis, all of us still agrees, that a run cannot be 1-1 matched to a user, for all the reasons we've discussed.

Yet, for tracking reasons, we'd want to do this.
I do not want to add this value to our database schema, so we have to pay this price.

fabis94 (approved):

PR #2366, opened 2024-06-13
feat: sourcemaps + testing

fabis94:
generate source maps in prod build + allow devs to trigger errors (for now) through secret query param

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -312,6 +312,16 @@ workflows:
           requires:
             - test-viewer
 
+      - frontend-2-sourcemaps:
+          context:
+            - datadog-sourcemaps-publish
+          filters: *filters-publish
+          requires:
+            - get-version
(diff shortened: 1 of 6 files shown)


fabis94:
Example sourcemap upload command:
```
// CWD ./packages/frontend-2
DATADOG_SITE=datadoghq.eu DATADOG_API_KEY=XXXX yarn datadog-ci sourcemaps upload ./.output/public/_nuxt \
--service=fe2-dev/test \
--release-version=local-1.1.1 \
--minified-path-prefix=http://localhost:8081/_nuxt \
```

fabis94:
@iainsproat --minified-path-prefix is the bit that includes the domain of the app

fabis94:
Apparently we may be able to avoid having to specify the app domain, and we can just specify the path prefix that leads to the minified assets (/_nuxt)

PR #2387, opened 2024-06-18
Fix: Change /downloads links to https://speckle.systems/features/connectors/

Mikehrn:
## Description & motivation

This changes links from `/downloads` to `https://speckle.systems/features/connectors/` to point users in the right direction (desktop app). Also opens in a new tab now it's a link to the speckle website instead of the app

### DIFF
--- a/packages/frontend-2/components/header/NavUserMenu.vue
+++ b/packages/frontend-2/components/header/NavUserMenu.vue
@@ -20,17 +20,18 @@
           class="absolute right-4 top-14 w-56 origin-top-right bg-foundation outline outline-2 outline-primary-muted rounded-md shadow-lg overflow-hidden"
         >
           <MenuItem v-slot="{ active }">
-            <a
+            <NuxtLink
               :class="[
                 active ? 'bg-foundation-focus' : '',
                 'flex gap-3 border-b border-primary items-center px-3 py-3 text-sm text-primary cursor-pointer transition mb-1'
               ]"
               target="_blank"
-              href="https://speckle.systems/features/connectors/"
+              external
+              :href="connectorsPageUrl"
             >
               <CloudArrowDownIcon class="w-5 h-5" />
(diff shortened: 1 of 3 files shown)


packages/frontend-2/components/header/NavUserMenu.vue
```diff
@@ -20,16 +20,17 @@
...
           <MenuItem v-slot="{ active }">
-            <NuxtLink
+            <a
               :class="[
                 active ? 'bg-foundation-focus' : '',
                 'flex gap-3 border-b border-primary items-center px-3 py-3 text-sm text-primary cursor-pointer transition mb-1'
               ]"
-              @click="goToConnectors()"
+              target="_blank"
+              href="https://speckle.systems/features/connectors/"
```
fabis94, line 29: I know that this issue was there before, but repeating the same magic string (the path/url to the connectors page) multiple times in multiple files is not [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

We usually define these as constants in `lib/common/helpers/route.ts` and then use those instead.

That said, I have some questions about the necessity of doing this task in the first place, I've left a comment in the ticket

fabis94 (changes requested):

packages/frontend-2/components/header/NavUserMenu.vue
Mikehrn: Cool, most of the other URLs I saw where also hardcoded so I wasn't sure if there was a centralized place we store them, but I will update it :)

benjaminvo: I answered on the Linear issue. TLDR: We want to do this now so we stop linking to /downloads from the product. It's not how we want users to install connectors. But /downloads will replace releases.speckle.systems at some point so we are not killing it.

fabis94 (approved):

PR #2388, opened 2024-06-18
feat(fe2): Filter out ViewerExplorerTreeItem children with speckle_type displaystyle

andrewwallacespeckle:
## Description & motivation
It was decided that DisplayStyle's should not show in the Scene Explorer.
[Linear Ticket
](https://linear.app/speckle/issue/WEB-1079/dont-output-displaystyles-in-the-scene-explorer)

## Changes:
- Updated ViewerExplorerTreeItem to exclude children with a specified speckle_type value in singleCollectionItems and arrayCollections computed properties.
- Ensured that filtered items do not appear in the UI tree structure


benjaminvo (approved):

packages/frontend-2/components/viewer/explorer/TreeItem.vue
```diff
@@ -251,8 +256,10 @@ const arrayCollections = computed(() => {
 
     const ids = val.map((ref) => ref.referencedId) // NOTE: we're assuming all collections have refs inside; might revisit/to think re edge cases
 
-    const actualRawRefs = props.treeItem.children.filter((node) =>
-      ids.includes(node.raw?.id as string)
+    const actualRawRefs = props.treeItem.children.filter(
+      (node) =>
+        ids.includes(node.raw?.id as string) &&
+        node.raw?.speckle_type !== 'Objects.Other.DisplayStyle'
```
fabis94, line 262: This same check is duplicated twice in 2 separate and very complicated functions, meaning - it can be pretty easy to update one and forget about the other.

Move the condition to a fn, something like this:
```ts
// array to enable easy addition of extra types in the future
const isAllowedType = (node: ExplorerNode) => ['Objects.Other.DisplayStyle'].includes(node.raw?.speckle_type || '')
```

fabis94 (changes requested):

fabis94 (approved):

PR #2389, opened 2024-06-18
fix(fe2): keyboard shortcuts in viewer don't work

andrewwallacespeckle:
## Description & motivation
This PR addresses the issue of key combinations not being correctly detected. In macOS, to the special handling of the Option (Alt) key results in special characters being input instead of the intended key combinations.

The change utilises the useMagicKeys functionality from the @vueuse/core library to improve the detection of keyboard shortcuts, ensuring a more consistent and reliable user experience across different operating systems.

I changed the main key to "Shift" as it does not have native conflicts, and should work well in PC and Mac. 

**Please check this on Windows - I have only tested on Mac**

## Changes:
- Use useMagicKeys from vueuse
- Swap key to "Shift"
- Update tooltips


packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -365,25 +365,25 @@ const {
...
-  `Scene Explorer (${getKeyboardShortcutTitle([ModifierKeys.AltOrOpt, 'e'])})`
+  `Scene Explorer (${getKeyboardShortcutTitle([ModifierKeys.Shift, 'e'])})`
 )
 const discussionsShortcut = ref(
-  `Discussions (${getKeyboardShortcutTitle([ModifierKeys.AltOrOpt, 't'])})`
+  `Discussions (${getKeyboardShortcutTitle([ModifierKeys.Shift, 't'])})`
 )
 const zoomExtentsShortcut = ref(
-  `Fit to screen (${getKeyboardShortcutTitle([ModifierKeys.AltOrOpt, 'Space'])})`
+  `Fit to screen (${getKeyboardShortcutTitle([ModifierKeys.Shift, 'Space'])})`
```
fabis94, line 377: This one doesn't work for me on windows

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -396,30 +396,31 @@ const toggleActiveControl = (control: ActiveControl) => {
   activeControl.value = activeControl.value === control ? 'none' : control
 }
 
-onKeyboardShortcut([ModifierKeys.AltOrOpt], 'm', () => {
+onKeyboardShortcut([ModifierKeys.Shift], 'm', () => {
```
fabis94, line 399: Mapping between Keyboard Shortcut and expected action is duplicated, and thus its easy to make the mistake of setting one thing above and different one below, especially when there's so many actions. That may already be the reason behind Shift+Space not working, I notice its ' ' not 'Space' below, but that's just a guess.

Create a DRY mapping kinda like this:
```
enum ViewerKeyboardActions {
  ToggleModels,
  ToggleExplorer,
...
}

const map: Record<ViewerKeyboardActions, Array<ModifierKeys[], string>> = {
  [ViewerKeyboardActions.ToggleModels]: [[ModifierKeys.Shift], 'x'],
...
}

const modelsShortcut = ref(
  `Models (${getKeyboardShortcutTitle(map[ViewerKeyboardActions.ToggleModels])})`
)

onKeyboardShortcut(...map[ViewerKeyboardActions.ToggleModels], () => {
  toggleActiveControl('models')
})
```

fabis94 (changes requested):

andrewwallacespeckle:
We discussed this in Copenhagen and decided to remove the modifier key. We already use WASD without alt/shift, so it makes sense to have the other shortcuts work in the same way. This should simplify the windows/mac problems, and make it easier for the user.

fabis94:
> We discussed this in Copenhagen and decided to remove the modifier key. We already use WASD without alt/shift, so it makes sense to have the other shortcuts work in the same way. This should simplify the windows/mac problems, and make it easier for the user.

@andrewwallacespeckle  Maybe it's fine, but theoretically that sounds worse haha, like random keypresses getting captured and triggering stuff in the viewer when the user didn't want it to. What happens if you're typing inside of a, lets say, comment reply box and you type any of the special characters, have you validated that they don't also cause those viewer actions?

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -364,26 +359,60 @@ const {
...
+  [ViewerKeyboardActions.ToggleModels]: 'm',
+  [ViewerKeyboardActions.ToggleExplorer]: 'e',
+  [ViewerKeyboardActions.ToggleDiscussions]: 't',
+  [ViewerKeyboardActions.ToggleMeasurements]: 'r',
+  [ViewerKeyboardActions.ToggleProjection]: 'p',
+  [ViewerKeyboardActions.ToggleSectionBox]: 'b',
+  [ViewerKeyboardActions.ZoomExtentsOrSelection]: ' '
+}
+
+const getKeyboardShortcutTitle = (key: string): string => key
```
fabis94, line 382: Currently pointless. You could at least do the `map[key]` bit here, instead of having to duplicate the map[..] reference in all usages below

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -396,31 +425,39 @@ const toggleActiveControl = (control: ActiveControl) => {
...
-onKeyboardShortcut([ModifierKeys.AltOrOpt], ' ', () => {
-  trackAndzoomExtentsOrSelection()
-})
-onKeyboardShortcut([ModifierKeys.AltOrOpt], 'p', () => {
-  toggleProjection()
-})
-onKeyboardShortcut([ModifierKeys.AltOrOpt], 'b', () => {
-  toggleSectionBox()
+Object.keys(map).forEach((actionKey) => {
+  const action = parseInt(actionKey) as ViewerKeyboardActions
```
fabis94, line 455: Why the parseInt? The indices should already be numbers.

If TS is saying that they might not be, I think it could be cause in runtime TS enums get converted to an object that has not only the key->value pairs you've specified, but also the inverse: value->key pairs. So it could be that you can also get each enum name as a key. If that's true (validate this), parseInt() is def not what you want to do and probably would trigger some buggy behaviour. You should instead filter those string keys out.

fabis94 (changes requested):
minor comments, mostly i'm concerned about keyboard shortcuts w/o modifier keys, but if you're ok with it then its fine I guess

andrewwallacespeckle:
> minor comments, mostly i'm concerned about keyboard shortcuts w/o modifier keys, but if you're ok with it then its fine I guess

I agree around the keyboard shortcuts, so i've re-added the modifier keys. The changes should also be a lot more DRY now, but please let me know if i misunderstood something, or it could be improved further. Thanks

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -364,28 +375,76 @@ const {
...
+      toggleSectionBox()
+      break
+    case ViewerKeyboardActions.ZoomExtentsOrSelection:
+      trackAndzoomExtentsOrSelection()
+      break
+  }
+}
+
+Object.entries(map).forEach(([actionKey, [modifiers, key]]) => {
+  const action = Number(actionKey) as ViewerKeyboardActions
```
fabis94, line 444: @andrewwallacespeckle  If my assumption about numeric enums during runtime is correct, this doesn't resolve the issue. If `actionKey` is a string, I don't even know what you're gonna get out of `Number(string)`, maybe `NaN`? And then feeding that into the keyboard action handler, who know's whats happening.

In the forEach you should do `if (isString(actionKey)) return`

packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -364,28 +375,76 @@ const {
   diff: { enabled }
 } = useInjectedViewerInterfaceState()
 
+const map: Record<ViewerKeyboardActions, [ModifierKeys[], string]> = {
+  [ViewerKeyboardActions.ToggleModels]: [[ModifierKeys.Shift], 'm'],
+  [ViewerKeyboardActions.ToggleExplorer]: [[ModifierKeys.Shift], 'e'],
+  [ViewerKeyboardActions.ToggleDiscussions]: [[ModifierKeys.Shift], 't'],
+  [ViewerKeyboardActions.ToggleFilters]: [[ModifierKeys.Shift], 'f'],
```
fabis94, line 382: Shift+f does nothing and I don't even see any kind of component in the template for this, so I've no idea what this is even supposed to be

fabis94 (changes requested):
everything works except Shift+f

packages/frontend-2/components/viewer/Controls.vue
andrewwallacespeckle: Sorry, you're right. I copied the list from the `type ActiveControl` which includes a filters. But I've removed the shortcuts to this now.

fabis94 (approved):

PR #2396, opened 2024-06-18
feat(fe2): Improve property name display and search functionality in filtering

andrewwallacespeckle:
## Description & motivation
This PR addresses user feedback from the forums regarding the display of property names and the search functionality in our UI for Revit files. Users requested more user-friendly property names instead of raw database keys and improved search capabilities.

## Changes:
- Display user-friendly property names for Revit files instead of raw database keys.
- Enhance the search functionality to include user-friendly property names for Revit files.
- Ensure that non-Revit properties are displayed and searched as usual, without modification.


packages/frontend-2/components/viewer/explorer/Filters.vue
```diff
@@ -246,4 +230,25 @@ const refreshColorsIfSetOrActiveFilterIsNumeric = () => {
   // removePropertyFilter()
   applyPropertyFilter()
 }
+
+const isRevitProperty = (key: string): boolean => {
```
fabis94, line 234: 1. This same helper can be used on like 145
2. Dunno how many times this gets invoked (are there thousands of such checks?), but if there are many it would be more performant to have a regexp defined outside of the function and testing the string against that instead

packages/frontend-2/components/viewer/explorer/Filters.vue
```diff
@@ -246,4 +230,25 @@ const refreshColorsIfSetOrActiveFilterIsNumeric = () => {
...
+}
+
+const getPropertyName = (key: string): string => {
+  if (!key) return 'Loading'
+
+  if (key === 'level.name') return 'Level Name'
+  if (key === 'speckle_type') return 'Object Type'
+
+  if (isRevitProperty(key)) {
+    const correspondingName = props.filters.find(
```
fabis94, line 245: Are we sure that this is gonna be a StringPropertyInfo? PropertyInfo can inherit both StringPropertyInfo and NumericPropertyInfo filters.

Better to define and use a type guard instead: `isStringPropInfo(i: PropertyInfo): i is StringPropertyInfo => i.type === 'string'`

fabis94 (changes requested):

packages/frontend-2/components/viewer/explorer/Filters.vue
andrewwallacespeckle: This helper is not fully DRY to maintain the original logic structure, but it is better than it was

packages/frontend-2/components/viewer/explorer/Filters.vue
```diff
@@ -248,4 +234,27 @@ const refreshColorsIfSetOrActiveFilterIsNumeric = () => {
   // removePropertyFilter()
   applyPropertyFilter()
 }
+
+const isRevitProperty = (key: string): boolean => {
+  return revitPropertyRegex.test(key)
+}
+
+const getPropertyName = (key: string): string => {
```
fabis94, line 242: Seems like everywhere where we use this fn we have access to the entire filter object. If we pass that in, instead of just the key, we don't have to do the `props.filters.find()` bit

andrewwallacespeckle: Passing the entire filter object introduced complexity with a fallback mechanism and error handling in title. Let me know if this is enough to stop us merging and I can take another look on Monday but I'm running out of time today

fabis94 (approved):

fabis94 (approved):

PR #2409, opened 2024-06-19
WEB-1096: DataDog Errors

AlexandruPopovici:
## Description & motivation
Complete information in https://linear.app/speckle/issue/WEB-1096/clean-up-viewer-errors#comment-5f5a5b63
This PR fixes part of the reported errors, however for the remaining I could not find a way to replicate and there is a good chance that they are caused by data anomalies in speckle streams, in which case we need the actual data associated with the data dog report in order to be able to reproduce

## Changes:
-  We're now catching possibe QuateExceeded errors from IDB in the local storage
- `findId` with undefined or null id now returns null

## Checklist:


packages/objectloader/src/index.js
```diff
@@ -576,15 +576,18 @@ export default class ObjectLoader {
...
+      const store = this.cacheDB
+        .transaction('objects', 'readwrite')
+        .objectStore('objects')
+      for (const obj of objects) {
+        const idAndData = obj.split('\t')
+        store.put(idAndData[1], idAndData[0])
+      }
+      return this.promisifyIdbRequest(store.transaction)
+    } catch (e) {
+      this.logger.error(e)
```
fabis94, line 589: There's no longer a return in the function in the error scenario. Even if `undefined`/`void` is fine, we should be explicit about it.

fabis94 (changes requested):

fabis94 (approved):

PR #2429, opened 2024-06-25
refactor(preview-service): conforms to design patterns, adds tests, and is typed

iainsproat:
## Description & motivation

Preview Service was difficult to maintain.
It was without tests which allowed regressions to occur.
There was little visibility as to what was happening within Puppeteer/Chromium in production.

## Changes:

- puppeteer/chromium logs are output to console
- dependencies on knex, puppeteer, and fetch are now dependency injected
- typescript all the server components
- rearrange the directory structure to separate concerns; including moving knex/SQL to repositories, and isolating puppeteer scripts from server-side code.
- CI job to lint and test the preview-service
- (WIP): unit & integration tests with vitest
- remove unused Makefile
- environment variables are read via helpers in utils directory

## To-do before merge:

- [x] have typescript resolve imports properly
- [x] fix db migrations in tests
- [x] add more tests

## Screenshots:

Screenshot 2024-06-26 at 01 23 16

Behold these logs of much beauty, including logs from within Chromium/Puppeteer 🥳  . (FYI I've enabled pretty printing of the logs while capturing this screenshot, in production the logs are json formatted)
Screenshot 2024-06-25 at 10 51 05

And we have tests 🎉 
Screenshot 2024-06-26 at 11 55 09
Screenshot 2024-06-26 at 11 54 59

## Checklist:

### DIFF
--- a/packages/preview-service/Dockerfile
+++ b/packages/preview-service/Dockerfile
@@ -58,7 +58,7 @@ RUN apt-get update && \
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y \
     --no-install-recommends \
-    google-chrome-stable=128.0.6613.84-1 \
+    google-chrome-stable=128.0.6613.119-1 \
     fonts-ipafont-gothic=00303-23 \
     fonts-wqy-zenhei=0.9.45-8 \
     fonts-thai-tlwg=1:0.7.3-1 \


packages/preview-service/server/routes/api.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getObjectsStreamFactory } from '../../repositories/objects'
+import { SpeckleObjectsStream } from '../../utils/speckleObjectsStream'
+import { pipeline, PassThrough } from 'stream'
+import type { Knex } from 'knex'
+
+const apiRouterFactory = (deps: { db: Knex }) => {
+  const { db } = deps
+  const apiRouter = express.Router()
+
+  const isSimpleTextRequested = (req: express.Request) =>
```
fabis94, line 12: This could be a helper imported from a helper module, since I saw other endpoints also using a check like this

packages/preview-service/server/server.ts
```diff
@@ -0,0 +1,107 @@
...
+  metricsServer.listen(metricsPort, host)
+
+  return { app, server, metricsServer }
+}
+
+/**
+ * Normalize a port into a number, string, or false.
+ */
+function normalizePort(val: string | number) {
+  const port = typeof val === 'string' ? parseInt(val, 10) : val
```
fabis94, line 59: Use lodash
```ts
const port = _.toNumber(val)
if (!_.isNan(port) && port >= 0) return port

throw new Error(...)
```

packages/preview-service/server/server.ts
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Event listener for HTTP server "error" event.
+ */
+
+const onErrorFactory = (port: string | number | false) => (error: Error) => {
+  if ('syscall' in error && error.syscall !== 'listen') {
+    throw error
+  }
+
+  const bind = typeof port === 'string' ? 'Pipe ' + port : 'Port ' + port
```
fabis94, line 82: _.isString is more feature complete than a `typeof XX === string` check

packages/preview-service/services/screenshot.ts
```diff
@@ -0,0 +1,120 @@
...
+      headless: shouldBeHeadless(),
+      userDataDir: getPuppeteerUserDataDir(),
+      executablePath: getChromiumExecutablePath(),
+      protocolTimeout: 3600_000, //TODO make this configurable to match the maximum timeout of the service
+      // we trust the web content that is running, so can disable the sandbox
+      // disabling the sandbox allows us to run the docker image without linux kernel privileges
+      args: ['--no-sandbox', '--disable-setuid-sandbox', '--disable-dev-shm-usage']
+    }
+
+    await deps.puppeteerClient.init(launchParams)
```
fabis94, line 34: It bothers me a little bit that PuppeterClient creation & initialization are disconnected and done in different files. Makes it possible to work with a PuppeteerClient that hasn't been initialized without knowing it.

tbh, I'd convert PuppeteerClient away from a class into an async function, to achieve the equivalent of an "async constructor", that way you wouldn't have to worry about passing around a non-initialized client instance

```
const createPuppeteerClient = async (...initParams) => {
  return {
    loadPageAndEvaluateScript,
    dispose
  }
}
```

Also - what if the client is already initialized? It seems that init() is not idempotent and can be ran multiple times, creating a new browser each time and apparently not cleaning the old one up

fabis94: Maybe you can achieve async initialization with classes too? I don't know enough about them to say, but in other programming languages you could have a private constructor and a public async static function that would be the only way to initialize the class.

The functional approach seems less complicated to me imo

packages/preview-service/services/taskManager.ts
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    try {
+      const task = await deps.getNextUnstartedObjectPreview()
+
+      // notify the healthcheck that we are still alive
+      deps.updateHealthcheckData()
+
+      if (!task) {
+        setTimeout(repeatedlyPollForWorkFactory(deps), 1000)
```
fabis94, line 41: weird/smelly logic here, but i assume this is largely a copy of what existed before.

smelly cause:
1. Duplicate `setTimeout(...)` calls that could be extracted to a parameterizable function `await timeout(xx)`
2. Devs shouldn't have to (remember to) manually invoke the next loop of the timeout. In fact, the actual looping/timeout bit feels like a separate concern altogether. I'd probably make this function just return a status enum, and then a different function would handle invoking this one, and setting up the correct setTimeout() depending on the status enum

packages/preview-service/tests/hooks.ts
```diff
@@ -0,0 +1,70 @@
...
+        .select('tablename')
+        .where({ schemaname: 'public' })
+        .whereRaw("tablename not like '%knex%'")
+        .whereNotIn('tablename', protectedTables)
+    ).map((table: { tablename: string }) => table.tablename)
+  }
+
+  if (!tableNames || !tableNames.length) return
+
+  await db.raw(`truncate table ${tableNames.join(',')} cascade`)
```
fabis94, line 28: Doesn't all of this happen in the same DB that `server` works off of, at least locally? We def wouldn't want to truncate/delete all of that data

fabis94: Even if this is only meant for a test DB, I just want to be sure that there's no way this can accidentally run on the main local speckle server DB and wipe everything there

packages/preview-service/tests/integration/repositories/objectPreview.spec.ts
```diff
@@ -0,0 +1,34 @@
...
+import { beforeEachContext } from '../../hooks'
+
+describe('Repositories: ObjectPreview', () => {
+  const db = getTestDb()
+  beforeEach(async () => {
+    await beforeEachContext({ db })
+  })
+  describe('getNextUnstartedObjectPreview', () => {
+    it('should return the next unstarted object preview', async () => {
+      const ObjectPreview = () => db('object_preview')
```
fabis94, line 14: typed db instance preferably + move this to the repository layer?

packages/preview-service/tests/testKnex.ts
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable camelcase */
+require('ts-node/register') //required to allow knex to understand typescript migration files
```
fabis94, line 2: Weird, you shouldn't need to do this. If knex migrations are ran by vitest, it should already be running w/ ts-node (or the equivalent).

Also - we probably want to do the same thing we do in `server` and only allow built .js migrations (in dist folder) to be ran, so that you can't accidentally run the same migration twice - once as a ts file with ts-node, and the 2nd time when its ran through dist.

packages/preview-service/tests/unit/services/taskManager.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+          called['notifyUpdate'] = called['notifyUpdate']++ || 1
+          expect(task).toEqual({ streamId: 'streamId', objectId: 'objectId' })
+        },
+        onExit: () => {
+          called['onExit'] = called['onExit']++ || 1
+        },
+        logger
+      })
+
+      repeatedlyPollForWork()
```
fabis94, line 44: Shouldn't you await this?

packages/preview-service/services/screenshot.ts
```diff
@@ -0,0 +1,120 @@
...
+  (deps: {
+    puppeteerClient: PuppeteerClientInterface
+    logger: Logger
+    serviceOrigin: string
+  }): GetScreenshot =>
+  async (params) => {
+    const objectUrl = `${deps.serviceOrigin}/streams/${params.streamId}/objects/${params.objectId}`
+    const launchParams = {
+      //TODO as the launch params are touching environment variables, should they be dependency injected?
+      headless: shouldBeHeadless(),
```
fabis94, line 25: I noticed that all of these 3 env vars have defaults, altho we probably don't want to pass in anything at least in dev environments, so it does its own algo to figure out where to find chromium, where to put stuff etc.

packages/preview-service/services/taskManager.ts
```diff
@@ -0,0 +1,81 @@
+import type { UpdateHealthcheckData } from 'clients/execHealthcheck'
```
fabis94, line 1: broken import, the path should be relative

packages/preview-service/package.json
```diff
@@ -14,10 +14,14 @@
...
+    "dev": "LOG_PRETTY=true LOG_LEVEL=debug nodemon --trace-deprecation ./bin/www",
     "dev:headed": "PREVIEWS_HEADED=true yarn dev",
-    "build": "webpack --env dev --config webpack.config.render_page.js && webpack --env build --config webpack.config.render_page.js",
-    "lint": "eslint ."
+    "build": "tsc -p ./tsconfig.build.json && webpack --env dev --config webpack.config.renderPage.js && webpack --env build --config webpack.config.renderPage.js",
+    "lint": "yarn lint:tsc && yarn lint:eslint",
+    "lint:ci": "yarn lint:tsc",
+    "lint:tsc": "tsc --noEmit",
+    "lint:eslint": "eslint .",
+    "test": "NODE_ENV=test LOG_LEVEL=silent LOG_PRETTY=true vitest --run"
```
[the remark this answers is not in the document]
fabis94: I mean the "main" key at the top level of package.json, outside of scripts

PR #2435, opened 2024-06-25
chore(server): marking all old (non-FE2) GQL schema fields as deprecated

fabis94:
Note that this includes fields still in use by DUI3 and Connectors/Manager

### DIFF
--- a/packages/server/assets/accessrequests/typedefs/accessrequests.graphql
+++ b/packages/server/assets/accessrequests/typedefs/accessrequests.graphql
@@ -4,7 +4,9 @@ extend type Query {
   """
   streamAccessRequest(streamId: String!): StreamAccessRequest
     @hasServerRole(role: SERVER_GUEST)
-    @deprecated(reason: "Part of the old API surface and will be removed in the future")
+    @deprecated(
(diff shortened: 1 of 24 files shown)


packages/server/assets/accessrequests/typedefs/accessrequests.graphql
```diff
@@ -4,13 +4,20 @@ extend type Query {
   """
   streamAccessRequest(streamId: String!): StreamAccessRequest
     @hasServerRole(role: SERVER_GUEST)
+    @deprecated(
```
fabis94, line 7: Stream Access Requests don't have any FE2 implementation

packages/server/assets/activitystream/typedefs/activity.graphql
```diff
@@ -11,6 +11,9 @@ extend type User {
   ): ActivityCollection
     @hasServerRole(role: SERVER_GUEST)
     @hasScope(scope: "users:read")
+    @deprecated(
```
fabis94, line 14: There's nothing to do with activity streams in FE2

gjedlicska (approved):

PR #2437, opened 2024-06-25
Change web app font

Mikehrn:
Change the font to Inter, to align with the new design system. This change only effects the web app, so the font is not included in other places (it uses the existing defined systems fonts as fall back).

There are some packages that can be used for this, like [Nuxt Google Fonts](https://google-fonts.nuxtjs.org/), but it felt overkill just include a package just to load one font. Maybe something to consider in the future when/if we will need multiple.


gjedlicska:
Yay for Inter

benjaminvo (dismissed):
Approved! But would like quick review from @fabis94 too on how font is imported.

fabis94:
@Mikehrn @benjaminvo This is a pretty big change with implications across our entire design system and apps that use it (DUI3, FE2, Automate), are we OK with this new font now being the main one everywhere? Might want to confirm with Dim

packages/frontend-2/app.vue
```diff
@@ -36,6 +36,8 @@ watchAuthQueryString()
 await useMixpanelInitialization()
 </script>
 <style>
+@import 'https://fonts.googleapis.com/css2?family=Inter:wght@100..900&display=swap';
```
fabis94, line 39: @Mikehrn You shouldn't do this in app.vue, it's a change that should belong to the design system and thus be auto-applied anywhere where our tailwind theme is used. Not only that, but we don't want these fonts to be loaded from an external API, because then our app's performance becomes dependant on external resources that we can't control (e.g. what if googleapis.com goes down? or undergoes high load and becomes slow?).

tl;dr; These fonts should be actually downloaded into the `tailwind-theme` package, and then imported/applied by the tailwind theme plugin (or preset?), so that you don't need to do anything manually in FE2 or any other apps to ensure that the font is loaded.

fabis94 (changes requested):

packages/frontend-2/app.vue
Mikehrn: The reason why I went with the CDN is because in most cases it would be more performant when the user already has this file cached (which is with Inter a pretty high chance), in case it's not available it will fallback on the fonts we are currently using. We only added it to FE2 on purpose, to only affect the web app for now, but we will talk with Dim to check if we can just enable it everywhere :) If so I'll change it to the `tailwind-theme`

fabis94: @Mikehrn If it's only meant for FE2, then there shouldn't be any changes to the tailwind-theme package. That said, I don't see why we would want the font to differ between apps, if the theme/design system is the same.

packages/frontend-2/assets/css/fonts.css
```diff
@@ -0,0 +1,74 @@
+/* cyrillic-ext */
```
Mikehrn, line 1: @fabis94 I can include this as well in dui3 so it has the correct font (if not it will fallback on the fonts we are already using it), but I dont know how to test that. Should I change it or leave it as is for now?

fabis94 (approved):

PR #2452, opened 2024-06-28
Soft Pipeline Reset

AlexandruPopovici:
## Description & motivation
This PR finally improves on how and when we reset the rendering pipeline. In this context a pipeline reset means a restart of the accumulation stage of the rendering pipeline. In the accumulation stage, static AO gets progressively computed. Normally, there are discrete moments where this reset needs to occur. Generally we only need to re-generate static AO when either the camera changes in any way, or there is a change in scene geometry of any kind. Previously, to make it easier on us, and mostly because the old camera controller was so horrible at determining camera movement, we would just reset the pipeline regardless if it was needed or not. The downside of this is things looking worse than they should for now good reason. Here's an example of how the measurement tool use to work and how often the pipeline got reset. i.e the image switches between "fuzzy" and "not fuzzy", aka switching between dynamic low quality AO and the static higher quality one

https://github.com/specklesystems/speckle-server/assets/11405391/79a0b8c1-7076-44bb-954c-781da482744f

Now, with soft pipeline reset, this is no longer happening

https://github.com/specklesystems/speckle-server/assets/11405391/fd78e70b-2c9e-41ac-9515-de8171eca72e

Besides the measuring tool there are other instances where the pipeline was reset during
- Highlighting objects
- Selecting objects
- Changing light params
- When using the section tool, hovering on the move gizmo no longer causes a reset. This was as frequent and as annoying as the measurement tool one

## Checklist:


benjaminvo (dismissed):
Veery nice update!

benjaminvo (approved):

PR #2459, opened 2024-07-01
Feat (dui3): shared files

oguzhankoral:
# People still use files

"Requesting Write Access" from DUI3 still a product question that we should clarify, so far checked with Claire and Bilal, but need your idea on this. We can disable it easliy. LMK what do you think @didimitrie.

- We have `projectAccount` in `ProjectModelGroup` and `CardBase` to have related account as fallback if it doesn't match with provided account id but match with server url.
- We mark project group as read only if user doesn't have write access even if server matches. We prevent to click operation on Sender card with this info
  - `Sender` can't do send operation if account has `stream:reviewer` role on project.
  - `Receiver` can receive regardless if project is public.
 
   image
- User can request write-access by clicking lock button and we listen `userProjectsUpdatedSubscription` to refetch data. Whenever owner of the project accept request, project become normal state that can send objects only if owner assigned as `stream:collaborator` or `stream:owner`

  Rhino_9cLIkkUDWj
- Whenever user remove the collaborator from project, UI also catch it and mark project read-only again. This is good because it solves the common issue if some roles changed on the way.
  
  Rhino_QFQrbpIN5r

- When the project is private, we show common apollo error
  image

PS: Most of the logic handled in UI but it also has alignment changes on sharp side -> https://github.com/specklesystems/speckle-sharp/pull/3550


didimitrie:
Looks good. I'm inclined to piggy back on this PR with some extra little things: 
- in the send wizard's first step, disable all projects where `role <= reviewer`
- possibly nuance on a per model card basis based on project role: 

image

I'm a reviewer in this project, so this means: 
- i can receive things
- i cannot load things

didimitrie:
Preventing selecting projects where role = reviewer in send wizard: 

image

Reworks the whole "can't send data here" paradigm to work only on sender cards: 

image

didimitrie (approved):

oguzhankoral (commented):
That's fine for now I believe. Merging it!

PR #2467, opened 2024-07-02
WEB-1147 project creation emit domain event

cdriesler:
## Changes:

- Update stream/project creation to accept a `workspaceId`
- Authorize user for given operation
- Emit `ProjectCreated` on create


packages/server/assets/core/typedefs/projects.graphql
```diff
@@ -30,6 +30,7 @@ input ProjectCreateInput {
   name: String
   description: String
   visibility: ProjectVisibility
+  workspaceId: String
```
gjedlicska, line 33: I'm a bit unsure about this, maybe we'll split the project creation resolver for workspaces into the workspaces schema. For now this is ok, but we should go through all the licensing implications.

gjedlicska (changes requested):

packages/server/assets/core/typedefs/projects.graphql
cdriesler: Leaning more towards the split more, but also in agreement that this can wait

cdriesler:
@gjedlicska Back to you, extracted the check to `isWorkspacesModuleEnabled()`, even though it's a bit silly, so we have one place to check during the eventual refactor to The Real Check.

gjedlicska (approved):

PR #2468, opened 2024-07-02
Object Sender: Adds @Detach and @Chunkable decorators

vwnd:
## Description & motivation

Adds the ability for using decorators in class declarations to describe detached and chunkable fields.

```typescript
class RandomJoe extends Base {
  @Detach()
  @Chunkable(10)
  numbers: number[]

  constructor(props?: Record<string, unknown>) {
    super(props)
    this.numbers = times(100, () => 42)
  }
}
```

The main reason why I am proposing is that I noticed while playing with the object sender that the frontend-2 assumes certain shapes. For example: Collection.elements will give me a nice UI, but Collection['@elements'] won't. I imagine there are other parts of the ecosystem that have those behaviours.

image

image

## Changes:

- Added @Detach() and @Chunkable(n) decorators
- Added examples of classes using the decorators
- Added a createCommit helper

## References

https://www.typescriptlang.org/docs/handbook/decorators.html#metadata

### DIFF
--- a/packages/objectsender/src/examples/browser/main.ts
+++ b/packages/objectsender/src/examples/browser/main.ts
@@ -1,5 +1,6 @@
 import { send, Base, type SendResult, Detach } from '../../index'
 import { times } from '#lodash'
+import { createCommit } from './utils'
 
 interface ExampleAppWindow extends Window {
   send: typeof import('../../index').send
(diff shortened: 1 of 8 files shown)


packages/objectsender/src/examples/browser/main.ts
```diff
@@ -118,3 +127,33 @@ class RandomFoo extends Base {
...
+  @Chunkable(10)
+  numbers: number[]
+
+  constructor(props?: Record<string, unknown>) {
+    super(props)
+    this.numbers = times(100, () => 42)
+  }
+}
+
+export class Collection<T extends Base> extends Base {
```
didimitrie, line 142: Nice! I think some of this in the future can be slowly extracted out to canonical speckle types proper, and exported out by the lib.

didimitrie (approved):

didimitrie:
Thanks for the sweet contribution @vwnd! Merging now.

PR #2470, opened 2024-07-03
feat(user-emails): create user emails table and type

alemagio:
## Description & motivation

- create a new module for user-emails (I'm not sure if we want it inside another module, anyway we can move it any time)
- create user_emails table
- create UserEmail domain type

## Checklist:

### DIFF
--- a/packages/server/modules/core/tests/integration/user-emails.spec.ts
+++ b/packages/server/modules/core/tests/integration/user-emails.spec.ts
@@ -1,0 +1,41 @@
+import { before } from 'mocha'
+import { createUser } from '../../services/users'
+import { beforeEachContext } from '@/test/hooks'
+import { expect } from 'chai'
+import { getUserByEmail } from '../../repositories/users'
+import { UserRecord } from '../../helpers/types'
+
+describe.only('Core @user-emails', () => {
+  before(async () => {
+    await beforeEachContext()
+  })
+  describe('getUserByEmail', () => {
+    it('should return null if user does not exist', async () => {
+      expect(await getUserByEmail('test@example.org')).to.be.null
+    })
+
+    it('should return user if user-email does not exist', async () => {
+      await createUser({
+        name: 'John Doe',
+        email: 'test@example.org',
+        password: 'sn3aky-1337-b1m'
+      })
+
+      const user = (await getUserByEmail('test@example.org')) as UserRecord
+      expect(user.name).to.eq('John Doe')
+      expect(user.email).to.eq('test@example.org')
+    })
+
+    it('should return user merged with user-email', async () => {
+      await createUser({
+        name: 'John Doe',
+        email: 'test@example.org',
+        password: 'sn3aky-1337-b1m'
+      })
+
+      const user = (await getUserByEmail('test@example.org')) as UserRecord
+      expect(user.name).to.eq('John Doe')
+      expect(user.email).to.eq('test@example.org')
+    })
+  })
+})
--- a/packages/server/modules/core/tests/integration/user-emails.spec.ts
+++ b/packages/server/modules/core/tests/integration/user-emails.spec.ts
@@ -26,6 +26,7 @@ describe('Core @user-emails', () => {
         email,
         password: 'sn3aky-1337-b1m'
       })
+      // TODO: delete user email
 
       const user = (await getUserByEmail('test@example.org')) as UserRecord
       expect(user.name).to.eq('John Doe')
--- a/packages/server/modules/core/tests/integration/user-emails.spec.ts
+++ b/packages/server/modules/core/tests/integration/user-emails.spec.ts
@@ -4,6 +4,11 @@ import { beforeEachContext } from '@/test/hooks'
 import { expect } from 'chai'
 import { getUserByEmail } from '../../repositories/users'
 import { UserRecord } from '../../helpers/types'
+import crs from 'crypto-random-string'
+
+function createRandomEmail() {
+  return `${crs({ length: 6 })}@example.org`
+}
 
 describe('Core @user-emails', () => {
   before(async () => {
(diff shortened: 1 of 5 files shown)


fabis94 (changes requested):

packages/server/modules/user-emails/migrations/20240703084247_user-emails.ts
```diff
@@ -0,0 +1,25 @@
+import { Knex } from 'knex'
+
+const USER_EMAILS_TABLE_NAME = 'user_emails'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(USER_EMAILS_TABLE_NAME, (table) => {
+    table.string('id').notNullable().primary()
```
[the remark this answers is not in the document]
fabis94: @iainsproat Can you elaborate on both points? I don't fully understand what you mean in the 1st scenario, and the 2nd point is a new one to me - why not?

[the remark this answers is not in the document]
fabis94: Can't you do that now already? If you're first to register an e-mail you're first and no one can do that afterwards, regardless of whether its verified or not.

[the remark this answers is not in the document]
fabis94: Not sure if that's true, if we still mostly rely on users.email and just use this for extra emails. I'm also concerned about opening up a whole new can of worms of multiple people having the same email attached and all of the validation and syncing needed to manage all of that.

That said - I suppose this can be merged and adjusted afterwards when there's an actual usage of this. I don't really get the point of PRs this small when the change doesn't really make sense in isolation, I would've put this together with the logic that actually uses the table.

One thing that I think we should still add is at least a unique index on (userId, email)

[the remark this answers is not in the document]
alemagio: 1. the idea is to drop the `email` col in the `users` table and only use this one. For this reason we need the `primary` column here
2. makes sense to me, we can make the association userId + email as primary key, I don't particularly like the idea of having the email as the PK alone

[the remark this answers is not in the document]
alemagio: I think it is better if we connect later on this @fabis94 @iainsproat

packages/server/modules/core/tests/integration/user-emails.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+describe('Core @user-emails', () => {
+  before(async () => {
+    await beforeEachContext()
+  })
+  describe('getUserByEmail', () => {
+    it('should return null if user does not exist', async () => {
+      expect(await getUserByEmail('test@example.org')).to.be.null
+    })
+
+    it('should return user if user-email does not exist', async () => {
```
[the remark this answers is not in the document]
alemagio: actually no, I created this test to ensure the function still return the user has no user-email because until we will apply the migration to remove the user.email field we want this to work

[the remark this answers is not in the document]
alemagio: oh shit, you're right, I am an idiot :joy:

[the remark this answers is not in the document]
alemagio: this test was not supposed to be pushed here, I added a TODO to fix it in the next PR

packages/server/modules/core/tests/integration/user-emails.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+      const email = createRandomEmail()
+      await createUser({
+        name: 'John Doe',
+        email,
+        password: 'sn3aky-1337-b1m'
+      })
+      // TODO: delete user email
+
+      const user = (await getUserByEmail('test@example.org')) as UserRecord
```
fabis94, line 31: @alemagio Why the forced type cast at the end?

packages/server/modules/core/tests/integration/user-emails.spec.ts
```diff
@@ -0,0 +1,49 @@
+import { before } from 'mocha'
+import { createUser } from '../../services/users'
```
fabis94, line 2: Absolute paths plz

packages/server/modules/emails/tests/verifications.spec.ts
```diff
@@ -23,7 +23,7 @@ import { EmailSendingServiceMock } from '@/test/mocks/global'
 const mailerMock = EmailSendingServiceMock
 
 const cleanup = async () => {
-  await truncateTables([Users.name, EmailVerifications.name])
+  await truncateTables([Users.name, EmailVerifications.name, 'user_emails'])
```
fabis94, line 26: No magic strings please, create a DB helper in dbschema.ts or at the very least stored this as a constant in `domain` or wherever

packages/server/modules/core/tests/integration/user-emails.spec.ts
fabis94: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#75db5175359244e7bb988d753bfa07cd

packages/server/modules/core/tests/integration/user-emails.spec.ts
fabis94: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#3387742436de414082cd4e9607259ad9

packages/server/modules/emails/tests/verifications.spec.ts
fabis94: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#5a51bc8fa7034a9bb5b610bbdba5c847

packages/server/modules/core/tests/integration/user-emails.spec.ts
alemagio: because of the nullable type, I don't see a big issue casts in tests

fabis94: use the non-null assertion operator then https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

at least in that case you're not overriding the entire type, only whether its null or not

packages/server/modules/user-emails/migrations/20240703084247_user-emails.ts
```diff
@@ -0,0 +1,30 @@
+import { Knex } from 'knex'
+import { USER_EMAILS_TABLE_NAME } from '../constants'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable(USER_EMAILS_TABLE_NAME, (table) => {
```
fabis94, line 5: @alemagio You generally don't want to use imported constants in migrations, cause in case the constant changes the migration is gonna run differently (which should never happen)

packages/server/modules/user-emails/migrations/20240703084247_user-emails.ts
```diff
@@ -0,0 +1,29 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.createTable('user_emails', (table) => {
```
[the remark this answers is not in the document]
alemagio: @fabis94 raised this point https://github.com/specklesystems/speckle-server/pull/2470#discussion_r1666811100
And I think it makes sense

gjedlicska (approved):

packages/server/modules/user-emails/constants.ts
gjedlicska, line 1: This file is not needed, can clean up later

PR #2473, opened 2024-07-04
fix(fe2): edit profile dialog - only save name if not empty

andrewwallacespeckle:
## Description & motivation
Michal noticed that although the fields were required, we still saved the name if it is empty.

We now check the contents of name before saving it.

### DIFF
--- a/packages/frontend-2/components/user/profile/EditDialog/Bio.vue
+++ b/packages/frontend-2/components/user/profile/EditDialog/Bio.vue
@@ -16,7 +16,7 @@
             show-label
             show-required
             :rules="[isRequired, isStringOfLength({ maxLength: 512 })]"
-            @change="save"
+            @change="handleNameChange"
           />
           <FormTextInput
             v-model="company"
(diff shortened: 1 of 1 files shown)


benjaminvo (approved):

packages/frontend-2/components/user/profile/EditDialog/Bio.vue
```diff
@@ -16,7 +16,7 @@
             show-label
             show-required
             :rules="[isRequired, isStringOfLength({ maxLength: 512 })]"
-            @change="save"
+            @change="handleNameChange"
```
fabis94, line 19: We shouldn't need to track extra events and do manual validation, when we already have vee-validate rules in place. In fact, the rules set up here should already prevent empty names.

It probably doesn't work because we don't run the validation in any way. Introduce `useForm` here and invoke the form's submission on @change

fabis94 (changes requested):

Mikehrn:
@andrewwallacespeckle We can close this one, it has been fixed in the new modal :)

PR #2476, opened 2024-07-05
sRGB Vertex Colors

AlexandruPopovici:
## Description & motivation
We're now treating all input mesh vertex colors as sRGB. Not all connectors might send them as sRGB, but this is what we agreed on for now

## Checklist:


didimitrie (dismissed):

AlexandruPopovici:
We either need to treat mesh vertex colors coming from connectors as being
in linear space (before PR)
in sRGB space (after PR)

Connectors do not know what color space the host applications send the vertex colors. Regardless we need to choose one as there is no information regarding the color space. Since all other colors are currently being treated as sRGB, it makes half sense to also treat vertex colors as sRGB.

Regarding being a breaking change, it's relative. If a certain host app sends vertex colors in linear space, then the vertex colors will be incorrectly converted. However, it looks like Rhino already sends the colors as sRGB (from the guy's post), so doing nothing about it make the colors wrong as it currently stands. If it's backwards compatible? Same answer

One solution is to not do anything until connectors have standardized the way they send the mesh vertex colors. Realistically, that won't happen very soon as they probably have a lot on their plate already. What we can do is to send a simple mesh with vertex colors as a solid color from all connectors and see what we get.

AlexandruPopovici:
I went over the connectors trying to see what color space they send the vertex colors as. All connectors that do send vertex colors are here https://latest.speckle.systems/projects/5a6609a4b9 Text wise:
qgis: Sends vertex colors as sRGB
rhino: Sends vertex colors as sRGB
unity: Sends vertex colors as sRGB
grasshopper: Sends vertex colors as sRGB
Because we treat incoming vertex color as being in linear space, the resulting color is incorrect. This is also visible with the naked eye in each stream. The natural conclusion is that this PR does not contain breaking changes because all connectors that do send vertex colors send it as sRBG

benjaminvo (approved):

didimitrie (approved):

PR #2483, opened 2024-07-09
fix(modules): correct module names list for workspaces / core spli

cdriesler:
We caught this was wrong and then we made it right


cdriesler:
lmao

gjedlicska:
@cdriesler what happened with this?

cdriesler:
@gjedlicska `/workspacesCore` doesn't exist yet, going to do this after the current workspaces PR in flight is merged

PR #2489, opened 2024-07-10
Alessandro/web 1171 change the users repository to update the verified field in

alemagio:
## Description & motivation

- when user is created a user email is also created
- when user is updated the user email is also updated
- if an email is verified also the user-email is verified
- also fixes https://linear.app/speckle/issue/WEB-1173/change-the-users-repository-to-update-the-email-field

## Checklist:


packages/server/modules/core/helpers/test-helpers.ts
```diff
@@ -0,0 +1,9 @@
+import crs from 'crypto-random-string'
```
fabis94, line 1: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#0746f1c947b741788cf581b9e1659f51

packages/server/modules/core/tests/integration/create-user.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+      name: 'John Doe',
+      email: createRandomEmail(),
+      password: createRandomPassword()
+    }
+
+    // create user
+    await createUser(newUser)
+
+    // try to create user with same email
+    await createUser(newUser)
```
fabis94, line 65: Why .then/.catch instead of a more legible try-catch?

fabis94: Also: We have an `expectToThrow()` util

packages/server/modules/core/tests/integration/create-user.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+    try {
+      await createUser({
+        name: 'Dim Sum',
+        email: createRandomEmail(),
+        password: createRandomPassword(5)
+      })
+    } catch {
+      return
+    }
+    assert.fail('short pwd')
```
fabis94, line 52: 1. expectToThrow()
2. You should check what the error actually is, instead of accepting all errors as an OK result

packages/server/modules/core/tests/integration/update-user.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+  beforeEach(async () => {
+    await beforeEachContext()
+  })
+
+  it('should throw an error if avatar is too big', async () => {
+    const longAvatar = ''.padEnd(524288, '1') + '0'
+    try {
+      await updateUser('123456', { avatar: longAvatar })
+
+      expect.fail('Did not throw')
```
fabis94, line 23: expectToThrow() here and elsewhere in the file

packages/server/modules/user-emails/domain/operations.ts
```diff
@@ -0,0 +1,22 @@
+import { UserEmail } from './types'
```
fabis94, line 1: Do user emails really deserve their own speckle module? There's not much going on in there

packages/server/modules/user-emails/repositories/userEmails.ts
```diff
@@ -0,0 +1,42 @@
+import { Knex } from 'knex'
+import crs from 'crypto-random-string'
+import { CreateUserEmail, DeleteUserEmail, UpdateUserEmail } from '../domain/operations'
+import { UserEmail } from '../domain/types'
+import { USER_EMAILS_TABLE_NAME } from '../constants'
+
+export const createUserEmailFactory =
+  ({ db }: { db: Knex }): CreateUserEmail =>
+  async (userEmail) => {
```
fabis94, line 9: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#29f30b10e5724281b812fe9869b56467

packages/server/modules/user-emails/repositories/userEmails.ts
```diff
@@ -0,0 +1,42 @@
+import { Knex } from 'knex'
+import crs from 'crypto-random-string'
+import { CreateUserEmail, DeleteUserEmail, UpdateUserEmail } from '../domain/operations'
+import { UserEmail } from '../domain/types'
```
fabis94, line 4: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#3387742436de414082cd4e9607259ad9

fix this in all files please. VSCode should be configured to auto-import absolute paths too, so check your config

fabis94 (changes requested):

packages/server/modules/core/repositories/users.ts
```diff
@@ -6,6 +6,9 @@ import { metaHelpers } from '@/modules/core/helpers/meta'
 import { UserValidationError } from '@/modules/core/errors/user'
 import { Knex } from 'knex'
 import { Roles, ServerRoles } from '@speckle/shared'
+import { updateUserEmailFactory } from '@/modules/core/repositories/userEmails'
+import knexInstance from '@/db/knex'
```
fabis94, line 10: No big deal, but i'd just name the import `knex` or even `db` ([to support object property shorthand syntax in factories](https://gomakethings.com/object-property-shorthand-values-with-vanilla-javascript/)), the `Instance` suffix is redundant considering that the difference between the instance and the knex factory is only relevant within `@/db/knex`

packages/server/modules/core/services/users/email-verification.ts
```diff
@@ -0,0 +1,13 @@
+import {
```
fabis94, line 1: Fix this for all files https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#0746f1c947b741788cf581b9e1659f51

packages/server/modules/core/services/users/email-verification.ts
```diff
@@ -0,0 +1,13 @@
...
+  UpdateUserEmail
+} from '@/modules/core/domain/userEmails/operations'
+
+export const markUserEmailAsVerifiedFactory =
+  ({
+    updateUserEmail
+  }: {
+    updateUserEmail: UpdateUserEmail
+  }): MarkUserEmailAsVerified =>
+  async (email) =>
```
fabis94, line 12: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#29f30b10e5724281b812fe9869b56467

packages/server/modules/core/tests/integration/create-user.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+import knexInstance from '@/db/knex'
+import {
+  createRandomEmail,
+  createRandomPassword
+} from '@/modules/core/helpers/testHelpers'
+import { USER_EMAILS_TABLE_NAME } from '@/modules/core/dbSchema'
+import { expectToThrow } from '@/test/assertionHelper'
+import { PasswordTooShortError } from '../../errors/userinput'
+
+const db = knexInstance
```
fabis94, line 15: or just import it with that name, as mentioned above

packages/server/modules/core/tests/integration/create-user.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    expect(storedUser.email).to.equal(user.email.toLowerCase())
+  })
+  it('Should not create a user with a too small password', async () => {
+    const err = (await expectToThrow(() =>
+      createUser({
+        name: 'Dim Sum',
+        email: createRandomEmail(),
+        password: createRandomPassword(5)
+      })
+    )) as Error
```
fabis94, line 54: The `X as Error` type cast is unnecessary here and in other tests as well. The only other thing it can be is a `UnexpectedErrorStructureError` which is already a subclass of `Error` and thus has all of the same properties.

So there's no benefit to it + you also break TypeScript's understanding of what the value is, because it **can** in theory also be a `UnexpectedErrorStructureError` if the test breaks, even if its very unlikely

packages/server/modules/core/tests/integration/create-user.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+    const userId = await createUser({
+      name: 'John Doe',
+      email,
+      password: createRandomPassword()
+    })
+
+    const user = (await getUserById({ userId })) as UserRecord
+    expect(user.email).to.eq(email)
+
+    const userEmail = await userEmailsDB.where({ email, userId, primary: true }).first()
```
fabis94, line 84: Surely this can be a function in the repo. Doesn't necessarily have to be a public domain level operation, but it still belongs in the repo layer. Especially since this seems like a common query that you would want to reuse every once in a while.

packages/server/modules/core/repositories/users.ts
fabis94: @alemagio Is this called `knexInstance` because of VSCode auto-imports, not you manually typing it out that way? If so, I suggest changing the name of the `knexInstance` const inside of `@/db/knex.js` to what we want the import to be called, cause that's what I think VSCode resolves the auto-import name from. Maybe `const db`? So that auto-import always does `import db from '@/db/knex'`

packages/server/modules/core/tests/integration/email-verification.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+  createRandomPassword
+} from '@/modules/core/helpers/testHelpers'
+import { markUserEmailAsVerifiedFactory } from '@/modules/core/services/users/email-verification'
+import { expect } from 'chai'
+import { USER_EMAILS_TABLE_NAME } from '@/modules/core/dbSchema'
+
+const userEmailTable = knexInstance(USER_EMAILS_TABLE_NAME)
+
+describe('Verification @user-emails', () => {
+  it('should mark user email as verified', async () => {
```
fabis94, line 16: Do we really need the test files to be this granular? There is an overheard to each test file - initializing a new test context, running `beforeEachContext` etc.

Some if not all of these can probably be merged into the same test file

fabis94 (changes requested):

packages/server/modules/core/repositories/users.ts
alemagio: yes, that's the auto import.
To avoid breaking existing code I will simply add another export so from now on we can simply have
```ts
import { db } from '...'
```

fabis94: @alemagio Breaking how? The default export can have any kind of arbitrary name, changing the const inside of knex.js only will change VSCode's mechanism for figuring out what to call it, all other places that still do `import whateverArbitraryName from '@/db/knex'` will still work as before

fabis94: image

fabis94 (approved):
there's still some relative paths in tests and duplicated DB queries that could be put in the repo layer, but for the most part it looks alright

packages/server/modules/core/repositories/users.ts
alemagio: oh you meant renaming it in the import 
I thought you meant in the export.
Anyway I think I found a good compromise

PR #2499, opened 2024-07-11
feat(workspaces): assign project roles to new users in a workspace

cdriesler:
## Description & motivation

- We are continuing to use project/stream roles for authorization purposes for projects in a workspace
- These roles are implied by, and need to be kept in sync with, a user's workspace role

TODO: It is unclear if project roles should always be kept in sync with workspace roles. Can I own a stream I create in a workspace without being an admin? Or maybe only admins can create projects? What happens if they stop being an admin? etc etc

## Changes:

- Adds logic to the `setWorkspaceRole` service function to update workspace project roles on change
- Adds logic to the `deleteWorkspaceRole` service function to delete workspace project roles

## Checklist:


gjedlicska (changes requested):

cdriesler:
@gjedlicska done and back to you

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -102,3 +105,11 @@ export const upsertWorkspaceRoleFactory =
       .onConflict(['userId', 'workspaceId'])
       .merge(['role'])
   }
+
+export const getWorkspaceProjectsFactory =
```
gjedlicska, line 109: Not to be that guy, but this doesn't seam to have integration tests

packages/server/modules/workspaces/utils/mapWorkspaceRoleToProjectRole.ts
gjedlicska, line 1: This, arguably could be be `packages/server/modules/workspaces/DOMAIN/mapWorkspaceRoleToProjectRole.ts`

packages/server/modules/workspaces/repositories/workspaces.ts
gjedlicska: Also, i think it would be probably better to reuse the existing project repository function by extending it with an optional `workspaceId` argument. The benefit there would be, that the project list functionality will be consistent both in and outside of a workspace.

gjedlicska (changes requested):

packages/server/modules/workspaces/utils/mapWorkspaceRoleToProjectRole.ts
fabis94: Also: Why such an uber specific module name? One function per util module seems overly granular. I'd just have `roles.ts` with various role related helpers

packages/server/modules/workspaces/utils/mapWorkspaceRoleToProjectRole.ts
```diff
@@ -0,0 +1,16 @@
+import { StreamRoles, WorkspaceRoles } from '@speckle/shared'
+
+/**
+ * Given a user's workspace role, return the role they should have for workspace projects.
+ */
+export const mapWorkspaceRoleToProjectRole = (
+  workspaceRole: WorkspaceRoles
+): StreamRoles => {
+  switch (workspaceRole) {
+    case 'workspace:guest':
```
fabis94, line 10: avoid string literals, use the `Roles` const from @speckle/shared

cdriesler: roger, done

packages/server/modules/workspaces/repositories/workspaces.ts
cdriesler: no, fair. I like extending the existing repo functions, but I'm not thrilled by the two options (I found):

- `getDiscoverable` is too focused on _public_ projects
- `getUser` requires a user, which we can do here but may not always be the case

I gave it a swing, though. What do you think? If you're ok with this strategy I'll come back in and add the tests.

packages/server/modules/workspaces/utils/mapWorkspaceRoleToProjectRole.ts
cdriesler: moved and consolidated/renamed. just a personal bias for one-function-one-file when I'm making "utils" but no strong opinion in this case. `roles.ts` looks quite nice with the other `domain` files

packages/server/modules/workspaces/services/workspaceProjects.ts
```diff
@@ -0,0 +1,31 @@
+import { StreamRecord } from '@/modules/core/helpers/types'
```
fabis94, line 1: Do `workspaceProjects` and `workspaceProjectRoleCreation` really need to be 2 different services? It's kind of subjective, so if you think they do, then we can leave it at that, but I'd like to avoid having dozens of tiny services that only have 1 function in them

packages/server/modules/workspaces/services/workspaceProjects.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    do {
+      const { streams: workspaceProjects, cursor: pageCursor } = await getUserStreams({
+        userId,
+        workspaceIdWhitelist: [workspaceId],
+        cursor
+      })
+      projects.push(...workspaceProjects)
+      cursor = pageCursor ?? undefined
+    } while (!!cursor)
```
fabis94, line 28: 1. while/do-while is generally scary and should always have some kind of failsafe, e.g. always breaking after X loops. X can be something practically unreachable like 500 pages of projects, but at the very least it ends somewhere instead of executing indefinitely
2. just infinitely batching up all projects as long as there are more is likely to be a performance bomb - there could be a shit ton of projects OR there could be a bug in the repo query causing an infinitely increasing array of projects. instead, use JS generators and yield each page individually so that we never have to keep more than 1 page in memory - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator

packages/server/modules/workspaces/tests/unit/services/workspaceProjectRoleCreation.spec.ts
```diff
@@ -0,0 +1,52 @@
+import { StreamAclRecord, StreamRecord } from '@/modules/core/helpers/types'
+import { WorkspaceAcl } from '@/modules/workspacesCore/domain/types'
+import { grantWorkspaceProjectRolesFactory } from '@/modules/workspaces/services/workspaceProjectRoleCreation'
+import { Roles } from '@speckle/shared'
+import { expect } from 'chai'
+import cryptoRandomString from 'crypto-random-string'
+
+describe('Workspace project role services', () => {
+  describe('grantWorkspaceProjectRolesFactory creates a function, that', () => {
```
fabis94, line 9: Do we really need the test files to be this granular? There is an overheard to each test file - initializing a new test context, running `beforeEachContext` etc.

Some if not all of these can probably be merged into the same test file

fabis94: Ah, this is a unit test, so its probably lightweight enough to stand on its own. That said, I'd still evaluate if some of the tests can't be merged just to avoid the overheard there is and also be able to reuse more of test setup, without having to duplicate it

fabis94 (changes requested):

packages/server/modules/workspaces/tests/unit/services/workspaceProjectRoleCreation.spec.ts
gjedlicska: Its a unittest, being verbose in the test context is actually good.

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -0,0 +1,32 @@
...
+  const grantWorkspaceProjectRoles = grantWorkspaceProjectRolesFactory({
+    getWorkspaceRoles: getWorkspaceRolesFactory({ db }),
+    // TODO: Instantiate via factory function
+    grantStreamPermissions
+  })
+
+  await grantWorkspaceProjectRoles({ projectId, workspaceId })
+}
+
+export function initializeEventListener() {
```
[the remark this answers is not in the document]
cdriesler: is better, done

packages/server/modules/workspaces/tests/unit/services/workspaceProjectRoleCreation.spec.ts
cdriesler: will push back on this one - I think the verbosity/overhead is worth it here, but will keep this in mind going forwards

cdriesler:
@fabis94 Back to you. No strong opinions on module names/boundaries here, just a personal default bias for smaller and more numerous files. I think this is now closer to what you're suggesting and to the kind of thing I see in other modules.

packages/server/modules/workspaces/services/management.ts
```diff
@@ -0,0 +1,187 @@
...
+
+    // Delete workspace project roles
+    const projectIds: string[] = []
+    for await (const project of queryAllWorkspaceProjectsFactory({ getStreams })(
+      workspaceId
+    )) {
+      projectIds.push(project.id)
+    }
+    await Promise.all(
+      projectIds.map((streamId) => revokeStreamPermissions({ streamId, userId }))
```
fabis94, line 116: I mean this works, but ideally we'd want the generator to return the projects in batches, instead of one by one, so that you can then also `revokeStreamPermissions` in batches. As opposed to now collecting potentially hundreds or thousands of IDs and then invoking hundreds/thousands of concurrent promises that may kill the server.

```ts
for await (const projectsPage of projectPageGenerator) {
  await Promise.all(projectsPage.map(p => revokeStreamPermissions({ streamId: p.id, userId })))
}
```

That said, this is already better than what was here before

packages/server/modules/workspaces/services/management.ts
```diff
@@ -0,0 +1,187 @@
...
+    )
+
+    // Emit deleted role
+    emitWorkspaceEvent({ eventName: WorkspaceEvents.RoleDeleted, payload: deletedRole })
```
fabis94, line 120: Isn't this async? You should await these

packages/server/modules/workspaces/services/projects.ts
```diff
@@ -0,0 +1,37 @@
...
+        orderBy: null,
+        limit: 1000,
+        visibility: null,
+        searchQuery: null,
+        streamIdWhitelist: null,
+        workspaceIdWhitelist: [workspaceId]
+      })
+
+      for (const stream of streams ?? []) {
+        yield stream
```
fabis94, line 31: as mentioned above, i'd yield the entire batch so you can work with it in a more optimized way

packages/server/modules/workspaces/tests/unit/services/management.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+        role: Roles.Workspace.Member
+      }
+      const workspaceRoles: WorkspaceAcl[] = [workspaceRole]
+      const workspaceProjects: StreamRecord[] = [{ id: projectId } as StreamRecord]
+
+      let projectRoles: StreamAclRecord[] = [
+        { userId, role: Roles.Stream.Contributor, resourceId: projectId }
+      ]
+
+      const deleteWorkspaceRole = deleteWorkspaceRoleFactory({
```
fabis94, line 214: Figuring out how DRY tests should be is an art form in and of itself - if you go too DRY you can run into way too abstract tests that are too hard to understand and change (e.g. `modules/comments/tests/comments.graph.spec.js`). That said, copy and pasting the same setup over and over again when it's expected to be the same for most if not all test cases, even if it does change in the future, is going too far in the other direction - just changing the deps a little bit for the role factory will require the dev to go and make a dozen identical changes to all factory invocations across this test file.

I definitely want to push back about this assumption that DRY is not a concern in tests - it most definitely is, you just should not go overboard with it and make it "too clever" and you have to evaluate which bits of code are duplicate only currently and very likely to differ in the future and which are more or less always going to be the same. In this case there's so much duplication of common init, it could easily be extracted to a helper at the top of the file. In the case of factory init, you can also always provide a param that allows you to override some of the dependencies, if needed, while all of the other ones stay standard.

For examples on how I've approached this you can check `automations.spec.ts` or `trigger.spec.ts`

fabis94 (changes requested):

packages/server/modules/workspaces/services/management.ts
cdriesler: I went back and forth but this is a very good reason to go by page, done

packages/server/modules/workspaces/services/management.ts
cdriesler: ope, yes

packages/server/modules/workspaces/tests/unit/services/management.spec.ts
cdriesler: (discussed in person but some late thoughts) I think this has landed somewhere much better, lmk what you think, and ty for the push

fabis94 (approved):

gjedlicska (dismissed):

gjedlicska (approved):

PR #2501, opened 2024-07-12
decoupled inter-module communication & third party (NPM) module support

fabis94:
Read 1-api.ts first, then 2-apiFilters.ts and then the readme.


gjedlicska:
we'll pick up the plugin api at some point, so far its not needed

PR #2506, opened 2024-07-15
Fix: Changed date formatting to be clearer

Mikehrn:
This changes the day formatting to be a bit clearer, following the rules laid out in this ticket: https://linear.app/speckle/issue/WEB-1290/improve-our-time-ago-date-formatting

I've moved some of the date logic to the plugin so it's easier to consistently use across various places and made some minor changes to styling (mainly removing bold text).

We've decided to go for the 24-hour time format now, however as it's now all defined in the plugin this would be easy to change to AM/PM if we want in the future.


benjaminvo (changes requested):
Dates don't match :). Also I'd like a `,` between the year and timestamp in the tooltip so it's `Jan 25, 2023, 11:01`.

CleanShot 2024-07-15 at 21 47 33@2x
CleanShot 2024-07-15 at 21 47 23@2x

packages/frontend-2/plugins/dayjs.ts
```diff
@@ -2,9 +2,51 @@ import dayjs from 'dayjs'
...
 import duration from 'dayjs/plugin/duration.js'
+import updateLocale from 'dayjs/plugin/updateLocale'
 
 export default defineNuxtPlugin(() => {
   dayjs.extend(relativeTime)
   dayjs.extend(localizedFormat)
   dayjs.extend(duration)
+  dayjs.extend(updateLocale)
+
+  const customRelativeTime = (date: string): string => {
```
Mikehrn, line 13: DayJS has build-in options to change the formatting of dates, however there are a few issues with it

- There is a default threshold of 28 days for it to become "1 month ago", this threshold is adjustable however it's kind of buggy, quite regularly it would apply the threshold after already rendering. This can be fixed by removing the original threshold first, and applying the new one, however this was really hacky

- For any date bigger than 14 days we just want to output the formatted days, and not relative time, this could be solved with custom formatters, however that required quite a couple extra lines, as well as additional checks for rendering the "ago" text.

Because our use case is quite simple I decided to just create a wrapper function that still relies on DayJS for the calculations, but just handles the custom formattings, as it's a much less hacky/shorter solution

packages/frontend-2/plugins/dayjs.ts
```diff
@@ -7,4 +7,44 @@ export default defineNuxtPlugin(() => {
...
+    return unit.includes('second') || unit.includes('minute') || unit.includes('hour')
+  }
+
+  const getFullDate = (input: string) => dayjs(input).format('MMM D, YYYY, H:mm')
+  const getTrunicatedRelativeDate = (input: string) => customRelativeTime(input)
+  const getTrunicatedRelativeDateWithPrefix = (input: string) =>
+    isClockUnit(input) ? customRelativeTime(input) : `on ${customRelativeTime(input)}`
+
+  return {
+    provide: {
```
fabis94, line 44: Not a fan of this being a nuxt plugin - this means you can only ever use these in Vue scope or you have to get these from the NuxtApp instance.

Since the dayjs import is a singleton (you can import dayjs(), you don't need to work with this specific instance in this plugin), you could just have all of these functions be importable helpers. Better yet - if you define these somewhere under the `utils/` folder, they're gonna be auto-imported everywhere so that you don't even need to type out any imports.

Then you also don't need to have the dollar sign prefix for these which is kind of annoying too

packages/frontend-2/plugins/dayjs.ts
```diff
@@ -7,4 +7,44 @@ export default defineNuxtPlugin(() => {
...
+
+    return 'Just now'
+  }
+
+  const isClockUnit = (date: string) => {
+    const unit = customRelativeTime(date)
+    return unit.includes('second') || unit.includes('minute') || unit.includes('hour')
+  }
+
+  const getFullDate = (input: string) => dayjs(input).format('MMM D, YYYY, H:mm')
```
fabis94, line 38: You should add a JSDoc block above these helpers, cause the names aren't 100% self-explanatory. Maybe specify an example in there or something.

packages/frontend-2/plugins/dayjs.ts
```diff
@@ -7,4 +7,44 @@ export default defineNuxtPlugin(() => {
...
+  const getTrunicatedRelativeDate = (input: string) => customRelativeTime(input)
+  const getTrunicatedRelativeDateWithPrefix = (input: string) =>
```
fabis94, line 40: 1. Trunicated - misspelling
2. `input: string` should be optional, because often you want dayjs() to work with the current date. Also, it can be way more than a string - the dayjs() factory also accepts numbers, other Dates, other Dayjs instances. Make the input type the same one that the factory accepts: `dayjs.ConfigType`
3. `getTruncatedRelativeDateWithPrefix` doesn't have to be its own function, you can just add that as an optional param in the other function - `getTruncatedRelativeDate(input: string, options?: Partial<{prefix: boolean}>)`
4. These names are too long IMO. I'd just have `fullDate()` and `relativeDate()`

packages/frontend-2/components/viewer/compare-changes/Version.vue
```diff
@@ -4,27 +4,22 @@
...
       {{ isNewest ? 'New' : 'Old' }} Version
     </div>
   </div>
 </template>
 <script setup lang="ts">
-import dayjs from 'dayjs'
 import type { ViewerModelVersionCardItemFragment } from '~~/lib/common/generated/gql/graphql'
 
-const props = defineProps<{
+defineProps<{
```
fabis94, line 21: It's still often preferable to keep the results in a computed for caching. If you just use the function directly, it's going to be re-invoked every time the template re-renders, even if the date didn't actually change.

fabis94 (changes requested):

packages/frontend-2/plugins/dayjs.ts
Mikehrn: Sure I can move them to the utils folder 👍

packages/frontend-2/plugins/dayjs.ts
Mikehrn: @fabis94 Updated :)

packages/frontend-2/plugins/dayjs.ts
Mikehrn: Added 👍

packages/frontend-2/components/viewer/compare-changes/Version.vue
Mikehrn: I was under the assumption generally these components don't re-render unless the information actually is changed, but I'll change it back to computed 👍

fabis94: @Mikehrn Some other info may have updated and triggered a re-render, the date may still be the same. But since you're invoking a function in the rendering pipeline it's going to re-execute every time a re-render occurs.

https://vuejs.org/guide/essentials/computed.html#computed-caching-vs-methods

Mikehrn: Yeah I understand that, I was just under the impression that if the info of the model changes in any case the date also changes making it a bit redudant. But in this case it's actually a createdAt instead of updatedAt so it will actually always be the original date; in any case I'll change these back to computed 👍

fabis94 (approved):

packages/frontend-2/utils/dateFormatter.ts
```diff
@@ -0,0 +1,79 @@
+import type { ConfigType } from 'dayjs'
+import dayjs from 'dayjs'
+import relativeTime from 'dayjs/plugin/relativeTime'
+import localizedFormat from 'dayjs/plugin/localizedFormat'
+import duration from 'dayjs/plugin/duration.js'
+
+dayjs.extend(relativeTime)
```
fabis94, line 7: move back to plugin, or if you left the plugin as is, just remove these calls

fabis94 (changes requested):

packages/frontend-2/utils/dateFormatter.ts
```diff
@@ -0,0 +1,72 @@
...
+ * @example
+ * Assuming today is January 1st 2024
+ * formattedRelativeDate('2023-12-01') - returns "Dec 12, 2023"
+ * formattedRelativeDate('2023-12-01', { prefix: true }) - returns "on Dec 12, 2023"
+ * formattedRelativeDate('2023-12-31') -  returns "1 day ago"
+ * formattedRelativeDate('2023-12-31', { prefix: true }) -  returns "1 day ago"
+ */
+export const formattedRelativeDate = (
+  date: ConfigType,
+  options?: Partial<{ prefix: boolean; capitalize: true }>
```
Mikehrn, line 63: I added the `capitalize` option because in some cases (when there is no "Updated" in front of the date) we want to capitalize "Just now". We can't use TW class for this because `capitalize` would also change "2 minutes ago" to "2 Minutes ago"

packages/frontend-2/components/viewer/compare-changes/Version.vue
Mikehrn: Updated them

fabis94 (approved):

benjaminvo (approved):

PR #2507, opened 2024-07-15
Fix: Adjust sizes in invite dialog

Mikehrn:
Some small adjustments in the invite to project dialog to align the sizes. Button feels a bit big but it's the only one we have atm that matches the dropdown input. We can probably revisit this when we update the ui components.

Screenshot 2024-07-15 at 19 01 03
Screenshot 2024-07-15 at 18 52 47
Screenshot 2024-07-15 at 18 58 02


benjaminvo (approved):
I like it!

PR #2509, opened 2024-07-16
Fix: Updated formatting of text throughout the app

Mikehrn:
Had some time while waiting for a pipeline to finish so I took a look at the text formatting because it was a bit inconsistent. 

Now we've decided to update the text formatting (mainly capitalization) I went through the app and adjusted the copy to consistently use the new way of formatting (only capitalizing the first letter of a sentence) and fixed some typos. Should fix the text formatting in most of the app. Only touches copy. 

Ignored the settings modal as I've adjusted the text formatting in the other PR

### DIFF
--- a/packages/frontend-2/components/auth/LoginPanel.vue
+++ b/packages/frontend-2/components/auth/LoginPanel.vue
@@ -63,7 +63,7 @@ const props = withDefaults(
   }>(),
   {
     dialogMode: false,
-    title: 'Speckle Login',
+    title: 'Speckle login',
     subtitle: 'Connectivity, Collaboration and Automation for 3D'
   }
 )
--- a/packages/frontend-2/pages/downloads.vue
+++ b/packages/frontend-2/pages/downloads.vue
@@ -1,13 +1,13 @@
 <template>
   <div>
     <Portal to="navigation">
-      <HeaderNavLink :to="'/downloads'" name="Speckle Connectors"></HeaderNavLink>
+      <HeaderNavLink :to="'/downloads'" name="Speckle connectors"></HeaderNavLink>
     </Portal>
     <div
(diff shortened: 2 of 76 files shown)


packages/frontend-2/components/auth/LoginPanel.vue
```diff
@@ -63,7 +63,7 @@ const props = withDefaults(
   }>(),
   {
     dialogMode: false,
-    title: 'Speckle Login',
+    title: 'Speckle login',
     subtitle: 'Connectivity, Collaboration and Automation for 3D'
```
Mikehrn, line 67: Wasn't sure if we wanted to change this tagline, or we use it across different places with this formatting

benjaminvo: Yeah let's just keep this as is for now

packages/frontend-2/pages/downloads.vue
```diff
@@ -1,21 +1,21 @@
...
+        <h5 class="h4 font-bold">Connector downloads</h5>
         <div class="text-sm text-foreground-2 max-w- max-w-sm">
           Most of our connectors are available through Speckle Manager. You can also
           direct download the individual installers below.
         </div>
       </div>
       <div>
         <FormButton size="lg" full-width @click="showManagerDownloadDialog = true">
-          Download Manager
+          Download manager
```
benjaminvo, line 18:
```suggestion
          Download Manager
```
Should be uppercase. Manager is a product name, like "Speckle" or "Revit".

benjaminvo (changes requested):
Veeery nice. I noticed one tiny thing!

packages/frontend-2/pages/downloads.vue
Mikehrn: Got it! Updated :)

benjaminvo (approved):

PR #2510, opened 2024-07-16
chore(comments): update comments module to typescript

cdriesler:
⚠️ IMO this PR is exempt from the latest "prioritize PRs over everything else" directive. Review at your leisure and I'll ping when it's getting stale/becomes urgent.

## Description & motivation

- Multi-region looms in the distance, with a mountain of refactor work to be done
- I'm not tired and have some space between other work

## Changes:

- Moves comment module files to .ts
- Ignored test files and migrations. I'll get them too but they're they failed the pain-priority math for tonight

## To-do before merge:

- How does `SmartTextEditorValueSchema` interact with the `string` typing on `CommentRecord`? Looks like a later change that .js didn't care about but .ts does? It appears we want `SmartTextEditorValueSchema` in all cases now, but there were just enough `string` references that I wasn't sure.
- A few queries were not typed (at the gql level) to provide filter args. I defaulted their values to `false` because that's what the missing values (often written as `(...input.filter || {})`) would resolve to
- Several resolvers did not strictly assert that optional/nullable values existed. I opted for non-null assertions here, even though they're icky, instead of throwing errors. This was to limit changes to business logic

## Checklist:


cdriesler:
> @cdriesler - is it possible to do a `git mv x.js x.ts` to retain the git history of the original files? At the moment it looks as if the existing has been deleted, and the typescript is brand new without any history.

I did this, fwiw! Was going to ask in standup how to fix the diff because I agree this is not really reviewable as-is. Separate commits maybe?

cdriesler:
pls hold

PR #2517, opened 2024-07-18
chore(core): eslint rule to avoid relative imports

alemagio:
## Checklist:

### DIFF
--- a/packages/server/eslint.config.mjs
+++ b/packages/server/eslint.config.mjs
@@ -38,6 +38,12 @@ const configs = [
       }
     },
     rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: ['.*']
+        }
+      ],
       '@typescript-eslint/no-explicit-any': 'error',
       '@typescript-eslint/no-unsafe-return': 'error',
       '@typescript-eslint/no-base-to-string': 'off',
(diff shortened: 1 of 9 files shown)


packages/server/eslint.config.mjs
```diff
@@ -38,6 +38,12 @@ const configs = [
       }
     },
     rules: {
+      'no-restricted-imports': [
+        'error',
+        {
+          patterns: ['.*']
```
fabis94, line 44: How's this work? Doesn't a `.*` regex resolve to literally any kind of string?

fabis94 (changes requested):

packages/server/eslint.config.mjs
fabis94: Ah, these can be glob patterns too apparently. Well, in any case, there's some paths that need fixing picked up by pre-commit

fabis94: @iainsproat That's alright, we can add an eslint-disable comment there, cause that's an exception scenario.

packages/server/eslint.config.mjs
```diff
@@ -38,6 +38,12 @@ const configs = [
       }
     },
     rules: {
+      'no-restricted-imports': [
```
fabis94, line 41: Why is this only assigned to '*.ts' files?

alemagio: I assume at some point we will move all the files to .ts right?
Do we want to enforce the rule also in js files?

fabis94 (approved):

PR #2519, opened 2024-07-18
Alessandro/web 1172 change the users repository to abstract the email field in

alemagio:
## Description & motivation

In an effort to reduce PR size this is the first PR related to this user story.
It can be merged safely even if the work is not complete.

- updated the `getUsers` and `listUsers` functions to take `email` and `verified` fields from the `user_emails` table.
- create migration to backfill `user_emails`

## Checklist:


packages/server/modules/core/migrations/20240710154658_user-emails-backfill.ts
```diff
@@ -0,0 +1,26 @@
+import { UserRecord } from '@/modules/core/helpers/types'
```
fabis94, line 1: All of our migration files follow snake_case naming convention

packages/server/modules/core/migrations/20240710154658_user-emails-backfill.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Knex } from 'knex'
+import crs from 'crypto-random-string'
+
+export async function up(knex: Knex): Promise<void> {
+  const users = await knex<UserRecord>('users').select(['id', 'email', 'verified'])
```
fabis94, line 6: You definitely need to do batching here, you're just retrieving all users at once into memory, while there could be thousands of them.

Pull them in batches of ~1000 or something

packages/server/modules/core/migrations/20240710154658_user-emails-backfill.ts
```diff
@@ -0,0 +1,26 @@
...
+          primary: true
+        }))
+      )
+      .onConflict(['userId', 'email'])
+      .ignore()
+  }
+}
+
+export async function down(knex: Knex): Promise<void> {
+  await knex('user_emails').delete()
```
fabis94, line 25: That's not a valid down migration. Do the same thing you did in `up()` but in reverse.

packages/server/modules/core/repositories/users.ts
```diff
@@ -51,16 +51,30 @@ export async function getUsers(
   userIds = isArray(userIds) ? userIds : [userIds]
 
   const q = Users.knex<UserWithOptionalRole[]>().whereIn(Users.col.id, userIds)
+  q.leftJoin(
+    USER_EMAILS_TABLE_NAME,
```
fabis94, line 55: At this point I don't see why you shouldn't create and use table helpers similar to what we have for all other tables. You're manually templating column strings which has no type checking whatsoever.

Either use `buildTableHelper()` to create a type-checked helper for the table, or use the approach Gergo showed for type-checked column names (`db<Type>().where('col', val) // 'col' should be type checked). The table helper has nicer DX easily.

packages/server/modules/core/repositories/users.ts
```diff
@@ -51,16 +51,30 @@ export async function getUsers(
...
+    `${USER_EMAILS_TABLE_NAME}.userId`,
+    Users.col.id
+  ).where({ primary: true })
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
fabis94, line 60: I assume you want all columns besides email and verified, that's why you're doing this? In that case just use lodash's `omit()`. An approach that requires you to disable eslint rules, while clever maybe, is not very clean

packages/server/modules/core/repositories/users.ts
```diff
@@ -51,16 +51,30 @@ export async function getUsers(
   userIds = isArray(userIds) ? userIds : [userIds]
 
   const q = Users.knex<UserWithOptionalRole[]>().whereIn(Users.col.id, userIds)
+  q.leftJoin(
+    USER_EMAILS_TABLE_NAME,
+    `${USER_EMAILS_TABLE_NAME}.userId`,
+    Users.col.id
+  ).where({ primary: true })
```
fabis94, line 58: If you use the DB table helper, all column names are always prefixed with an alias/table name to avoid ambiguity about which table does the column belong on. There may only be 1 `primary` col now, but that's some nice future proofing you get out of the box.

packages/server/modules/core/repositories/users.ts
```diff
@@ -51,16 +51,30 @@ export async function getUsers(
...
-      // Getting first role from grouped results
-      knex.raw(`(array_agg("server_acl"."role"))[1] as role`)
-    ])
-    q.leftJoin(ServerAcl.name, ServerAcl.col.userId, Users.col.id)
-    q.groupBy(Users.col.id)
+    // Getting first role from grouped results
+    columns.push(knex.raw(`(array_agg("server_acl"."role"))[1] as role`))
   }
 
+  q.columns(columns)
```
fabis94, line 72: 1. ServerAcl only needs to be joined if `withRole`
2. Why group on email and verified too?

packages/server/modules/core/repositories/users.ts
```diff
@@ -94,16 +108,24 @@ export async function listUsers({
   cursor: Date | null
 } & UserQuery): Promise<UserWithRole[]> {
   const sanitizedLimit = clamp(limit, 1, 200)
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const { verified, email, ...userCols } = Users.col
```
fabis94, line 112: All of the same questions I have about the repo fn above

fabis94 (changes requested):

packages/server/modules/core/repositories/users.ts
alemagio: 1. I didn't change that logic; do you think this should be changed?
2. email and verified are grouped because it is a one-to-many relationship. Giving the `primary=true` condition makes it a unique connection but I still get an error running it.

packages/server/modules/core/repositories/users.ts
```diff
@@ -136,19 +140,28 @@ export async function getUserByEmail(
   email: string,
   options?: Partial<{ skipClean: boolean; withRole: boolean }>
 ) {
-  const q = Users.knex<UserWithOptionalRole[]>().whereRaw('lower(email) = lower(?)', [
-    email
-  ])
+  const userEmail = await db<UserEmail>(UserEmails.name)
```
fabis94, line 143: Why invoke this separate query? I'm sure its possible to have this function only invoke a single query to get the user by their email

packages/server/modules/core/services/users.js
```diff
@@ -156,20 +155,50 @@ module.exports = {
    * @returns {Promise<import('@/modules/core/helpers/types').UserRecord | null>}
    */
   async getUserById({ userId }) {
-    const user = await Users().where({ id: userId }).select('*').first()
+    const user = await Users()
```
fabis94, line 158: It really sucks that this service does so much repository-layer query building and while I get that all of that already was there before your changes, this PR is an opportunity to at least get rid of some of that. At the very least not make it worse, which is what's happening as the new queries you're introducing in the service are even bigger and more  complicated than what was there before.

Not just that but these seem to largely duplicate what's already existing in the repo layer and also what's existing in other functions in this same service.

So why not just make these service calls invoke the repo functions instead?

packages/server/modules/core/tests/integration/find-users.spec.ts
```diff
@@ -0,0 +1,103 @@
+import { describe } from 'mocha'
```
fabis94, line 1: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#0746f1c947b741788cf581b9e1659f51

packages/server/modules/core/tests/usersAdminList.spec.ts
```diff
@@ -190,7 +190,7 @@ describe('[Admin users list]', () => {
   })
 
   after(async () => {
-    await cleanup()
+    // await cleanup()
```
fabis94, line 193: why?

fabis94 (changes requested):

packages/server/modules/core/migrations/20240710154658_user_emails_backfill.ts
```diff
@@ -0,0 +1,35 @@
...
+          id: crs({ length: 10 }),
+          userId: user.id,
+          email: user.email,
+          verified: user.verified,
+          primary: true
+        }))
+      )
+      .onConflict(['userId', 'email'])
+      .ignore()
+  } while (users.length > 0)
```
fabis94, line 30: Always add failsafes for while/do-while loops, cause if you have a bug in there, you could run into an infinite loop. It can be an impractical limit that you would never actually reach, but it would at least ensure it stops somewhere - e.g. after a 100 pages.

This specific migration actually would've caused exactly that - you never increase the offset, so when this would run in prod, it would inifinitely run the same queries on the same set of first 1000 users

packages/server/modules/core/helpers/userHelper.ts
```diff
@@ -25,3 +26,23 @@ export function removePrivateFields(
 }
 
 export type { LimitedUserRecord, UserRecord }
+
+export const getUsersBaseQuery = (
+  query: Knex.QueryBuilder,
+  { searchQuery, role }: { searchQuery: string | null; role: string | null }
+) => {
+  if (searchQuery) {
+    query.where((queryBuilder) => {
```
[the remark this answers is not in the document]
alemagio: I understand and agree with you.
The problem is that this particular function is used once in the repo and some services.
The name is also misleading.
For now, I am trying to change the least possible amount of logic this PR has already introduced many changes.

Ideally, this function should not even exist outside of the repository because composing queries outside of the repository doesn't make any sense.
Putting this function inside the repository and exporting it means I should also add a domain type for it which is incorrect since that function is a utility but has no domain meaning.

packages/server/modules/core/helpers/userHelper.ts
```diff
@@ -25,3 +26,23 @@ export function removePrivateFields(
 }
 
 export type { LimitedUserRecord, UserRecord }
+
+export const getUsersBaseQuery = (
```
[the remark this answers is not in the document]
alemagio: See next comment

packages/server/modules/core/helpers/userHelper.ts
```diff
@@ -25,3 +26,23 @@ export function removePrivateFields(
...
+        .where((qb) => {
+          qb.where(UserEmails.col.email, 'ILIKE', `%${searchQuery}%`).where({
+            primary: true
+          })
+        })
+        .orWhere(Users.col.name, 'ILIKE', `%${searchQuery}%`)
+        .orWhere(Users.col.company, 'ILIKE', `%${searchQuery}%`)
+    })
+  }
+  if (role) query.where({ role })
```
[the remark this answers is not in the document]
alemagio: I joined two very similar functions, the only difference between the 2 was the role parameter.

packages/server/modules/core/repositories/users.ts
alemagio: The first query is only used to check if the email exists.
To do that without the email field in the `users` table we should make the join with the `user_emails` table, which we do, but only if necessary.
If the email is not found we just return null with no other operations.

packages/server/modules/core/services/users.js
alemagio: I agree with all this.
My only issue is that we have a deadline to respect, and refactoring this module looks like a long journey.
Also consider that we will need to change this logic anyway during the multi-region refactor.

packages/server/modules/core/tests/usersAdminList.spec.ts
alemagio: leftover

packages/server/modules/core/helpers/userHelper.ts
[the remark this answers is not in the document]
fabis94: > Putting this function inside the repository and exporting it means I should also add a domain type

Why? I don't see why that would be the case at all. Repository layer is for repository logic, that's all it means, there is no expectation that "everything in the repository layer must be of domain level importance"

packages/server/modules/core/repositories/users.ts
fabis94: @alemagio I mean, sure, but that doesn't answer my original question - why invoke this as a separate query when this could easily be one? Sure, the scenario when the email doesn't exist may be a simpler query, but then in all other scenarios the operation is twice as heavy as it needs to do 2 serial SQL queries, both the simple one and then a complex one right after.

Pretty much always it's preferable to use 1 query over 2, as long as the queries are relatively simple ones like the ones we have here

packages/server/modules/core/helpers/scanTable.ts
```diff
@@ -0,0 +1,17 @@
...
+  db
+}: {
+  db: Knex<TRecord>
+}) =>
+  async function* ({ tableName, batchSize }: { tableName: string; batchSize: number }) {
+    const [rowsCount] = await db(tableName).count()
+    const MAX_LIMIT = parseInt(rowsCount.count.toString())
+    let offset = 0
+
+    while (offset <= MAX_LIMIT) {
```
fabis94, line 13: Same issue I mentioned before with no failsafes for while/do-while

packages/server/modules/core/helpers/userHelper.ts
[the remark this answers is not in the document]
fabis94: Fine, leave it

packages/server/modules/core/migrations/20240710154658_user_emails_backfill.ts
```diff
@@ -0,0 +1,31 @@
...
+import { scanTableFactory } from '@/modules/core/helpers/scanTable'
+
+export async function up(knex: Knex): Promise<void> {
+  const batchSize = 1000
+
+  for await (const rows of scanTableFactory({ db: knex })({
+    tableName: 'users',
+    batchSize
+  })) {
+    if (rows.length) {
```
fabis94, line 12: You should add `debug` level logging here, so in case this gets stuck or is very long, you can at least get info about how many rows have been processed and how many are still pending

packages/server/modules/core/services/users.js
fabis94: Ok, as a compromise lets leave it for future changes, but its really not great that not only this already shitty service is not improved, but actually made more complicated. Maybe leave a TODO for cleanup

packages/server/modules/core/tests/integration/scanTable.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+import {
+  createRandomEmail,
+  createRandomPassword
+} from '@/modules/core/helpers/testHelpers'
+
+const scanTable = scanTableFactory<UserRecord>({ db })
+
+describe('Helpers scanTable @core', () => {
+  before(async () => {
+    await db(Users.name).delete()
```
fabis94, line 17: We have utils for truncating tables: `import { truncateTables } from '@/test/hooks'`

packages/server/modules/core/migrations/20240710154658_user_emails_backfill.ts
fabis94: Also: Maybe you should batch up inserts for improved speed? Just an idea, but maybe batch up around 50 inserts and insert them all at once with one call

fabis94 (changes requested):
For the purposes of getting this out quicker, I'm ok with leaving some of the refactoring for the future. That said, there are still some more minor comments that need addressing

packages/server/modules/core/helpers/scanTable.ts
```diff
@@ -0,0 +1,17 @@
...
+      yield db<TRecord>(tableName).limit(batchSize).offset(offset)
```
fabis94, line 14: This could miss out on some rows if there are any created after `.count()` but before `.select()`. Instead of looping until MAX_LIMIT, loop until the db query no longer returns any rows. And then on top of that, add a failsafe that's practically uncreachable (I dunno, batchSize * 1000) to ensure it stops somewhere even in broken/infinite loop scenarios.

fabis94 (dismissed):
Seems alright, altho the scanTable helper could fail to actually scan the full table in some scenarios

fabis94 (approved):

PR #2524, opened 2024-07-22
refactor(ui-components): define and use new font styles

andrewwallacespeckle:
## Description & motivation
Michal has compiled a new style guide for fonts. This consists of 2 new styles, body and heading, each with their own size variants such as `text-body-sm`.

One thing to note, the size in the new class does not correspond to the tailwind class. So the above class does not correspond to `text-sm`. 

Headings should always be `font-semibold`, so this class is part of `text-heading`, but `text-body` has no font weight defined. Each class also sets line height and letter spacing. 

As well as defining the new styles, I have went through and used them in fe2 in as many places as possible in the time. I concentrated on the heavier trafficked parts of fe2, but there are still areas where I didn't make the change. For all new development, we should try and use the new font style classes. If you are working on something that has the old classes, please update as you see it. 

In general, this change has reduced font size across the app. This will take a bit of getting used to, but will allow us to display more useful information to the user at the same time. 

## Changes:
- Add new heading style
- Add new body style
- Update usages in FE2 only
- Add new Storybook page listing these styles. 
- Reduced padding in some areas to match the new condensed font styles
- Replaced h1,h2,h3 etc with text-heading equivalent


packages/tailwind-theme/src/plugin.ts
```diff
@@ -242,28 +242,43 @@ export default plugin(function ({ addComponents, addBase }) {
...
-      '@apply font-bold bg-gradient-to-r from-blue-500 via-blue-400 to-blue-600 text-transparent bg-clip-text':
+      '@apply font-semibold bg-gradient-to-r from-blue-500 via-blue-400 to-blue-600 text-transparent bg-clip-text':
         {}
     },
-    // Font sizes
-    '.h1': {
-      '@apply text-5xl leading-10': {}
+    // New heading styles
+    '.text-heading-2xl, .h1': {
+      '@apply font-medium text-3xl leading-10 tracking-[-0.63px]': {}
```
Mikehrn, line 250: `tracking-[-0.63px]` I guess this is because we are converting % from Figma to px values?

Mikehrn (approved):
Looks a lot cleaner 😁

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,219 @@
...
+      size="lg"
+      name="search"
+      :custom-icon="MagnifyingGlassIcon"
+      color="foundation"
+      full-width
+      search
+      :show-clear="!!searchString"
+      placeholder="Search Invitations"
+      class="rounded-md border border-outline-3"
+      @update:model-value="debounceSearchUpdate"
```
fabis94, line 28: Use `useDebouncedTextInput` instead

packages/frontend-2/utils/dateFormatter.ts
```diff
@@ -65,7 +65,7 @@ export const formattedRelativeDate = (
   if (options?.prefix) {
     return isClockUnit(date)
       ? customRelativeTime(date, options?.capitalize)
-      : `on ${customRelativeTime(date)}`
+      : `${customRelativeTime(date)}`
```
fabis94, line 68: Why? What's the point of the `prefix` option then, if it doesn't add a prefix? If we no longer want the prefix, then we should remove the option instead of having a confusing option called `prefix` that doesn't actually add a prefix.

fabis94 (changes requested):
Code looks OK for the most part, I haven't actually validated what it looks like tho.

That said, there are a few minor issues I noticed

packages/frontend-2/pages/server-management/pending-invitations.vue
```diff
@@ -0,0 +1,219 @@
...
+const successfullyResentInvites = ref<string[]>([])
+const showInviteDialog = ref(false)
+
+const {
+  result: extraPagesResult,
+  fetchMore: fetchMorePages,
+  variables: resultVariables,
+  onResult,
+  loading
+} = useQuery(getInvitesQuery, () => ({
```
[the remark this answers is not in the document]
Mikehrn: This page should actually be removed, it moved to the modal, I think it accidentally got added after merging back `main` again

packages/frontend-2/utils/dateFormatter.ts
Mikehrn: Same here, we should have the `on` still in there

packages/tailwind-theme/src/plugin.ts
andrewwallacespeckle: Yeah, these pixel values were provided by Michal. Maybe one to come back to in future if we experience issues around browser base font changes.

packages/frontend-2/pages/server-management/pending-invitations.vue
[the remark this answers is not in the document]
andrewwallacespeckle: I have removed this file.

packages/frontend-2/pages/server-management/pending-invitations.vue
andrewwallacespeckle: Same file - removed

packages/frontend-2/utils/dateFormatter.ts
andrewwallacespeckle: There is a bug here, but I have reverted my change, and @Mikehrn will take a look at the bug in it's own branch

packages/frontend-2/components/common/prose/MarkdownDescription.vue
```diff
@@ -1,6 +1,6 @@
 <!-- eslint-disable vue/no-v-html -->
 <template>
-  <div :class="proseClasses" v-html="cleanReadmeHtml"></div>
+  <div class="prose-sm max-w-auto overflow-auto" v-html="cleanReadmeHtml"></div>
```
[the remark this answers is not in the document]
andrewwallacespeckle: Re-added proseClasses

packages/frontend-2/components/header/NavUserMenu.vue
```diff
@@ -17,113 +17,111 @@
...
+                Connector downloads
+              </NuxtLink>
+            </MenuItem>
+          </div>
           <MenuItem v-if="activeUser" v-slot="{ active }">
             <NuxtLink
               :class="[
-                active ? 'bg-foundation-focus' : '',
-                'flex gap-3.5 items-center px-3 py-2.5 text-sm text-foreground cursor-pointer transition mx-1 rounded'
+                active ? 'bg-highlight-1' : '',
```
fabis94, line 40: What is bg-highlight-1? Where does that come from?

fabis94 (approved):

PR #2527, opened 2024-07-23
Fix: Change server settings and discussion to usePaginatedQuery

Mikehrn:
Following the comment on the settings modal PR this PR changes the tables that use infinite loading to use `usePaginatedQuery`, as well as the discussions page. 

There are a few more places where this can be done (dashboard and models), I will address these in a separate PR when I have some time in between things or when we touch these pages, which will be soon.

(also fixes some minor styling issues that I noticed in the search bar on the settings pages)

### DIFF
--- a/packages/frontend-2/components/project/page/discussions/Results.vue
+++ b/packages/frontend-2/components/project/page/discussions/Results.vue
@@ -43,7 +43,7 @@ const {
     projectId: props.project.id,
     filter: { includeArchived: !!props.includeArchived }
   })),
-  resolveKey: (vars) => [vars.projectId, vars.filter.includeArchived],
+  resolveKey: (vars) => [vars.projectId, vars.filter],
   resolveCurrentResult: (res) => res?.project.commentThreads,
   resolveNextPageVariables: (baseVars, cursor) => ({
     ...baseVars,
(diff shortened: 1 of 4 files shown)


packages/frontend-2/components/project/page/discussions/Results.vue
```diff
@@ -39,61 +33,26 @@ const props = defineProps<{
...
+  identifier,
+  onInfiniteLoad,
+  query: { result }
+} = usePaginatedQuery({
+  query: latestCommentThreadsQuery,
+  baseVariables: computed(() => ({
+    projectId: props.project.id,
+    filter: { includeArchived: !!props.includeArchived }
+  })),
+  resolveKey: (vars) => [vars.projectId],
```
fabis94, line 46: Just the projectId isn't enough - vars.filter being different will also result in a different query. If you leave vars.filter outside of the key, a change to the filters won't reset the pagination state, a new query won't be invoked etc.

fabis94 (changes requested):
Looks good for the most part, I did find 1 issue tho

packages/frontend-2/components/project/page/discussions/Results.vue
Mikehrn: Good point, updated it :)

fabis94 (dismissed):

Mikehrn:
@fabis94 I actually had to change it, there was a type error (I guess because it was an array with both a string and an object), I changed it now to be just an object (looking at the function, this should work?), functionally works good now

fabis94 (approved):

PR #2530, opened 2024-07-25
feat(server): serverInvites refactor + workspace invites CRUD & GQL API

fabis94:
Most of the changes actually relate to a serverInvites module refactor that was necessary to enable workspaces support without having to write any workspaces code in the `serverinvites` module. The algorithm is refactored to support any kind of arbitrary resource targets.

Besides that and workspace invites functionality, there's also various improvements to test utils (workspace test utils for easy workspace CRUD in tests), dataloaders (modularizable dataloaders) and the GQL schema (e.g. @hasWorkspaceRole directive)


packages/server/knexfile.js
```diff
@@ -86,14 +86,16 @@ const config = {
     connection: {
       connectionString: connectionUri || 'postgres://127.0.0.1/speckle2_test',
       application_name: 'speckle_server'
-    }
+    },
+    asyncStackTraces: true
```
fabis94, line 90: Can't believe we hadn't turned this on - this finally enables stack traces that properly tell you where the broken SQL query was triggered from

packages/server/modules/core/graph/resolvers/streams.ts
```diff
@@ -1,8 +1,5 @@
-'use strict'
-const { UserInputError } = require('apollo-server-express')
-const { withFilter } = require('graphql-subscriptions')
-
-const {
+import { UserInputError } from 'apollo-server-express'
```
fabis94, line 1: I didn't really want to, but had to refactor the streams resolvers to `.ts` cause there was a lot of Invites related code here and without TS linting there was no way to be sure that my extensive invite refactors hadn't broke anything

packages/server/modules/core/repositories/streams.ts
```diff
@@ -983,6 +987,19 @@ export async function revokeStreamPermissions(params: {
...
 
+  const existingPermission = await StreamAcl.knex()
+    .where({
+      [StreamAcl.col.resourceId]: streamId,
+      [StreamAcl.col.userId]: userId
+    })
+    .first()
+  if (!existingPermission) {
+    // User already doesn't have permissions
+    return await Streams.knex<StreamRecord>()
```
fabis94, line 998: This allows invoking the repo fn even if the user already doesn't have any kind of role. Useful in tests

packages/server/modules/serverinvites/domain/types.ts
```diff
@@ -1,22 +1,48 @@
...
-  inviterId: string
-  createdAt: Date
-  used: boolean
-  message: Nullable<string>
-  resourceTarget: typeof ResourceTargets.Streams | null
-  resourceId: Nullable<string>
-  role: Nullable<string>
-  token: string
-  serverRole: Nullable<string>
+export interface InviteResourceTargetTypeMap {
```
fabis94, line 8: Might seem unnecessary, but this allows for dynamic interface augmentation from the workspaces module

packages/server/modules/serverinvites/migrations/20240716134617_migrate_to_resources_array.ts
```diff
@@ -0,0 +1,144 @@
...
+  resourceTarget: string | null
+  resourceId: Nullable<string>
+  role: Nullable<string>
+  token: string
+  serverRole: Nullable<string>
+}
+
+export async function up(knex: Knex): Promise<void> {
+  // Iterate over all rows and update the resources field w/ empty resource
+  const qStream = knex
```
fabis94, line 36: Our largest server (app) has around ~2.7k invites right now

packages/server/modules/serverinvites/services/creation.ts
```diff
@@ -0,0 +1,215 @@
...
+    // collect and validate all resource targets
+    const resources = await collectAndValidateResourceTargets({
+      input: params,
+      inviter,
+      inviterResourceAccessLimits,
+      target: targetData,
+      targetUser,
+      serverInfo
+    })
+    const finalPrimaryResource = getPrimaryResourceTarget(resources)
```
fabis94, line 127: cc Gergo: It doesn't make any sense to save all collected resource targets in the DB. Not only does that complicate SQL queries, but it also enables a situation where we decide that project invites also should invite people to a new type of resource, but all of pre-existing project invites won't do it, because they resolved their targets at the moment of invite creation, not when accepting it

packages/server/modules/workspaces/graph/dataloaders/workspaces.ts
```diff
@@ -0,0 +1,36 @@
...
+import { defineRequestDataloaders } from '@/modules/shared/helpers/graphqlHelper'
+import { getWorkspacesFactory } from '@/modules/workspaces/repositories/workspaces'
+import { WorkspaceWithOptionalRole } from '@/modules/workspacesCore/domain/types'
+import { keyBy } from 'lodash'
+
+const { FF_WORKSPACES_MODULE_ENABLED } = getFeatureFlags()
+
+declare module '@/modules/core/loaders' {
+  interface AllRequestDataLoaders
+    extends Partial<ReturnType<typeof dataLoadersDefinition>> {}
```
fabis94, line 12: Making workspaces dataloaders optional, cause they are not gonna be there if the module isn't enabled

packages/server/test/plugins/graphql.ts
```diff
@@ -22,24 +23,52 @@ const graphqlChaiPlugin: Chai.ChaiPlugin = (_chai, utils) => {
...
+          if (matchMessage) {
+            msg = `Expected GraphQL response to have no errors containing "${matchMessage}", but found some`
+          } else {
+            msg = 'Expected GraphQL response to have no errors, but found some'
+          }
+
+          msg += `\nErrors: ${JSON.stringify(errorsArr, null, 2)}`
         }
+
+        throw new AssertionError(msg)
```
fabis94, line 71: The current error messages coming out from this assertion suck and don't actually tell you what the errors are

packages/server/test/assertionHelper.ts
```diff
@@ -11,3 +12,16 @@ export const expectToThrow = async (fn: () => MaybeAsync<any>) => {
 
   throw new AssertionError("Function was expected to throw, but didn't")
 }
+
+/**
+ * Create parameterizable test cases for each element in an array
+ */
+export const itEach = <T>(
```
fabis94, line 19: Finally! it.each support

packages/server/modules/workspaces/helpers/roles.ts
```diff
@@ -0,0 +1,56 @@
...
+  switch (workspaceRole) {
+    case Roles.Workspace.Guest:
+    case Roles.Workspace.Member:
+      return Roles.Stream.Reviewer
+    case Roles.Workspace.Admin:
+      return Roles.Stream.Owner
+  }
+}
+
+export const mapProjectRoleToWorkspaceRole = (
```
fabis94, line 32: Does this make sense? When a user gets invited to a workspace project, and has to be also added to the workspace once the invite is accepted - what kind of project role should they get?

packages/server/knexfile.js
fabis94: This is already only enabled for dev & test envs. And on top of that they do themselves say that its a small overhead.

packages/server/modules/shared/index.js
```diff
@@ -66,7 +66,7 @@ const getUserAclEntry = async ({ aclTableName, userId, resourceId }) => {
  * @param  {string | null | undefined} userId
  * @param  {string} resourceId
  * @param  {string} requiredRole
- * @param {import('@/modules/core/domain/tokens/types').TokenResourceIdentifier[] | undefined | null} [userResourceAccessLimits]
+ * @param {import('@/modules/core/domain/tokens/types').TokenResourceIdentifier[] | undefined | null} userResourceAccessLimits
```
fabis94, line 69: Making these harder to ignore

alemagio (dismissed):

packages/server/modules/workspaces/helpers/roles.ts
cdriesler: discussed a bit in standup the other day - I plan to spend a little time tomorrow writing down more specific requirements for project/workspace roles

packages/server/modules/core/services/streams/management.ts
```diff
@@ -88,12 +90,23 @@ export async function createStreamReturnRecord(
...
-        insertInviteAndDeleteOld: insertInviteAndDeleteOldFactory({ db })
-      })
+        insertInviteAndDeleteOld: insertInviteAndDeleteOldFactory({ db }),
+        collectAndValidateResourceTargets: collectAndValidateCoreTargetsFactory({
+          getStream
+        }),
+        buildInviteEmailContents: buildCoreInviteEmailContentsFactory({
+          getStream
+        }),
+        emitServerInvitesEvent: ({ eventName, payload }) =>
```
cdriesler, line 103: I've just been passing `emit` directly in cases like this. Something like:

```
const { emit } = getEventBus()
emitServerInvitesEvent: emit
```

I think this "does everything" event bus was created after we all discussed factory functions/domain types. So the naming here feels a bit silly to me. (`emitServerInvitesEventButReallyAnythingYouWant`)

So this is kind of an ignorable style nit but I had some thoughts while writing it out. Maybe we reconsider how these get created/injected.

packages/server/modules/emails/services/emailRendering.ts
```diff
@@ -113,3 +114,12 @@ const renderEmailText = async (
   )
   return fullText
 }
+
+/**
+ * Sanitize message that potentially has HTML in it
+ */
+export function sanitizeMessage(message: string, stripAll: boolean = false): string {
```
cdriesler, line 121: unexpected change, did something change with emails?

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -37,16 +52,59 @@ export const onProjectCreatedFactory =
     )
   }
 
+export const onInviteFinalizedFactory =
```
cdriesler, line 55: I think this is controversial/undiscussed behavior? Adding a user to a workspace when they're invited to a workspace project?

It makes sense to me but we definitely should get explicit somewhere that this is how we want things to happen. (I expected the flow to be "invite to workspace" => "have access to workspace projects")

cdriesler (changes requested):
Requested changes are small, plus a handful of clarifying comments. I think it's still worth you getting another eye on the refactor in general - I don't have enough history/context to object if there's something important there. All appears in order and working.

packages/server/modules/emails/services/emailRendering.ts
fabis94: Nah, this isn't anything new, it was moved from one of the deleted files

packages/server/modules/workspaces/events/eventListener.ts
fabis94: @cdriesler This is based on this figjam that Gergo did for me: https://www.figma.com/board/kJgZgku0oW6CfeKMe3HQfI/Workspace-invites?node-id=0-1&t=nUTTufUxovJVwbsp-0

Our reqs are strewn all over the place and not always formalized which is why we're not all on the same page about them haha

cdriesler: Totally fair, will try to collect notes today. Ready to approve after the generics change or something like this from you:

Creating_Bugs_Bunny's__No_

fabis94: @cdriesler ye, im workin through all of ur other comments rn 👍

cdriesler (dismissed):
🙏

cdriesler (approved):

PR #2534, opened 2024-07-26
feat(server): modularized mocks + workspace mocks for Mike

fabis94:


packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -2,6 +2,14 @@ extend type Query {
   workspace(id: String!): Workspace!
     @hasServerRole(role: SERVER_USER)
     @hasScope(scope: "workspace:read")
+
```
fabis94, line 5: I copied over all of my typedefs from the invites PR

cdriesler (dismissed):
This is very cool. Am I understanding correctly that some of the mocks that use `faker.datatype.boolean()` are written to fail randomly when called? For example:

```
            delete: () => {
              const val = faker.datatype.boolean()
              if (!val) {
                throw new Error('Fake workspace delete error')
              }

              return val
            }
```
Is this something we'd prefer the caller to be able to ask for/trigger reliably, even if it's faked? I like the approach for `Query.workspace`. Maybe this is something we're meant to address/let Mike etc. to decide when they start using the endpoints? No one is hitting this endpoint any time soon

fabis94:
@cdriesler Yes, the idea is to fail randomly. You're right that it would be better to be able fail deterministically (like what I've done for workspaceCreate, but I didn't want to spend too much time on this yet until its clear how Mike will use these. We will likely do what you've described afterwards.

packages/server/modules/shared/helpers/mocks.ts
```diff
@@ -0,0 +1,206 @@
...
+        store.set(info.parentType.name, parent.id, info.fieldName, val)
+      }
+
+      return val
+    }
+
+    return wrapperResolver
+  }
+
+  return {
```
fabis94, line 143: Pretty proud of these haha, these greatly simplify the entire mock definition experience and improve DX

Mikehrn (approved):
🚀

PR #2535, opened 2024-07-29
Feat: Add option to show modal not fullscreen on mobile

Mikehrn:
There were some issues with DUI 3 and the modal being always being fullscreen on sm screensizes, this adds the options to specificy the fullscreen options (default on mobile, optional to show on all, only desktop or none)

### DIFF
--- a/packages/ui-components/src/components/layout/Dialog.vue
+++ b/packages/ui-components/src/components/layout/Dialog.vue
@@ -16,7 +16,12 @@
       </TransitionChild>
       <div class="fixed top-0 left-0 z-10 h-screen !h-[100dvh] w-screen">
         <div
-          class="flex md:justify-center items-end md:items-center h-full w-full md:p-6"
+          class="flex md:justify-center h-full w-full md:p-6"
+          :class="[
+            fullscreen === 'none' || fullscreen === 'desktop'
+              ? 'p-4 items-center'
+              : 'items-end md:items-center'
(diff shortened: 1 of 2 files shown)


packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -214,6 +221,56 @@ const widthClasses = computed(() => {
...
+  () => props.fullscreen === 'desktop' || props.fullscreen === 'all'
+)
+
+const dialogPanelClasses = computed(() => {
+  const classParts: string[] = [
+    'transform md:rounded-xl text-foreground overflow-hidden transition-all bg-foundation text-left shadow-xl  flex flex-col md:h-auto'
+  ]
+
+  if (isFullscreenDesktop.value) {
+    classParts.push('md:h-full md:h-[98vh] md:!h-[98dvh]')
```
andrewwallacespeckle, line 234: Perfect! It looks a bit hacky, but i prefer this than using css

andrewwallacespeckle (approved):
Nice work. I didn't test this locally, but the code changes look good

packages/ui-components/src/components/layout/Dialog.vue
Mikehrn: It became very hard to read as inline within the template haha, I saw we used the pattern in other places so I switched it to this :)

PR #2540, opened 2024-07-29
feat(workspaces): workspace member role resolvers

cdriesler:
## Changes:

- Adds resolvers and tests for workspace member role update (create) and delete
- I know we discussed test setup overhead in a previous PR but, if we have `management` and `invites`, `roles` seemed like sufficiently different setup (and will have more tests later)


packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -46,13 +46,12 @@ type WorkspaceMutations {
     @hasScope(scope: "workspace:create")
   delete(workspaceId: String!): Boolean! @hasScope(scope: "workspace:delete")
   update(input: WorkspaceUpdateInput!): Workspace! @hasScope(scope: "workspace:update")
-  """
-  TODO: `@hasWorkspaceRole(role: WORKSPACE_ADMIN)` for role changes
-  """
-  updateRole(input: WorkspaceRoleUpdateInput!): Workspace!
+  updateRole(input: WorkspaceRoleUpdateInput!): Boolean!
```
fabis94, line 49: @cdriesler 

1. This should return Workspace not Boolean, so that the client (frontend) can immediately ask back for the updated Workspace.team without having to refetch https://www.apollographql.com/docs/react/data/mutations/#include-modified-objects-in-mutation-responses
2. Why not have the same resolver for update/delete like we do for projects?

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -46,13 +46,12 @@ type WorkspaceMutations {
...
     @hasScope(scope: "workspace:update")
-  deleteRole(input: WorkspaceRoleDeleteInput!): Workspace!
+    @hasWorkspaceRole(role: ADMIN)
+  deleteRole(input: WorkspaceRoleDeleteInput!): Boolean!
```
fabis94, line 52: I don't expect mocks to be broken (I don't see any new resolvers), but it would be great if you could enable them and validate that workspace queries still work (in Apollo Explorer in localhost:8081/graphql)

fabis94 (changes requested):

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
cdriesler: 1. Done, makes good sense to me
2. I did not catch that this was the pattern and like it much better, done

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
cdriesler: ty for nudge, will get in the habit of doing this! Went to do in now though and got this:

image

Will check in tmr about it

fabis94: @cdriesler That just happens when the server is restarting in the BG due to hotreload, so I don't expect this to be an actual issue

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -46,13 +49,9 @@ type WorkspaceMutations {
...
   delete(workspaceId: String!): Boolean! @hasScope(scope: "workspace:delete")
   update(input: WorkspaceUpdateInput!): Workspace! @hasScope(scope: "workspace:update")
-  """
-  TODO: `@hasWorkspaceRole(role: WORKSPACE_ADMIN)` for role changes
-  """
   updateRole(input: WorkspaceRoleUpdateInput!): Workspace!
     @hasScope(scope: "workspace:update")
-  deleteRole(input: WorkspaceRoleDeleteInput!): Workspace!
-    @hasScope(scope: "workspace:update")
+    @hasWorkspaceRole(role: ADMIN)
```
fabis94, line 54: This ain't gonna work, IIRC @hasWorkspaceRole only works when attached to fields directly on the `Workspace` type. You can check the implementation - it relies on resolving what the workspace that you're referring to is based on `parent.id`

packages/server/modules/workspaces/tests/integration/roles.graph.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+    const testWorkspace: BasicTestWorkspace = {
+      id: '',
+      ownerId: testAdminUser.id,
+      name: 'My Test Workspace'
+    }
+
+    await createTestWorkspace(testWorkspace, testAdminUser)
+
+    workspace = testWorkspace
```
fabis94, line 67: Why not just have `testWorkspace` be a const defined outside of `before()` next to `testAdminUser` and `testMemberUser`? This variable is redundant.

If you're worried about id being left empty, the `createTestWorkspace` call will fill it with the final ID. `ownerId` can be left as an empty string too, cause its gonna fill that based on the user you pass in too.

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -28,7 +28,10 @@ input WorkspaceUpdateInput {
 input WorkspaceRoleUpdateInput {
   userId: String!
   workspaceId: String!
-  role: WorkspaceRole!
+  """
+  Leave role null to revoke access entirely
+  """
+  role: WorkspaceRole
```
fabis94, line 34: The convention so far (e.g. for projects) has been to return the real/internal role string (e.g. "workspace:admin"). The main reason for these secondary GQL enums is just to have GQL level validation/type checking for input objects - e.g. when updating a role or whatever.

It is a bit confusing, I know haha. I'd be open to revisiting these role enums and possibly dumping them in favor of just always using the real role values.

packages/server/modules/workspaces/tests/integration/roles.graph.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+    it('should create a role if none exists', async () => {
+      const res = await apollo.execute(UpdateWorkspaceRoleDocument, {
+        input: {
+          userId: testMemberUser.id,
+          workspaceId: workspace.id,
+          role: mapMainWorkspaceRoleToGqlRole(Roles.Workspace.Admin)
+        }
+      })
+
+      const { data } = await apollo.execute(GetWorkspaceDocument, {
```
fabis94, line 90: Now that `UpdateRole` also returns the workspace, you no longer need this second `GetWorkspace` operation in your test cases (the ones below too)

fabis94 (changes requested):

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
cdriesler: ty for flagging I didn't realize that's how it worked. Using `authorizeResolver` now. In a previous PR, we opted to do this in the service function level, but here that go real messy real quick wrt testing and DI. Opted to do it in the resolver here (there's only one for roles now so it feels a bit better)

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
cdriesler: noted, definitely got these two twisted a few times while working too. but will do this going forwards

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -175,11 +171,41 @@ export = FF_WORKSPACES_MODULE_ENABLED
...
         },
-        updateRole: async () => {
-          throw new WorkspacesNotYetImplementedError()
-        },
-        deleteRole: async () => {
-          throw new WorkspacesNotYetImplementedError()
+        updateRole: async (_parent, args, context) => {
+          const { userId, workspaceId, role } = args.input
+
+          authorizeResolver(context.userId, workspaceId, Roles.Workspace.Admin, null)
```
fabis94, line 177: I'd still put this in the service layer, to be sure that this check always happens regardless of where you trigger it from. IMO its fine that it can't be mocked out/injected. That said, you can leave it as is for now.

An important fix that has to happen, however, is that you should not ever leave the 4th param of `authorizeResolver` empty, if possible. This is where those resource access rules from the GQL `context` should be injected to validate that the token has appropriate rights.

Make sure this is done in all relevant resolvers in this PR

fabis94 (changes requested):
It's pretty much ready to merge, just need to fix the incomplete authorization check

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
cdriesler: sorry sorry, this had been explained and I forgot. will keep in mind 👍

will move `authorizeResolver` in a follow-up, ty for letting this through

fabis94 (approved):

PR #2541, opened 2024-07-30
chore(workspaces): tests for workspace project creation

cdriesler:
NOTE: The tests are passing but we appear to have some gql validation _very_ upset that I removed an input field. Unsure what to do about this, will check in tmr.

## Description & motivation

We already indirectly implemented "workspace project creation" by adding the `workspaceId` param to stream and project creation resolvers in a previous PR. This finishes the job by bringing the change up to latest standards (code quality, style, and tests)

## Changes:

- Removes the possibility to do this through deprecated `streamCreate`. This feels more correct but I can bring it back.
- Adds tests and supporting queries
  - I think workspace projects is distinct enough from the other gql suites, but it's a bit silly in its own file rn with only one test. If the overhead is too much I'm happy to park it in another file for now and break it out when we have more tests for working with workspace projects.


fabis94:
@cdriesler I've sent out an invite to Apollo Studio, if you think the changes are safe you can mark them as such (remember to do so for all affected operations) and then once the CI reruns its gonna succeed

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -95,6 +95,7 @@ type Workspace {
   description: String
   createdAt: DateTime!
   updatedAt: DateTime!
+  logoUrl: String
```
fabis94, line 98: The logoUrl param in the WorkspaceCreateInput type doesn't make any sense. Whether we go with storing the base64 data string in the DB or store it in some kind of blob storage, the client is def not the one posting a logo URL to the server - they're gonna post the base64 string or post it as multiform data or whatever. The logo URL comes from the server (if the server creates one at all).

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -314,9 +315,18 @@ export = FF_WORKSPACES_MODULE_ENABLED
...
-          // Get projects in workspace
-          throw new WorkspacesNotYetImplementedError()
+        projects: async (parent, args) => {
+          const getWorkspaceProjects = getWorkspaceProjectsFactory({ getStreams })
+          return await getWorkspaceProjects(
+            {
+              workspaceId: parent.id
+            },
+            {
+              limit: args.limit ?? 25,
```
fabis94, line 325: `||` is in most cases better than `??` cause it'll use your fallback value not just when the 1st operand is `null` but when its generally falsy in any way, which is usually what you're after. Surely you want the fallback value irregardless of whether cursor is null, or `false` or an empty string or 0?

packages/server/modules/workspaces/services/projects.ts
```diff
@@ -33,3 +35,42 @@ export const queryAllWorkspaceProjectsFactory = ({
...
   }
+
+type GetWorkspaceProjectsArgs = {
+  workspaceId: string
+}
+
+type GetWorkspaceProjectsOptions = {
+  limit: number | null
+  cursor: string | null
+  filter: UserProjectsFilter | null
```
fabis94, line 46: Even if UserProjectsFilter has the same values you want right now, you shouldn't use it and should introduce your own type cause the semantics are different. The types being the same right now doesn't mean they're gonna be the same in the future. In the future the type may get other more user specific fields added to it, and then your GetWorkspaceProjects field is gonna ask for user-specific filtering fields.

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      context: createTestContext({
+        auth: true,
+        userId: testUser.id,
+        token,
+        role: testUser.role,
+        scopes: AllScopes
+      })
+    })
+
+    const { data } = await apollo.execute(CreateWorkspaceDocument, {
```
fabis94, line 57: Why not use the `createTestWorkspace` helper?

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      const projectName = cryptoRandomString({ length: 6 })
+
+      const createRes = await apollo.execute(CreateWorkspaceProjectDocument, {
+        input: {
+          name: projectName,
+          workspaceId: workspace.id
+        }
+      })
+
+      const getRes = await apollo.execute(GetWorkspaceProjectsDocument, {
```
fabis94, line 79: This query should have proper tests that also test pagination filtering etc.

fabis94 (changes requested):

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
cdriesler: +1, not sure where this came from, done

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
cdriesler: back and forth on this one but will prefer `||` going forwards, this is a good reason

packages/server/modules/workspaces/services/projects.ts
cdriesler: roger, done

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
cdriesler: I am silly and like to duplicate work (done)

packages/server/modules/workspaces/services/projects.ts
```diff
@@ -33,3 +34,44 @@ export const queryAllWorkspaceProjectsFactory = ({
...
+
+type GetWorkspaceProjectsArgs = {
+  workspaceId: string
+}
+
+type GetWorkspaceProjectsOptions = {
+  limit: number | null
+  cursor: string | null
+  filter: {
+    search?: string | null
```
fabis94, line 46: Workspace.projects typedefs still reference `UserProjectsFilter` btw, u also need a new type there

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    id: '',
+    name: 'John Nobody',
+    email: 'john-nobody@example.org',
+    role: Roles.Server.User
+  }
+
+  before(async () => {
+    await beforeEachContext()
+    await Promise.all(
+      [testUser, testNonWorkspaceMemberUser].map((user) => createTestUser(user))
```
fabis94, line 45: No need to change this, but there's also `createTestUsers` (plural) that would remove the need to do this Promise.all & mapping bit

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+
+      workspace = testWorkspace
+
+      const workspaceProjects = [
+        { name: 'Workspace Project A', workspaceId: workspace.id },
+        { name: 'Workspace Project B', workspaceId: workspace.id },
+        { name: 'Workspace Project C', workspaceId: workspace.id }
+      ]
+
+      await Promise.all(
```
fabis94, line 118: This one has a plural version too: `createTestWorkspaces`

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    before(async () => {
+      const testWorkspace: BasicTestWorkspace = {
+        id: '',
+        ownerId: testUser.id,
+        name: 'My Test Workspace'
+      }
+
+      await createTestWorkspace(testWorkspace, testUser)
+
+      workspace = testWorkspace
```
fabis94, line 110: Here like in the other PR, there's no need to have 2 variables for the same workspace - testWorkspace and workspace. Just move `testWorkspace` outside of `before()` and use that.

fabis94 (changes requested):

packages/server/modules/workspaces/services/projects.ts
cdriesler: oops, just sloppy, fixed

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
cdriesler: I need to learn to read, done

packages/server/modules/workspaces/tests/integration/projects.graph.spec.ts
cdriesler: much better, no sure why I did it this way

fabis94 (approved):

PR #2558, opened 2024-08-01
Feat: Add base for rendering workspaces items in the settings sidebar

Mikehrn:
Abstracted this from the adding general workspaces branches, so it doesn't blocks adding other pages.
Fetches all the workspaces and adds them to the settings sidebar. Doesn't add any links to specific pages but they will be added in the related PRs. Will only be shown when the user has actual workspace so it won't show for any existing users (as they wont have workspaces)

### DIFF
--- a/packages/frontend-2/components/settings/Dialog.vue
+++ b/packages/frontend-2/components/settings/Dialog.vue
@@ -60,7 +60,7 @@
                   'bg-highlight-2 hover:!bg-highlight-2':
                     targetMenuItem === key && targetWorkspaceId === workspaceItem.id
                 }"
-                @click="onWorkspaceMenuItemClick(workspaceItem.id, key)"
+                @click="onWorkspaceMenuItemClick(workspaceItem.id, `${key}`)"
               />
             </LayoutSidebarMenuGroup>
           </LayoutSidebarMenuGroup>
@@ -185,9 +185,9 @@ const selectedMenuItem = computed((): MenuItem | null => {
 })
 
 // Keep track of the selected workspace ID, to open the page for the correct workspace
-const onWorkspaceMenuItemClick = (id: string, target: string | number) => {
+const onWorkspaceMenuItemClick = (id: string, target: string) => {
   targetWorkspaceId.value = id
-  targetMenuItem.value = `${target}`
+  targetMenuItem.value = target
 }
 
 watch(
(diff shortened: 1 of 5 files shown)


andrewwallacespeckle (dismissed):
LGTM - Great to get the first PR out for workspaces. Because it's the first and we will base a lot on this structure, I'd wait for @fabis94 to review before we merge this one.

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -38,12 +37,33 @@
...
               @click="targetMenuItem = `${key}`"
             />
           </LayoutSidebarMenuGroup>
+          <LayoutSidebarMenuGroup v-if="hasWorkspaceItems" title="Workspace settings">
+            <template #title-icon>
+              <ServerStackIcon class="h-5 w-5" />
+            </template>
+            <LayoutSidebarMenuGroup
+              v-for="(workspaceItem, index) in workspaceItems"
+              :key="index"
```
fabis94, line 51: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#af99b815b2a6430da927ca10a1eb9007

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -87,7 +109,10 @@ type MenuItem = {
 
 const { activeUser: user } = useActiveUser()
 const breakpoints = useBreakpoints(TailwindBreakpoints)
+const { result: workspaceResult } = useQuery(settingsSidebarWorkspacesQuery)
```
fabis94, line 112: You need to introduce and start using a similar composable to `useIsAutomateModuleEnabled()` that check the workspaces feature flag.

Cause this query should not even be firing with that FF off - otherwise the server is gonna throw an error. So the workspace settings UI & also all the relevant GQL operations should be blocked by the feature flag.

fabis94 (changes requested):

packages/frontend-2/components/settings/Dialog.vue
Mikehrn: Good point; didn't know they would throw an error when not available. I've adjusted this to use a feature flag, following the example of automate

fabis94 (approved):

PR #2573, opened 2024-08-05
feat(viewer): handles missing instance components gracefully

didimitrie:
Description in here: [CNX-115: Viewer: handle gracefully instances with elements that failed to convert](https://linear.app/speckle/issue/CNX-115/viewer-handle-gracefully-instances-with-elements-that-failed-to)

It's a one liner fix, that if @AlexandruPopovici sees it as safe, we can merge.


AlexandruPopovici (commented):
Looks good to me. The only thing I would add is a
`Logger.warn("Some warning message")`
So that we notify we're skipping objects because of invalid data. That's usually what we've been doing so far in this type of scenarios

PR #2574, opened 2024-08-05
Feat: Add member settings page

Mikehrn:
This adds the base for members in workspace settings. As searching is not enabled yet I've commented out the search related parts, so we can add them back once we can. In additional we should add pagination at some point, but it's not possible yet.

There are few things that need to be followed-up on (I will create tickets for them and pick these up). Mainly the `WorkspaceRoles.vue` select component and the `ChangeRoleDialog.vue` component, to have the updated styling and be generic enough to be used in all settings related components. These are now mostly a copy of existing component with some tweaks.


packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -77,6 +77,7 @@
           !isMobile && 'simple-scrollbar overflow-y-auto flex-1'
         ]"
         :user="user"
+        :workspace-id="targetWorkspaceId"
```
fabis94, line 80: This dynamic component & property binding is not very nice, cause there's no type checking - you can forget to add a prop or misspell it, and there's not gonna be any TS errors here. While you don't have to change this now, ideally the `menuItemConfig` type should be adjusted so that you also specify the props for the component there and then bind them in the template using `v-bind="props"`. On top of that, there should be type checking that checks that the props you've provided for each component match the prop type signature for the component.

It would probably mean that MenuItem has to be some kind of generic type like:
```
type MenuItem<Component extends PropAnyComponent> = {
  title: string
  component: Component
  props?: InstanceType<Component>['$props']
}

```

packages/frontend-2/components/settings/shared/ChangeRoleDialog.vue
```diff
@@ -0,0 +1,58 @@
...
+}>()
+
+const props = defineProps<{
+  open: boolean
+  name: string
+  oldRole?: WorkspaceRoles
+  newRole?: WorkspaceRoles
+}>()
+
+const isOpen = computed({
```
fabis94, line 38: Replace this computed & the update:open emit w/ `declareModel('open')`

packages/frontend-2/components/settings/workspaces/Members.vue
```diff
@@ -0,0 +1,38 @@
+<template>
+  <section>
+    <div class="md:max-w-5xl md:mx-auto pb-6 md:pb-0">
+      <SettingsSectionHeader
+        hide-divider
+        title="Members"
+        text="Manage users in your workspace"
+      />
+      <LayoutTabsHoriztonal v-model:active-item="activeTab" :items="tabItems">
```
fabis94, line 9: typo in component name

packages/frontend-2/components/settings/shared/ChangeRoleDialog.vue
```diff
@@ -0,0 +1,58 @@
...
+          <ArrowRightIcon class="h-4 w-4" />
+          <span>{{ getRoleLabel(newRole).title }}</span>
+        </div>
+      </div>
+    </div>
+  </LayoutDialog>
+</template>
+
+<script setup lang="ts">
+import { LayoutDialog, type LayoutDialogButton } from '@speckle/ui-components'
```
fabis94, line 21: You don't need to import components from ui-components, they're auto-imported with the help of @speckle/ui-components-nuxt

packages/frontend-2/components/settings/workspaces/Members.vue
```diff
@@ -0,0 +1,38 @@
...
+          <div v-if="activeItem.id === 'invites'">Pending invites</div>
+        </template>
+      </LayoutTabsHoriztonal>
+    </div>
+  </section>
+</template>
+
+<script setup lang="ts">
+import type { LayoutTabItem } from '~~/lib/layout/helpers/components'
+import { LayoutTabsHoriztonal } from '@speckle/ui-components'
```
fabis94, line 25: You don't need to import components from ui-components, they're auto-imported with the help of @speckle/ui-components-nuxt

packages/frontend-2/components/settings/workspaces/members/Table.vue
```diff
@@ -0,0 +1,181 @@
...
+  () => ({
+    workspaceId: props.workspaceId
+  })
+)
+
+const showChangeUserRoleDialog = ref(false)
+const newRole = ref<WorkspaceRoles>()
+const userToModify = ref<UserItem>()
+
+const members = computed(() => {
```
fabis94, line 115: can be shortened
```ts
const members = computed(() => (result.value?.workspace.team || []).map(({ user, ...rest }) => ({...user, ...rest })))
```

fabis94: Also, there's a few unnecessary optional chaining operators, only add `?` operators where needed (TS will tell you)

`result?` - result will never be falsy, its a ref
`result.value.workspace?` - will never be falsy, the return type in the GQL schema is non-null, the resolver throws if a workspace isn't found (which should probably be handled somehow too)

packages/frontend-2/components/settings/workspaces/members/Table.vue
```diff
@@ -0,0 +1,181 @@
...
+      }
+    },
+    {
+      update: (cache, { data }) => {
+        const userId = userToModify.value?.id
+        const newRoleVal = newRole
+
+        if (data?.workspaceMutations?.updateRole) {
+          cache.modify({
+            id: getCacheId('WorkspaceCollaborator', userId as string),
```
fabis94, line 157: 1. Instead of overriding the type, store `userId` already in the scope where you know it exists - right after the check in line 140. The update() callback is invoked asynchronously and theoretically it is possible for userToModify to not have the same value then anymore.
2. This whole thing could be simpler without the need to do custom cache updates if you'd just ask back for the updated `Workspace.team` field in the mutation https://www.apollographql.com/docs/react/data/mutations/#include-modified-objects-in-mutation-responses

packages/frontend-2/lib/settings/graphql/queries.ts
```diff
@@ -12,3 +12,20 @@ export const settingsSidebarWorkspacesQuery = graphql(`
...
 `)
+
+export const settingsWorkspacesMembersQuery = graphql(`
+  query SettingsWorkspacesMembers($workspaceId: String!) {
+    workspace(id: $workspaceId) {
+      team {
+        role
+        id
+        user {
+          avatar
```
fabis94, line 23: [Always ask back for an `id`, wherever possible, to ensure that cache can be updated anywhere its needed: https://www.apollographql.com/docs/react/data/mutations/#include-modified-objects-in-mutation-responses](https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#2851514fbf9a406da31e430a57859565)

packages/frontend-2/lib/workspaces/graphql/mutations.ts
```diff
@@ -0,0 +1,12 @@
+import { graphql } from '~~/lib/common/generated/gql'
+
+export const workspaceUpdateRoleMutation = graphql(`
+  mutation WorkspaceUpdateRole($input: WorkspaceRoleUpdateInput!) {
```
fabis94, line 4: This seems to be a duplicate mutation of the settings one

fabis94 (changes requested):

packages/frontend-2/lib/workspaces/graphql/mutations.ts
Mikehrn: Forgot to delete the other one 😅

packages/frontend-2/components/settings/Dialog.vue
Mikehrn: That's a good point, it will probably be needed at some point as this might grow, different settings will require different props, I will make sure to adjust it

packages/frontend-2/lib/settings/graphql/queries.ts
```diff
@@ -12,3 +12,21 @@ export const settingsSidebarWorkspacesQuery = graphql(`
     }
   }
 `)
+
+export const settingsWorkspacesMembersQuery = graphql(`
+  query SettingsWorkspacesMembers($workspaceId: String!) {
+    workspace(id: $workspaceId) {
+      team {
```
fabis94, line 19: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#d40e40efc32f4e5ea67b68e848aadcd0

fabis94 (approved):
LGTM. I left a comment that you should keep in mind for future changes

packages/frontend-2/lib/settings/graphql/queries.ts
Mikehrn: Thanks :) I plan to make a PR to make a few adjustments, I've also noticed some other small things like fragment names not really matching anymore, Ill include fixing this

PR #2580, opened 2024-08-06
Feat: Add new dashboard

Mikehrn:
Adds the new dashboard page that will replace "projects" once we have the sidebar

Depends on Ghost API key env var being set (won't break otherwise, but it wont show any tutorials)

Screenshot 2024-08-06 at 08 18 50


packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
...
+      <div class="flex flex-col md:grid md:grid-cols-3 gap-3 pt-5">
+        <DashboardProjectCard
+          v-for="project in projects"
+          :key="project.id"
+          :project="project"
+        />
+      </div>
+    </section>
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">News &amp; tutorials</h2>
```
Mikehrn, line 26: Later this will be just tutorials, but now the feedback is a mix, we can update this text when we have tutorials only feed but not such a prio I think

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="flex flex-col gap-y-12">
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">Quick start</h2>
```
Mikehrn, line 4: Later this will be dismissible and the content might change, but there is no design for that yet

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
...
+        text: 'Ask a question',
+        props: { to: forumPageUrl, external: true }
+      }
+    ]
+  }
+])
+
+const projects = computed(() => projectsResult.value?.activeUser?.projects.items)
+
+async function fetchTutorials() {
```
Mikehrn, line 101: I was thinking to do this server-side, but it's non-block non-required content so I changed to being fetched async. Probably we can have some nicer loader for this later

packages/frontend-2/components/dashboard/ProjectCard.vue
```diff
@@ -0,0 +1,74 @@
...
+    id
+    name
+    role
+    updatedAt
+    models {
+      totalCount
+    }
+    team {
+      user {
+        avatar
```
fabis94, line 55: Use `LimitedUserAvatar` fragment, instead of hardcoding the specific fields UserAvatarGroup needs

packages/frontend-2/components/dashboard/ProjectCard.vue
```diff
@@ -0,0 +1,74 @@
...
+        id
+        name
+      }
+    }
+  }
+`)
+
+const props = defineProps<{
```
fabis94, line 63: What about subscriptions? You probably want to use `useGeneralProjectPageUpdateTracking` the same way its used in ProjectDashboardCard

packages/frontend-2/components/dashboard/TutorialCard.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <NuxtLink v-if="tutorial.url" :to="tutorial.url" external>
```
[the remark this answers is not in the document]
fabis94: + Its wasteful to even manage the component in memory if you don't actually want it. This `v-if` should be handled by the parent component - it should filter out any tutorial objects w/o an URL.

But this may be moot - is it really the case that TutorialCard.url can be optional? Feels like thats a prop that would always be there

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="flex flex-col gap-y-12">
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">Quick start</h2>
+      <div class="flex flex-col md:grid md:grid-cols-3 gap-3 pt-5">
+        <QuickStartCard
+          v-for="(quickStartItem, index) in quickStartItems"
+          :key="`quick-start-${index}`"
```
fabis94, line 8: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#af99b815b2a6430da927ca10a1eb9007

Adding a string suffix doesn't change the fact that the key is not a valid key because it doesn't uniquely identify the item. Each of those items has a unique title, why not use that?

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
...
+import { dashboardProjectsQuery } from '~~/lib/dashboard/graphql/queries'
+import type { TutorialCard, QuickStartItem } from '~~/lib/dashboard/helpers/types'
+import { useQuery } from '@vue/apollo-composable'
+import { useMixpanel } from '~~/lib/core/composables/mp'
+import GhostContentAPI from '@tryghost/content-api'
+import { docsPageUrl, forumPageUrl } from '~~/lib/common/helpers/route'
+import type { ManagerExtension } from '~~/lib/common/utils/downloadManager'
+import { downloadManager } from '~~/lib/common/utils/downloadManager'
+
+useHead({ title: 'Dashboard' })
```
fabis94, line 49: We should probably hide this behind a feature flag, if we don't want this to be accessible. Even the workspaces flag would work. You can return a 404 in an inline middleware.

fabis94 (changes requested):

packages/frontend-2/pages/dashboard.vue
Mikehrn: @fabis94 I can add this, but I dont really think we need to specifically hide this? It's not depended on anything workspace related, and you can only get there when know the URL (which is also fine, as it's a functional page)

fabis94: Ok, if its fine for unsuspecting visitors and search indexers to pick up on it, then its fine to leave it this way

packages/frontend-2/components/dashboard/TutorialCard.vue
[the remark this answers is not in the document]
Mikehrn: @fabis94 This comes from an external API, and technically `url` can be empty, it doesn't happen in practice but the type of the url allows for it

packages/frontend-2/lib/common/utils/downloadManager.ts
```diff
@@ -0,0 +1,23 @@
+export type ManagerExtension = 'exe' | 'dmg'
+
+// Util to download the Manager file
+export const downloadManager = (extension: ManagerExtension) => {
```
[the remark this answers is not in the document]
Mikehrn: @fabis94 Regarding 2, this snippet is used in both FE1 and FE2, I reused that. I added a promise because the browser can reject downloading files this way. We have a mixpanel even for succesful downloads and if the file itsn't actually downloaded it's misleading to throw that event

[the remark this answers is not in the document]
fabis94: @Mikehrn The try-catch makes sense then, but why the promise? There's no async logic here. And even if there were any, I'd think you could write it w/ async/await syntax instead of having to create a manual promise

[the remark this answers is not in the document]
Mikehrn: @fabis94 Yeah true, I'll remove that

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,129 @@
...
+
+const config = useRuntimeConfig()
```
fabis94, line 51: @Mikehrn There's also an auth middleware missing for this page, if a guest opens it its probably gonna throw

Mikehrn: @fabis94 Ah yeah good point :) Let me add it

packages/frontend-2/components/dashboard/ProjectCard.vue
```diff
@@ -0,0 +1,80 @@
...
+        <span class="pl-1 pr-2">•</span>
+        <span v-tippy="updatedAt.full">
+          {{ updatedAt.relative }}
+        </span>
+      </p>
+    </div>
+    <UserAvatarGroup :users="teamUsers" :max-count="4" />
+    <div>
+      <FormButton
+        :to="allProjectModelsRoute(project.id) + '/'"
```
fabis94, line 23: Why the trailing slash? The route helper should already format the route as needed. If it doesn't, then the helper needs changing, not the return of it

packages/frontend-2/lib/dashboard/graphql/queries.ts
```diff
@@ -0,0 +1,13 @@
+import { graphql } from '~~/lib/common/generated/gql'
+
+export const dashboardProjectsQuery = graphql(`
+  query DashboardProjectsQuery {
```
fabis94, line 4: Call it `DashboardProjectsPageQuery` - the convention so far has been to add the Page bit in the name for all top-level page queries

packages/frontend-2/lib/dashboard/graphql/queries.ts
```diff
@@ -0,0 +1,13 @@
...
+    activeUser {
```
fabis94, line 5: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#2851514fbf9a406da31e430a57859565

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,143 @@
...
+  } catch {
+    triggerNotification({
+      type: ToastNotificationType.Danger,
+      title: 'Download failed'
+    })
+  }
+}
+
+onMounted(() => {
+  fetchTutorials()
```
fabis94, line 141: 1. https://nuxt.com/docs/api/composables/use-async-data
2. use lazy key, if you want it  to be non blocking

fabis94 (changes requested):

packages/frontend-2/pages/dashboard.vue
fabis94: Its important to use nuxt's data fetching utils, cause they all kinds of stuff like caching, for example. Currently you will refetch the same data over and over again every time you open the dashboard, even if its already available.

Mikehrn: @fabis94 This will only work if we call an endpoint directly, right? It's using this Ghost API package, so I don't know if there is a way to make that work? Alternatively maybe I can switch not using their package and just call their API directly

```
const ghostContentApi = new GhostContentAPI({
  url: 'https://speckle.systems',
  key: config.public.ghostApiKey,
  version: 'v5.0'
})
```

fabis94: @Mikehrn No, useAsyncData allows you to do any kind of arbitrary async activity in it, you don't have to use their $fetch() util

Mikehrn: @fabis94 Nice, I didn't know that, I've updated, also added a small util to resize images returned from the Ghost API (as they seem to have no build-in option, so it always return full images which takes long time to load) :)

packages/frontend-2/components/dashboard/ProjectCard.vue
```diff
@@ -0,0 +1,80 @@
+<template>
+  <div
+    class="bg-foundation border border-1 border-outline-3 rounded-xl p-5 pb-4 flex flex-col gap-y-3"
```
andrewwallacespeckle, line 3: border-1 isn't a tailwind class

packages/frontend-2/components/dashboard/TutorialCard.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <NuxtLink :to="tutorial.url" external>
+    <div
+      class="bg-foundation border border-1 border-outline-3 rounded-xl flex flex-col overflow-hidden hover:border-outline-5 transition"
```
andrewwallacespeckle, line 4: border-1 isn't a tailwind class

packages/frontend-2/components/quick-start/Card.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div class="border border-1 border-outline-3 rounded-lg p-5 pt-4">
```
andrewwallacespeckle, line 2: border-1

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,142 @@
+<template>
+  <div class="flex flex-col gap-y-12">
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">Quick start</h2>
+      <div class="flex flex-col md:grid md:grid-cols-3 gap-3 pt-5">
```
andrewwallacespeckle, line 5: Why not `grid md:grid-cols-3 gap-3` rather than changing to and from flex?

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,142 @@
...
+          :key="quickStartItem.title"
+          :title="quickStartItem.title"
+          :description="quickStartItem.description"
+          :buttons="quickStartItem.buttons"
+        />
+      </div>
+    </section>
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">Recently updated projects</h2>
+      <div class="flex flex-col md:grid md:grid-cols-3 gap-3 pt-5">
```
andrewwallacespeckle, line 17: As above

packages/frontend-2/pages/dashboard.vue
```diff
@@ -0,0 +1,142 @@
...
+          v-for="project in projects"
+          :key="project.id"
+          :project="project"
+        />
+      </div>
+    </section>
+    <section>
+      <h2 class="text-heading-sm text-foreground-2">News &amp; tutorials</h2>
+      <div
+        class="flex flex-col gap-y-3 md:grid md:grid-cols-2 lg:grid-cols-4 gap-3 pt-5"
```
andrewwallacespeckle, line 28: As above, why flex and not just grid?

andrewwallacespeckle (commented):
Looks good, comments are very minor

packages/frontend-2/components/dashboard/TutorialCard.vue
```diff
@@ -0,0 +1,41 @@
+<template>
+  <NuxtLink class="w-full" to="tutorial.url" external>
```
fabis94, line 2: Please test that these links work, this is just a link w/ the URL "tutorial.url", it should be `:to=` not `to=`

fabis94 (changes requested):
Only found 1 remaining issue

packages/frontend-2/components/dashboard/TutorialCard.vue
fabis94: Also these should probably have target="_blank"

Mikehrn: @fabis94 Oops accidentally broke that in my last commit, fixed 😅

fabis94 (approved):

andrewwallacespeckle (approved):

PR #2581, opened 2024-08-06
fix(server): manual workspace role assignment on project invite create + more bugfixes

fabis94:
* Instead of mapping project roles to workspace roles, we default to a Guest workspace role or allow specifying it through a new mutation
* Fixes in OIDC auth strategy & serverRole assignment in stream invites


packages/server/modules/auth/strategies/oidc.js
```diff
@@ -97,21 +101,22 @@ module.exports = async (app, session, sessionStorage, finalizeAuth) => {
...
+          if (serverInfo.inviteOnly && !token) {
             throw new Error('This server is invite only. Please provide an invite id.')
           }
 
           // validate the invite
-          const validInvite = await validateServerInvite(
-            user.email,
-            req.session.inviteId
-          )
+          const validInvite = await validateServerInviteFactory({
```
fabis94, line 109: This is the OIDC bug: The service invocation referred to the old service fn

packages/server/modules/auth/tests/integration/registration.spec.ts
```diff
@@ -33,6 +37,22 @@ describe('Server registration', () => {
   let restApi: LocalAuthRestApiHelpers
   let apollo: TestApolloServer
 
+  const createInviteAsAdmin = async (
+    args: CreateProjectInviteMutationVariables | CreateServerInviteMutationVariables
+  ) => {
+    return await captureCreatedInvite(async () => {
```
alemagio, line 43: not a big deal, but as a general rule `return await` should be avoided

alemagio (approved):

packages/server/modules/auth/tests/integration/registration.spec.ts
fabis94: Here's a reason for why I did it this way: https://jsfiddle.net/p2bz4j5x/

If you don't await the return, the parent function does not appear in the error stack trace

fabis94: image

PR #2589, opened 2024-08-06
chore(server): typescriptified auth strategies

fabis94:
this was a long time coming...


packages/server/modules/auth/domain/logic.ts
```diff
@@ -1,11 +1,11 @@
-import type { UserInfo } from '@/modules/auth/domain/types'
+import type { UserinfoResponse } from 'openid-client'
```
gjedlicska, line 1: Lets not import an implementation type into our domain. Domain must not have implementation specific dependencies.

We're picking 3 string types.

alemagio:
@fabis94 
I tested OIDC locally with keycloak and it works

packages/server/modules/auth/domain/logic.ts
fabis94: @gjedlicska Is this about the "domain" folder (which i'm still not clear on - its purpose or the rules around it)? Cause I'll just move this helper function to helpers then, cause this helper that uses this type only exists to assist w/ openid-client, it has no other meaning in the module and is definitely not of domain level importance.

fabis94: I've done just that - moved this to helpers

gjedlicska (approved):

PR #2602, opened 2024-08-07
fix: objectloader totalChildrenCount

didimitrie:
one liner for the object loader to use a more reliable way of getting the total children count. the actual `totalChildrenCount` property is not really properly set anymore by the .NET sdk.

### DIFF
--- a/packages/objectloader/src/index.js
+++ b/packages/objectloader/src/index.js
@@ -150,7 +150,7 @@ class ObjectLoader {
 
     for await (const obj of this.getObjectIterator()) {
       if (first) {
-        this.totalChildrenCount = obj.totalChildrenCount
+        this.totalChildrenCount = Object.keys(obj?.__closure || {}).length
         first = false
         this.isLoading = true
       }


AlexandruPopovici (commented):
I might be wrong, since I didn't thoroughly look, but I don't think that part of the code gets reached currently with the way we load objects.

AlexandruPopovici:
*Coming back to this after holiday so we can properly handle it*
Is the PR triggered by https://speckle.community/t/percentage-progress-loader-for-speckle-viewer-api/12878?

We should definitely fix the `totalChildrenCount` thing so we can provide a progress value in `LoaderEvent.LoadProgress`. I guess my question would be, if `object-loader` should be where the number of total objects for a given stream is pre-computed. I would make sense I guess. Maybe even provided as a getter or function instead of a class member as it is now. That way, the client can query for it whenever it wants, and it's computation does not rely on timing. It would also remove a bit of state, and I'm always up for that :)

didimitrie:
It's triggered by SketchUp via dui3. This is good fix we already have in on the dui3 branch. Do you need it in another shape? We can discuss anytime!

didimitrie:
as discussed with @AlexandruPopovici, better solution incoming!

PR #2607, opened 2024-08-07
Fix(dui3): Use existing cards for latter operations

oguzhankoral:
Instead of creating new model cards, finds the existing one and patch it with important data before running send/receive operation.

Patching on send:
- `sendFilter`
- `isExpired`

Patching on receive:
- `selectedVersionId`
- `latestVersion`

SEND
SketchUp_CF9YQK47Yr-ezgif com-resize

RECEIVE
0iqTfHlSrB


didimitrie (approved):
i like it this way - well done!

PR #2608, opened 2024-08-08
local OIDC provider for testing

fabis94:
Should work automatically for new local instances, but existing ones (w/ existing DB) will need to run the SQL init file manually before starting up keycloak.

**LOCAL SETUP**
1. For existing docker compose deployments, invoke SQL found in `setup/db/11-docker_postgres_keycloack_init.sql` in your DB client manually (as the speckle user). New docker compose deployments can skip this step.
2. yarn dev:docker:up
3. Copy OIDC strategy env vars from .env-example into .env (in `server`)
4. When logging in in FE2, use speckle:speckle or register a new keycloak account

### DIFF
--- a/docker-compose-deps.yml
+++ b/docker-compose-deps.yml
@@ -46,10 +46,10 @@ services:
       KC_DB_USERNAME: keycloak
       KC_DB_PASSWORD: keycloak
 
-      # KC_HOSTNAME: 127.0.0.1
-      # KC_HOSTNAME_PORT: 9000
-      # KC_HOSTNAME_STRICT: false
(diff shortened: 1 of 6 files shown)


fabis94:
@iainsproat Alessandro's PR is incomplete, this does the full setup

fabis94:
@alemagio I didn't have to do any identity provider stuff, I just created a new realm w/ default clients, adjusted some client settings and created a couple of test users. Then I exported that realm and made sure its imported on keycloak init.

I was able to both register and log into Speckle with that, without having to mess around with any identity providers

PR #2610, opened 2024-08-08
chore(workspaces): repair ci breakages

cdriesler:
## Description & motivation

- Broke CI, should not be broken


cdriesler:
also puzzled, the failure was definitely linting though? but fabians's version appears to get things in order

PR #2613, opened 2024-08-08
Feat: Allow users to manage emails

Mikehrn:
Adds "Email addresses" page to the user settings

Screenshot 2024-08-08 at 12 49 17


andrewwallacespeckle (dismissed):
All good from me, but you should wait for a Fabs review on this one too

packages/frontend-2/lib/settings/graphql/queries.ts
```diff
@@ -30,3 +30,14 @@ export const settingsWorkspacesMembersQuery = graphql(`
     }
   }
 `)
+
+export const settingsUserEmailsQuery = graphql(`
+  query SettingsUserEmailsQuery {
+    activeUser {
```
fabis94, line 36: All of the ActiveUser fields should be its own `SettingsUserEmail_User` fragment that you can then reference in all of the mutations, instead of duplicating `id` and `emails` every time

fabis94 (dismissed):
Test out that apollo cache updates work as you expect (e.g. deleting an email, removes it from the list etc.)

packages/frontend-2/lib/settings/graphql/queries.ts
Mikehrn: Makes sense, I've updated it :) Everything works with the cache, didn't run into any issues when testing

fabis94 (approved):

PR #2615, opened 2024-08-08
fix(fe2): Use NuxtLink in model cards to allow cmd click

andrewwallacespeckle:
## Description & motivation
I noticed that users couldn't ctrl click on model cards, or right click and open in new tab. 

This is a big annoyance, so this PR adds NuxtLinks to handle this. 

1 NuxtLink is absolutely positioned to avoid hydration problems. Items do need to be relatively positioned on top, so other actions still work. 

These links should not appear in "Add model" type scenarios, so we use disableDefaultLinks.


packages/frontend-2/components/project/page/models/Card.vue
```diff
@@ -40,16 +50,17 @@
           class="px-4 w-full text-foreground-2 text-sm flex flex-col items-center space-y-1"
         />
         <template v-else-if="previewUrl">
-          <div
-            class="bg-foundation-page w-full h-48 rounded-xl border border-outline-2"
+          <NuxtLink
```
andrewwallacespeckle, line 53: Confirmed that when to is left undefined, NuxtLink just becomes a div

Mikehrn (approved):

packages/frontend-2/components/project/page/models/Results.vue
```diff
@@ -18,6 +18,7 @@
         :project-id="projectId"
         :source-apps="sourceApps"
         :contributors="contributors"
+        :disable-default-links="false"
```
fabis94, line 21: Its already false by default

fabis94 (approved):

PR #2616, opened 2024-08-08
feat(fe2): workspace project list

andrewwallacespeckle:
https://linear.app/speckle/issue/WEB-930/workspace-project-list

## Description & motivation
Adding the new workspace project list, where a single workspace will show all projects within that workspace. 

There are a few buttons here with no actions. This is either a task to be completed in another branch, or because a design is not ready.


packages/frontend-2/components/project/List.vue
andrewwallacespeckle, line 1: Most of this is existing code, I just moved it into it's own component.

packages/frontend-2/components/workspace/header/Actions.vue
andrewwallacespeckle, line 1: The design shows this button, but it doesn't show what the items should be. I've dropped in a few example items until this is clarified, and created a ticket to come back to this.

packages/frontend-2/middleware/requiresWorkspacesEnabled.ts
andrewwallacespeckle, line 1: This is the exact same implementation as automate FF, but I'm not sure it's working. 

When I remove the flag from my fe2 env, or even set it to false, it still loads this page. 

When I set the automate ff to false, it also doesn't disable automate

@fabis94 Have you any insight here? Maybe it's just not working locally for me?

packages/frontend-2/components/projects/AddDialog.vue
```diff
@@ -60,7 +64,8 @@ const mp = useMixpanel()
 const onSubmit = handleSubmit(async (values) => {
   await createProject({
     ...values,
-    visibility: visibility.value
+    visibility: visibility.value,
+    workspaceId: props.workspaceId
```
andrewwallacespeckle, line 68: This correctly adds the project to the Workspace. `useCreateProject` correctly updates the cache for `projectsDashboardQuery`, but I couldn't get it to also update `workspacePageQuery`.

@fabis94 @Mikehrn If either of you have any suggestions here let me know. If it's easier to jump on a call, or to just do it yourself, feel free to do that too.

packages/frontend-2/components/workspace/Dashboard.vue
```diff
@@ -0,0 +1,10 @@
+<template>
+  <WorkspaceProjectList :workspace-id="workspaceId" />
+</template>
+
+<script setup>
+import { useRoute } from 'vue-router'
+
+const route = useRoute()
+const workspaceId = route.params.id
```
fabis94, line 9: not reactive
https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#8a1d83812bf24d7ca4fabfa6d521cfd9

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -0,0 +1,246 @@
...
+          <FormTextInput
+            v-model="search"
+            name="modelsearch"
+            :show-label="false"
+            placeholder="Search..."
+            :custom-icon="MagnifyingGlassIcon"
+            color="foundation"
+            wrapper-classes="grow md:grow-0 md:w-60"
+            :show-clear="!!search"
+            @change="updateSearchImmediately"
```
fabis94, line 16: Seems to be largely duplicated from ProjectList. If you make it so that projects are fed in through props (remember to define fragments), you can re-use that component instead of duplicating this new one

fabis94: Also: If you still end up needing a new component, don't do this debouncing stuff directly and use `useDebouncedTextInput`

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -0,0 +1,246 @@
...
+
+onProjectsResult((res) => {
+  const projectsData = res.data?.workspace?.projects
+  if (projectsData) {
+    cursor.value = projectsData.cursor || null
+    infiniteLoaderId.value = JSON.stringify(projectsVariables.value?.filter || {})
+  }
+})
+
+const { onResult: onUserProjectsUpdate } = useSubscription(
```
fabis94, line 112: 1. This subscription and its document is duplicated here. Not only that but it has a duplicate operation name, which is not allowed for GQLGen.
2. This subscription relates to User.projects, not Workspace.projects, and thus it doesn't belong here

If this component remains, please clean it up and only leave in what's actually needed for it

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -0,0 +1,58 @@
...
+
+const props = defineProps<{
+  workspaceId: string
+}>()
+
+const {
+  result: workspaceInfoResult,
+  loading,
+  error
+} = useQuery(workspacePageQuery, () => ({
```
fabis94, line 48: Why is a header component invoking the query for the entire page? The query should be invoked at the top level where its first needed, and then fed down into child components through fragments. I can see that this same query is also independently invoked in WorkspaceProjectList - that's probably the best place for it (as nothing above it needs any of that data, but its own child components do).

packages/frontend-2/lib/workspaces/graphql/queries.ts
```diff
@@ -0,0 +1,31 @@
+import { graphql } from '~~/lib/common/generated/gql'
+
+export const workspacePageQuery = graphql(`
+  query WorkspacePageQuery(
+    $workspaceId: String!
+    $filter: WorkspaceProjectsFilter
+    $cursor: String
+  ) {
+    workspace(id: $workspaceId) {
+      name
```
fabis94, line 10: All of these fields are disconnected from the fragments and thus the Vue components that actually need them. No way to tell if any of this is necessary or not and gonna be a PITA to refactor when moving around/updating

https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#d40e40efc32f4e5ea67b68e848aadcd0

packages/frontend-2/middleware/requiresWorkspacesEnabled.ts
fabis94: @andrewwallacespeckle It should work, we can jump into a call if you want

packages/frontend-2/pages/workspaces/[id]/index.vue
```diff
@@ -0,0 +1,17 @@
...
+  <WorkspaceDashboard v-if="isLoggedIn" />
+  <div v-else class="mx-auto">
+    <CommonLoadingIcon />
+  </div>
+</template>
+<script setup lang="ts">
+const { isLoggedIn } = useActiveUser()
+
+definePageMeta({
+  middleware: ['auth', 'requires-workspaces-enabled']
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[id]/index.vue:11]: The isLoggedIn check is not necessary if we have the auth middleware. Also - if this ends up being a page with a single component in it, why not just move WorkspaceDashboard code to here? Otherwise there's an unnecessary layer where nothing happens, just extra complication and jumping around the codebase. There are some older pages that do this, but only because we were trying to render them in storybook at one point

fabis94 (changes requested):

packages/frontend-2/components/project/List.vue
```diff
@@ -0,0 +1,246 @@
...
+          <FormTextInput
+            v-model="search"
+            name="modelsearch"
+            :show-label="false"
+            placeholder="Search..."
+            :custom-icon="MagnifyingGlassIcon"
+            color="foundation"
+            wrapper-classes="grow md:grow-0 md:w-60"
+            :show-clear="!!search"
+            @change="updateSearchImmediately"
```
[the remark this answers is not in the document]
andrewwallacespeckle: > 2\. showClear being conditional is redundant, it already hides the clear button if there's no value

This isn't my code, I just moved it here, but removing show-clear prop breaks this. Now the button never appears, when the field is empty of with content.

packages/frontend-2/components/project/List.vue
```diff
@@ -0,0 +1,246 @@
...
+const debouncedSearch = ref('')
+const openNewProject = ref(false)
+const showLoadingBar = ref(false)
+
+const { activeUser, isGuest } = useActiveUser()
+const { triggerNotification } = useGlobalToast()
+const areQueriesLoading = useQueryLoading()
+const apollo = useApolloClient().client
+
+const {
```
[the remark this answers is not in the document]
andrewwallacespeckle: List has been merged into Dashboard

packages/frontend-2/components/project/List.vue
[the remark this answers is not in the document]
fabis94: You still need the prop, it just doesn't need to be conditional. Anyway, you can leave it

packages/frontend-2/components/projects/AddDialog.vue
andrewwallacespeckle: I've updated the createProject mutation to handle cache updates for Workspace.projects as you suggested. I resolved some TypeScript and ESLint errors in the process but no sure everything was necessary. I'd appreciate any feedback on whether there's a better approach.

packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -18,41 +18,36 @@
...
+        <div class="flex flex-col sm:flex-row gap-2">
+          <FormTextInput
+            name="modelsearch"
+            :show-label="false"
+            placeholder="Search..."
+            :custom-icon="MagnifyingGlassIcon"
+            color="foundation"
+            wrapper-classes="grow md:grow-0 md:w-60"
+            :show-clear="!!search"
+            :model-value="bind.modelValue"
```
fabis94, line 38: `v-bind="bind"` w/ the latest changes to useDebouncedTextInput (assuming you have main merged in) - change this everywhere

packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -130,18 +115,23 @@ const promoBanners = ref<PromoBanner[]>([
...
-const { triggerNotification } = useGlobalToast()
-const areQueriesLoading = useQueryLoading()
-const apollo = useApolloClient().client
+const {
+  on,
+  bind,
+  value: search
+} = useDebouncedTextInput({
+  debouncedBy: 800,
+  model: ref('')
```
fabis94, line 124: The model key is redundant here, the only point for it is to feed in some pre-existing model ref you want to use. If you're just creating a new one anyway, omit the model key altogether. Fix this everywhere

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -0,0 +1,154 @@
...
+const {
+  on,
+  bind,
+  value: search
+} = useDebouncedTextInput({
+  debouncedBy: 800,
+  model: ref('')
+})
+
+const {
```
fabis94, line 90: Simplify all of this pagination stuff w/ `usePaginatedQuery`. You can leave ProjectsDashboard.vue as is since its an old page, but this is a new one.

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -0,0 +1,154 @@
...
+            fixed-height
+          /> -->
+        </div>
+        <FormButton v-if="!isGuest" @click="openNewProject = true">
+          New project
+        </FormButton>
+      </div>
+    </div>
+
+    <CommonLoadingBar :loading="showLoadingBar" class="my-2" />
```
fabis94, line 36: There's still a large amount of duplication between this and ProjectsDashboard, but in an effort to get this out quicker lets leave it as is for now

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -0,0 +1,59 @@
...
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { graphql } from '~~/lib/common/generated/gql'
+import type { WorkspaceInfo_WorkspaceFragment } from '~~/lib/common/generated/gql/graphql'
+
+graphql(`
+  fragment WorkspaceInfo_Workspace on Workspace {
```
fabis94, line 38: The fragment name is supposed to be based on the component name: WorkspaceHeader_Workspace

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -0,0 +1,59 @@
...
+    name
+    logo
+    description
```
fabis94, line 41: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#2851514fbf9a406da31e430a57859565

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -133,6 +137,37 @@ export function useCreateProject() {
               },
               { fieldNameWhitelist: ['admin'] }
             )
+
+            // New cache update for Workspace.projects
+            if (input.workspaceId) {
+              const workspaceCacheId = getCacheId('Workspace', input.workspaceId)
+
+              modifyObjectFields<
+                { filter?: { search?: string } },
```
fabis94, line 146: You should use GQLGen types, instead of typing out these types manually, otherwise if they change this code is not gonna show errors.

`modifyObjectFields<WorkspaceProjectsArgs, Workspace['projects']>`

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -133,6 +137,37 @@ export function useCreateProject() {
...
+                { items: Reference[]; totalCount: number }
+              >(cache, workspaceCacheId, (fieldName, variables, value, details) => {
```
fabis94, line 148: There's a convention of prefixing unused params - fieldName - with an underscore. `(_fieldName, variables, values, details) => {}`

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -133,6 +137,37 @@ export function useCreateProject() {
...
+                if (fieldName !== 'projects') return
+                if (variables?.filter?.search?.length) return
```
fabis94, line 150: I just realized you can do the eviction from here so that you don't have to do the other evictObjectFields call.
```
if (variables?.filter?.search?.length) {
  return details.DELETE
}
```

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -133,6 +137,37 @@ export function useCreateProject() {
...
+
+                const newVal = { ...value }
+                if (Array.isArray(value.items)) {
+                  newVal.items = [details.ref('Project', newProject.id), ...value.items]
+                } else {
+                  newVal.items = [details.ref('Project', newProject.id)]
+                }
+                newVal.totalCount = (value.totalCount || 0) + 1
```
fabis94, line 158: There's a reason all of the fields of `value` are optional - you don't know which if any of those fields are actually going to be requested and in the cache. If `totalCount` is undefined, meaning - not requested, you can't just default to 0. Just because we didn't ask for a totalCount from the API (the list may actually have a 100 workspaces), doesn't mean that we should now show it has 1 on the frontend.

The correct way to handle this is to only update totalCount if it exists in the cache. If it doesn't - leave it as is.

```
const newTotalCount = isUndefined(value?.totalCount) ? undefined : Math.max(0, (value?.totalCount || 0) - 1)
return {
  ...value,
  ...(isUndefined(newTotalCount) ? {} : {totalCount: newTotalCount})
}
```

Same goes for all other fields like `items`

packages/frontend-2/lib/workspaces/graphql/queries.ts
```diff
@@ -0,0 +1,22 @@
+import { graphql } from '~~/lib/common/generated/gql'
+
+export const workspacePageQuery = graphql(`
+  query WorkspacePageQuery(
+    $workspaceId: String!
+    $filter: WorkspaceProjectsFilter
+    $cursor: String
+  ) {
+    workspace(id: $workspaceId) {
+      id
```
fabis94, line 10: Where's id and role coming from? And totalCount & cursor under projects? Is there a specific reason they're not covered by fragments?

fabis94 (changes requested):

packages/frontend-2/middleware/requireValidWorkspace.ts
```diff
@@ -0,0 +1,56 @@
...
+      context: {
+        skipLoggingErrors: true
+      }
+    })
+    .catch(convertThrowIntoFetchResult)
+
+  if (data?.workspace?.id) return
+
+  const isForbidden = (errors || []).find((e) => e.extensions['code'] === 'FORBIDDEN')
+  const isNotFound = (errors || []).find(
```
fabis94, line 29: Have you tested this out? Some of these don't make sense, like there is no error w/ the code `WORKSPACE_NOT_FOUND`.

You should actually try to reach a non-existant workspace or a workspace the user doesn't have access to, and then you'll see the proper error codes you're supposed to use

I checked, the not found error code is: WORKSPACE_NOT_FOUND_ERROR

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -0,0 +1,111 @@
...
+const {
+  on,
+  bind,
+  value: search
+} = useDebouncedTextInput({
+  debouncedBy: 800
+})
+
+const { query, onInfiniteLoad } = usePaginatedQuery({
+  query: workspacePageQuery,
```
fabis94, line 81: You can leave it as is for now, but its not great that we're asking for all of the full page requirements (like the header data) for every page, when we're only interested in the next page of project items.

That's why we tend to use a separate pagination query, as opposed to using the base query for pagination also.

usePaginatedQuery also has a `resolveInitialResult` param, that enables it to read the base page query's data as the first page result, and then use the new limited paginated query for subsequent pages

fabis94 (changes requested):
nearly there

packages/frontend-2/components/workspace/header/Header.vue
andrewwallacespeckle: It is resolved, I'm not sure what's happening. GitHub only showed the change when i updated this:
image

packages/frontend-2/components/workspace/ProjectList.vue
andrewwallacespeckle: This should now be done. let me know if i misunderstood anything.

fabis94 (approved):

fabis94 (approved):

PR #2622, opened 2024-08-08
fix(workspaces): guests do not get workspace-wide project access

cdriesler:
## Description & motivation

- I incorrectly wrote some workspace role => project role logic that granted workspace-wide project roles to workspace guests
- We should no do this

## Changes:

- Do not do this anymore

## Checklist:

### DIFF
--- a/packages/server/modules/workspaces/events/eventListener.ts
+++ b/packages/server/modules/workspaces/events/eventListener.ts
@@ -41,13 +41,17 @@ export const onProjectCreatedFactory =
     const workspaceMembers = await getWorkspaceRoles({ workspaceId })
 
     await Promise.all(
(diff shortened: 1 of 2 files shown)


fabis94 (dismissed):

cdriesler:
Small but important behavior change: updated the project-role-changing logic in `updateWorkspaceRole` to only fire when a member is first added to a workspace. This matches our current understanding that project roles are meant to be managed entirely separately _for now_.

There's still some work to iron out what a "synchronized" workspace-project relationship looks like, but with this change we should be safe for the Sept 1st release.

Also added a test for this to prevent regression

cc @fabis94 @iainsproat

fabis94 (approved):

PR #2635, opened 2024-08-12
Feat: Add general workspace settings page

Mikehrn:
Adds General workspace settings page, moved menuItemConfig to a composable as this will be something we will probably refactor this a bit more when we change this from a modal to the sidebar. There is some logic required after deleting a workspace (mainly where does the user gets redirected to), will create a ticket to check with Michal on this behavior and implement that when he is back. Same for the avatar, some design changes are needed for this component, there is a ticket to add this


cdriesler (approved):

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -141,80 +130,26 @@ const { result: workspaceResult } = useQuery(settingsSidebarWorkspacesQuery, nul
...
+    serverMenuItems.value,
+    workspaceMenuItems.value
   ]
   for (const category of categories) {
     if (targetMenuItem.value && targetMenuItem.value in category) {
       return category[targetMenuItem.value]
     }
   }
 
-  if (!isMobile.value && targetMenuItem.value) {
```
Mikehrn, line 207: Removed this because it was more annoying to add up on a random page, and not having a `targetMenuItem` is possible behavior

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -19,12 +19,12 @@
...
             <LayoutSidebarMenuGroupItem
-              v-for="(sidebarMenuItem, key) in menuItemConfig.user"
+              v-for="(sidebarMenuItem, key) in userMenuItems"
               :key="key"
               :label="sidebarMenuItem.title"
-              :class="{
-                'bg-highlight-2 hover:!bg-highlight-2': targetMenuItem === key
-              }"
+              :class="[
+                targetMenuItem === key && 'bg-highlight-2 hover:!bg-highlight-2'
```
fabis94, line 26: I don't think this is simpler than the check was this before. This relies on you understanding that the last truthy value of an AND check in JS is the one that gets returned - seems overly clever and I'd avoid it. Previous syntax was more self explanatory: `"expected classes": whetherTheyShouldBeOn"`

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -56,33 +56,20 @@
...
-                v-for="(workspaceMenuItem, itemKey) in menuItemConfig.workspace"
+                v-for="(workspaceMenuItem, itemKey) in workspaceMenuItems"
                 :key="`${key}-${itemKey}`"
                 :label="workspaceMenuItem.title"
-                :class="{
-                  'bg-highlight-2 hover:!bg-highlight-2':
-                    targetMenuItem === itemKey && targetWorkspaceId === workspaceItem.id
-                }"
+                :class="[
+                  targetMenuItem === itemKey &&
```
fabis94, line 63: So complicated! Extract to a computed and preferably avoid the return by `&& value`

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -102,37 +89,39 @@
...
-import { settingsSidebarWorkspacesQuery } from '~/lib/settings/graphql/queries'
-import { useIsWorkspacesEnabled } from '~/composables/globals'
+import { graphql } from '~~/lib/common/generated/gql'
 
-type MenuItem = {
-  title: string
-  component: ReturnType<typeof defineComponent>
-}
+graphql(`
+  fragment SettingsSidebarWorkspaces_WorkspaceCollection on WorkspaceCollection {
```
fabis94, line 110: 1. The name of the component is: SettingsDialog, so the fragment name is off
2. I think it would be better if the fragment were defined on `User` instead (Ask for `workspaces`), cause otherwise it feels like you're omitting a part of the requirement. Considering that this is the only fragment the query asks for, its a bit confusing that the fragment only paints part of the picture (asking for some kind of workspace collection) while the query then fills out the rest (specifying that its supposed to be the active user's workspace collection)

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
```diff
@@ -0,0 +1,94 @@
...
+import {
+  convertThrowIntoFetchResult,
+  getFirstErrorMessage
+} from '~~/lib/common/helpers/graphql'
+import { homeRoute } from '~/lib/common/helpers/route'
+import type { SettingsWorkspaceGeneralDelete_WorkspaceFragment } from '~~/lib/common/generated/gql/graphql'
+import { ToastNotificationType, useGlobalToast } from '~~/lib/common/composables/toast'
+
+graphql(`
+  fragment SettingsWorkspaceGeneralDelete_Workspace on Workspace {
```
fabis94, line 39: You aren't asking for this fragment in any other fragments/queries. It works now, cause the other fragments already ask for the same data, but you should always make the dependency explicit by asking for this in the parent component's fragment (SettingsWorkspaceGeneral)

packages/frontend-2/components/settings/workspaces/General.vue
```diff
@@ -0,0 +1,144 @@
...
+    triggerNotification({
+      type: ToastNotificationType.Danger,
+      title: 'Workspace update failed',
+      description: errorMessage
+    })
+  }
+})
+
+function toggleDeleteDialog() {
+  showDeleteDialog.value = true
```
fabis94, line 131: Always setting the same value is not what `toggle` means

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
```diff
@@ -0,0 +1,94 @@
...
+      the selected workspace?
+    </p>
+    <div
+      class="rounded border bg-foundation-2 border-outline-3 text-body-2xs text-foreground font-medium py-3 px-4 my-4"
+    >
+      {{ workspace.name }}
+    </div>
+    <p>
+      This action
+      <span class="font-medium">cannot</span>
```
fabis94, line 20: I'd still add the "type in your workspace name" check, even if there's no official designs and we have to go by the same design we have in project settings. This just feels way too easy to delete

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
```diff
@@ -0,0 +1,94 @@
...
+  }).catch(convertThrowIntoFetchResult)
+
+  if (result?.data) {
+    isOpen.value = false
+    triggerNotification({
+      type: ToastNotificationType.Success,
+      title: 'Workspace deleted',
+      description: `The ${props.workspace.name} workspace has been deleted`
+    })
+    navigateTo(homeRoute)
```
fabis94, line 66: The workspace is still in the local apollo cache tho, so it may appear elsewhere in the site (e.g. maybe settings workspace list).

You need to do a cache mutation where you evict the Workspace object from cache, and possibly even alter the totalCount of known workspace lists (e.g. User.workspaces)

```
// Evict object from cache
cache.evict({
  id: getCacheId('Workspace', workspaceId)
})

modifyObjectFields(cache, getCacheId('User', activeUserId), () => {..}, {fieldwhitelist: ['workspaces]})
```

fabis94 (approved):

fabis94 (changes requested):
misclick, wanted to request changes

packages/frontend-2/components/settings/Dialog.vue
Mikehrn: Personally I find this syntax more readable, but no strong preference, happy to change it back :)

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
Mikehrn: I agree but Michal was thinking about various options for the check, it was previously discussed to leave it out for now and implement it properly once he is back

packages/frontend-2/components/settings/Dialog.vue
fabis94: @Mikehrn The underlying problem is that this is too complicated to just be put in the :class attribute as is. If it can't be a computed, it can be a function.

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
Mikehrn: It only needs the ID and name, but I think naming the prop `workspace` was a bit misleading, change it be explict `workspaceId` and `workspaceName`

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
Mikehrn: Good point, thanks for for snippet. I completely forgot about this, I've updated it to remove the item from the cache and update the total count

packages/frontend-2/components/settings/Dialog.vue
Mikehrn: Get your point now, updated it :)

packages/frontend-2/components/settings/workspaces/General.vue
```diff
@@ -0,0 +1,143 @@
...
+
+function openDeleteDialog() {
+  showDeleteDialog.value = true
+}
+
+watch(
+  () => workspaceResult,
+  () => {
+    if (workspaceResult.value?.workspace) {
+      name.value = workspaceResult.value.workspace.name
```
fabis94, line 137: You're also asking for the workspace ID and workspace name, and yet these don't show up in the fragment. If the DeleteDialog removes those requirements, this seemingly unrelated component will break

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
```diff
@@ -0,0 +1,120 @@
...
+    workspaceId: props.workspace.id
+  }).catch(convertThrowIntoFetchResult)
+
+  if (result?.data) {
+    if (activeUser.value) {
+      cache.evict({
+        id: getCacheId('Workspace', props.workspace.id)
+      })
+
+      modifyObjectFields<{ workspaces: WorkspaceCollection }, WorkspaceCollection>(
```
fabis94, line 75: 1. The generic arguments are off, the 1st arg is the field's variables type and {workspaces} as variables doesn't make any sense.
2. modifyObjectFields<UserWorkspacesArgs, User['Workspaces']> - you can resolve both the field's variables and return type from gqlgen, without having to manually figure out what they are and keep them in sync

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
```diff
@@ -0,0 +1,120 @@
...
+        cache,
+        activeUser.value.id,
+        (fieldName, _variables, value) => {
+          return {
+            ...value,
```
fabis94, line 80: There's a reason all of the fields of `value` are optional - you don't know which if any of those fields are actually going to be requested and in the cache. If `totalCount` is undefined, meaning - not requested, you can't just default to 0. Just because we didn't ask for a totalCount from the API (the list may actually have a 100 workspaces), doesn't mean that we should now show it has 1 on the frontend.

The correct way to handle this is to only update totalCount if it exists in the cache. If it doesn't - leave it as is.

```
const newTotalCount = isUndefined(value?.totalCount) ? undefined : Math.max(0, (value?.totalCount || 0) - 1)
return {
  ...value,
  ...(isUndefined(newTotalCount) ? {} : {totalCount: newTotalCount})
}
```

packages/frontend-2/lib/settings/graphql/queries.ts
```diff
@@ -3,12 +3,15 @@ import { graphql } from '~~/lib/common/generated/gql'
 export const settingsSidebarWorkspacesQuery = graphql(`
   query SettingsSidebarWorkspaces {
```
fabis94, line 4: No big deal, but I'd remove the `Workspaces` part from the query name - sure, currently its about workspaces, but ultimately its about pulling in all of the data that the SettingsSidebar needs, which may include other stuff in the future.

fabis94 (changes requested):
Double check how cache mutations behave in practice, whether total counts are updated correctly etc.

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
fabis94: 3. Your modifyObjectFields callback should skip updating totalCount if variables.filter.search is defined, cause we can't know for sure if the deleted workspace appeared in that query. For those kinds of queries, just evict them instead: evictObjectFields

packages/frontend-2/components/settings/workspaces/General/DeleteDialog.vue
Mikehrn: Thanks for the explanation, not super familiar with best practices regarding this so it's helpful :) 
So I've updated it like this:

```      modifyObjectFields<UserWorkspacesArgs, User['workspaces']>(
        cache,
        activeUser.value.id,
        (_fieldName, variables, value, { DELETE }) => {
          if (variables?.filter?.search?.length) return DELETE

          const newTotalCount = isUndefined(value?.totalCount)
            ? undefined
            : Math.max(0, (value?.totalCount || 0) - 1)

          return {
            ...value,
            ...(isUndefined(newTotalCount) ? {} : { totalCount: newTotalCount })
          }
        },
        { fieldNameWhitelist: ['workspaces'] }
      )```
      
      So basically we first check if it's part of a search filter (which it can't be now, but I guess makes it more future proof), and then update the totalCount if it's set in the cache, is this the correct approach?

fabis94 (approved):

PR #2637, opened 2024-08-12
fix(objectloader): prevent cache read/write for broken data

fabis94:
objectloader was caching and retrieving from cache broken Objects API responses


AlexandruPopovici:
@fabis94 what's the `startsWith('<html')` check for?

fabis94:
@AlexandruPopovici It caches failed Objects API responses (e.g. ones with 503 status code) - ones that contain HTML in them, not the json output. The startsWith check checks that its a proper JSON object without doing the heavy JSON.parse call

fabis94:
@AlexandruPopovici 
image

AlexandruPopovici (approved):

PR #2644, opened 2024-08-12
fix(workspaces): add pagination to workspace team members

cdriesler:
## Description & motivation

- We want to paginate queries to `workspace.team`

## Changes:

- Introduces pagination with the user name as the cursor and order
- Return `WorkspaceCollaboratorCollection` instead of a flat array from query
- Tests


packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -145,7 +145,11 @@ type Workspace {
   Active user's role for this workspace. `null` if request is not authenticated, or the workspace is not explicitly shared with you.
   """
   role: String
-  team(filter: WorkspaceTeamFilter): [WorkspaceCollaborator!]!
+  team(
```
fabis94, line 148: This will likely break Mike's members frontend, coordinate with him on how to release this. Maybe his adjustments to the FE should be done in this same branch to avoid deploying a BE change that breaks the FE.

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -168,7 +168,7 @@ export const upsertWorkspaceRoleFactory =
 
 export const getWorkspaceCollaboratorsFactory =
   ({ db }: { db: Knex }): GetWorkspaceCollaborators =>
-  async ({ workspaceId, filter = {} }) => {
+  async ({ workspaceId, filter = {}, cursor, limit }) => {
```
fabis94, line 171: 1) There should be a default limit, otherwise the query is just gonna return everything in one long list
2) The limit, whatever it ends up being, should be clamped (lodash's `clamp()`) between an allowed minimum and maximum, lets say 0 and 100

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -192,13 +193,27 @@ export const getWorkspaceCollaboratorsFactory =
       query.andWhere(DbWorkspaceAcl.col.role, role)
     }
 
+    if (cursor) {
+      query.andWhere(Users.col.name, '>', cursor)
```
fabis94, line 197: I don't think Users.name is a good cursor, there could be multiple people with the same name and that will likely break pagination. Also, I don't think you should order by Users.name, it makes more sense to show new collaborators first. Can we add a createdAt/updatedAt to workspaces_acl and order by/cursor by that?

If not, at least order by/cursor by User.createdAt (makes less sense tho).

fabis94: Also cursors should be opaque: https://graphql.org/learn/pagination/

> Especially if the cursors are opaque, either offset or ID-based pagination can be implemented using cursor-based pagination (by making the cursor the offset or the ID), and using cursors gives additional flexibility if the pagination model changes in the future. As a reminder that the cursors are opaque and that their format should not be relied upon, we suggest base64 encoding them.

Use encode/decode cursor utilities from `import { decodeCursor, encodeCursor } from '@/modules/shared/helpers/graphqlHelper'`

packages/server/modules/workspaces/tests/integration/workspaces.graph.spec.ts
```diff
@@ -116,37 +117,174 @@ describe('Workspaces GQL CRUD', () => {
...
+          workspaceId: largeWorkspace.id,
+          limit: 25
         })
 
         expect(res).to.not.haveGraphQLErrors()
-        expect(res.data?.workspace.team.length).to.equal(2)
+        expect(res.data?.workspace.team.items.length).to.equal(6)
+      })
+
+      it('should return workspace members in the correct order', async () => {
```
fabis94, line 202: Add a test case for when you have multiple users with the same name

fabis94 (changes requested):

cdriesler:
Back to you @fabis94 , Mike is going to do the FE changes here before we merge. Welcome to wait for those or review backend changes now

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -161,24 +166,36 @@ export const upsertWorkspaceRoleFactory =
...
   }
 
 export const getWorkspaceCollaboratorsFactory =
   ({ db }: { db: Knex }): GetWorkspaceCollaborators =>
-  async ({ workspaceId, filter = {} }) => {
+  async ({ workspaceId, filter = {}, cursor, limit = 25 }) => {
     const query = DbWorkspaceAcl.knex(db)
       .select<Array<UserWithRole & { workspaceRole: WorkspaceRoles }>>([
         ...Users.cols,
         knex.raw(`${DbWorkspaceAcl.col.role} as "workspaceRole"`),
```
fabis94, line 180: 1. Why do 'withoutTablePrefix' if you do actually add the table prefix, just manually?
2. You can use DbWorkspaceAcl.colAs() to alias columns
```
DbWorkspaceAcl.colAs('role', 'workspaceRole'),
DbWorkspaceAcl.colAs('createdAt', 'workspaceCreatedAt'),
DbWorkspaceAcl.colAs('updatedAt', 'workspaceUpdatedAt')
```

But that's all moot, cause you don't need to avoid aliases and you don't need to alias columns either - because the table aliases are already there. So if you do orderby() or groupby() DbWorkspaceAcl.col.createdAt, its already distinct from any other kind of createdAt because of the col prefix

packages/server/modules/workspaces/tests/integration/repositories.spec.ts
```diff
@@ -140,7 +140,13 @@ describe('Workspace repositories', () => {
       const { id: userId } = await createAndStoreTestUser()
       const { id: workspaceId } = await createAndStoreTestWorkspace()
 
-      await upsertWorkspaceRole({ userId, workspaceId, role: 'workspace:member' })
+      await upsertWorkspaceRole({
+        userId,
+        workspaceId,
+        role: 'workspace:member',
+        createdAt: new Date(),
```
fabis94, line 147: seems like there's a lot of duplication here, each of those upsertWorkspaceRole calls has only 1-2 different lines. that said - up to you to evaluate if they are fundamentally duplicate or only temporarily duplicate and probably gonna change a lot from each other in the future

packages/server/modules/workspaces/tests/integration/workspaces.graph.spec.ts
```diff
@@ -116,37 +117,176 @@ describe('Workspaces GQL CRUD', () => {
...
+            },
+            'workspace:guest'
+          ]
+        ]
+
+        for (const [user, role] of workspaceMembers) {
+          await createTestUser(user)
+          await assignToWorkspace(largeWorkspace, user, role)
+          // Overly-careful guarantee that `createdAt` values are different
+          await wait(50)
```
fabis94, line 190: 50 seems too much, even wait(1) should be fine, I would think

packages/server/modules/workspaces/tests/unit/events/eventListener.spec.ts
```diff
@@ -15,17 +15,23 @@ describe('Event handlers', () => {
         {
           workspaceId,
           userId: cryptoRandomString({ length: 10 }),
-          role: Roles.Workspace.Admin
+          role: Roles.Workspace.Admin,
```
fabis94, line 18: So many different places needing this duplicate block of createdAt/updatedAt being added makes me feel like the type should've just been changed instead to make createdAt/updatedAt optional and then possibly renamed to WorkspaceAclInsert to represent the ACL model pre-insertion, and then possibly having another type with those fields made required whenever you're returning them from the DB.

I dunno, might be too big of a change so maybe not worth doing, but having to duplicate these dates everywhere does feel icky

fabis94 (changes requested):
nearly there, left some minor comments

packages/server/modules/workspaces/tests/integration/workspaces.graph.spec.ts
cdriesler: +1 done

packages/server/modules/workspaces/repositories/workspaces.ts
cdriesler: kind of silly, in retrospect. ty for breaking it down

fabis94: I guess it may not be obvious, but DBHelper.col.X does add table prefixes by default, so you don't need to manually ever do that. Its only really for update calls where you want to use .withoutTablePrefix., cause `SET X = Y` instructions can't have table prefixes in them

cdriesler:
Simplified query, deferring the WorkspaceAcl type change as discussed (but keeping in mind for future)

cdriesler:
On pause pending frontend changes to use new gql query shape

Mikehrn:
> On pause pending frontend changes to use new gql query shape

@cdriesler I started this Wednesday but wasn't able to finish it, will wrap it and up and push it today

packages/server/modules/workspacesCore/domain/types.ts
```diff
@@ -12,4 +12,10 @@ export type Workspace = {
 
 export type WorkspaceWithOptionalRole = Workspace & { role?: WorkspaceRoles }
 
-export type WorkspaceAcl = { userId: string; role: WorkspaceRoles; workspaceId: string }
+export type WorkspaceAcl = {
+  userId: string
+  role: WorkspaceRoles
+  workspaceId: string
+  createdAt: Date
+  updatedAt: Date
```
gjedlicska, line 20: as discussed, we only need createdAt, updatedAt has not usefull meaning for now.

gjedlicska (changes requested):

packages/server/modules/workspacesCore/domain/types.ts
cdriesler: created the `workspace_acl_updates` table as an exercise - now we can keep `workspace_acl` unchanged

cdriesler:
🚨 🚨 COOL SQL ALERT 🚨 🚨 

Created `workspace_acl_updates` table to log all changes to `workspace_acl` and give us some history.

Please give special attention to what I did in `getWorkspaceCollaboratorsFactory` to:

- join `workspace_acl_updates` (many) with `workspace_acl` (one)
- order things by the earliest `updatedAt` in `workspace_acl_updates` for user

cdriesler:
DO NOT MERGE THIS

cdriesler:
JK FALSE ALARM

gjedlicska (approved):

PR #2647, opened 2024-08-13
`getTotalObjectCount` implementation for object-loader

AlexandruPopovici:
## Description & motivation
`object-loader` now has a `getTotalObjectCount` function that can be used autonomously, separate of any loading process. There is an un-ideal aspect of it, where we parse the root object string data in order to get the closure length, but the as an alternative we can either do string gymnastics which might be as inefficient as parsing in the end. Or, make sure we only parse the root object once, regardless where that happens.

Additionally, the `SpeckleLoader` now uses `getTotalObjectCount` to provide correct progress values in `LoaderEvent.LoadProgress` event addressing the reported issue 

## Checklist:

### DIFF
--- a/packages/objectloader/src/index.js
+++ b/packages/objectloader/src/index.js
@@ -128,6 +128,18 @@ class ObjectLoader {
     Object.values(this.intervals).forEach((i) => clearInterval(i.interval))
   }
 
+  async getTotalObjectCount() {
+    /** This is fine, because it gets cached */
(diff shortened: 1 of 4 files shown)


didimitrie (approved):
lgtm - I've pinged @oguzhankoral as a reviewer for alignment with DUI3/sketchup. Just need to make sure we're not breaking sketchup progress reporting (though it doesn't look like it).

oguzhankoral (approved):

PR #2648, opened 2024-08-13
chore(liveness): fail liveness if db conns saturated for 10min

iainsproat:
## Description & motivation

Pods are placed in an unready state if the database connections are saturated for >20 seconds.

We currently have a situation where there are some long-running database queries. These cause the database connections available to a pod to be saturated for extended periods of time. This in turn causes the pod to be in an unready state. Over time, as the long-running queries accumulate, the number of pods which are unready increases. This eventually causes the entire cluster to be unavailable.

Currently the only way to recover is to restart the pod.

Including a check of saturated database connections in the liveness probe is a short-term hack to ensure affected pods are restarted whenever the encounter this issue.

## Checklist:


fabis94 (approved):
🤷

gjedlicska (approved):

PR #2656, opened 2024-08-14
Feat: Add guest table to members settings page

Mikehrn:
Simple table for displaying the guests tab, it's quite similar to Members but this table will be adjusted with different content


packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
```diff
@@ -0,0 +1,81 @@
+<template>
```
fabis94, line 1: Why not re-use MembersTable and just pass in a prop that does the minor adjustments that need to be made in the guests table (e.g. in the text labels)?

fabis94 (changes requested):

packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
fabis94: For the most part its the same exact component

Mikehrn: @fabis94 It was I did originally, but it only exactly the same temporarily, when we decide on the exact content we will we need to separate it anyway, so I already separated it so next time we just need change/update the queries/table etc

fabis94: Is it really gonna be that different? I'd think that only the labels might change and maybe some of the actions. If you really think its gonna be different enough to require its own component, then we can merge this.

fabis94 (approved):

packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
Mikehrn: Probably there will be different content in the table (like listening the parts this guest has access to), it's not very clear yet, if it's end up still being 99% the same I can still combine them :)

andrewwallacespeckle (dismissed):

andrewwallacespeckle (approved):

PR #2657, opened 2024-08-14
Alessandro/web 1594 calculate workspace cost

alemagio:
## Checklist:


packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -157,6 +157,7 @@ type Workspace {
     cursor: String
     filter: WorkspaceProjectsFilter
   ): ProjectCollection!
+  cost: Int!
```
fabis94, line 160: It it really just an integer? Not a structured object of price points and whatever? I don't understand what this is supposed to represent and other API consumers won't either - could you add a docblock to this field?

fabis94: Also - this might need @hasWorkspaceRole(ADMIN) on it, we probably don't want this to be shown to anyone else

packages/server/modules/workspaces/domain/operations.ts
```diff
@@ -96,6 +96,12 @@ export type GetWorkspaceRolesForUser = (
 
 export type UpsertWorkspaceRole = (args: WorkspaceAcl) => Promise<void>
 
+export type GetWorkspaceRolesCount = ({
```
fabis94, line 99: 1. Convert this to getting role counts for multiple workspaces (getWorkspacesRolesCount: () => Promise<Array<{workspaceId, role, count}>>
2. Then you can create a dataloader and use that in the Workspace.cost resolver

Usually converting the query from doing a select for a single entity to a select for a bunch of entities is not very complicated, and gives us an easy performance win in that we have a batched Workspace.cost() resolver that doesn't invoke the query separately for each requested workspace. And if you're ever interested in just 1 workspace, you can still use the query for that.

This should be whenever possible for any queries that are invoked in GQL field resolvers - if you can use a dataloader, do that

packages/server/modules/workspaces/helpers/cost.ts
```diff
@@ -0,0 +1,22 @@
+import { WorkspaceRoles } from '@speckle/shared'
+import { WorkspaceInvalidRoleError } from '@/modules/workspaces/errors/workspace'
+
+export const WORKSPACE_ADMIN_COST = 50
```
fabis94, line 4: Units of what? Could you add a docblock explaining these values?

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -220,3 +221,14 @@ export const workspaceInviteValidityFilter: InvitesRetrievalValidityFilter = (q)
       ).orWhereNotNull(Workspaces.col.id)
     })
 }
+
+export const getWorkspaceRolesCountFactory =
+  ({ db }: { db: Knex }): GetWorkspaceRolesCount =>
+  async ({ workspaceId }) => {
+    return await tables
```
fabis94, line 228: Easy to convert to retrieving roles for multiple workspaces:
```
WorkspacesACL.knex()
.select([workspaceId, role, knex.raw('count(*))])
.whereIn(workspaceId, WORKSPACEIDS)
.groupBy([workspaceId, role])
```

packages/server/modules/workspaces/tests/integration/repositories.spec.ts
```diff
@@ -281,4 +283,61 @@ describe('Workspace repositories', () => {
       await expectToThrow(() => upsertWorkspaceRole(role))
     })
   })
+
+  describe.only('getWorkspaceRolesCountFactory creates a function, that', () => {
```
fabis94, line 287: describe.only

If you want to only run a specific block, use the IDE test explorer buttons or run the test w/ `yarn test --grep="filter"`

packages/server/modules/workspaces/tests/integration/repositories.spec.ts
```diff
@@ -281,4 +283,61 @@ describe('Workspace repositories', () => {
...
+    it('returns an empty array with no users in workspace', async () => {
+      expect(
+        await getWorkspaceRolesCountFactory({ db })({
+          workspaceId: createRandomPassword()
+        })
+      ).to.deep.eq([])
+    })
+    it('returns an array containing counts for each role', async () => {
+      const { id: userId1 } = await createAndStoreTestUser()
```
fabis94, line 296: 1. There's a `createTestUsers` test util for batching up these calls
2. Or do Promise.all()?

packages/server/modules/workspaces/tests/integration/repositories.spec.ts
```diff
@@ -281,4 +283,61 @@ describe('Workspace repositories', () => {
...
+      const { id: userId2 } = await createAndStoreTestUser()
+      const { id: userId3 } = await createAndStoreTestUser()
+      const { id: userId4 } = await createAndStoreTestUser()
+
+      const { id: workspaceId } = await createAndStoreTestWorkspace()
+
+      await upsertWorkspaceRole({
```
fabis94, line 303: Promise.all

fabis94 (changes requested):

packages/server/modules/workspaces/domain/operations.ts
alemagio: I considered it, but for now, I decided to skip it for the following reasons:
1. I cannot directly use the dataloader, as we do in other situations, since I need to do some calculations. Considering our agreement, we want those calculations to be in the service, not the resolver or the repository. 
2. this part will be deleted once we will ship a proper pricing solution to the users and the workspace module will be a paying module.
3. We will show this banner in the workspace detail, so a DL wouldn't help

alemagio:
Closing this as it is replaced by https://linear.app/speckle/issue/WEB-1657/backend

PR #2658, opened 2024-08-14
Fix: Moved coming soon pages from workspaces to server

Mikehrn:
I added these incorrectly to workspace settings, should be under the server settings menu group. Didn't add redirect in the nuxt config for them as they shouldn't have direct URLs yet


fabis94 (approved):

Mikehrn:
> @gjedlicska mentioned (in Standup notes) that this might not be required?

That's correct, I will close this PR

PR #2666, opened 2024-08-14
Feat: Add projects page to settings

Mikehrn:
Abstracts project table into a shared component, implemented that for server project and workspace projects. Some minor styling tweaks and fixes.


packages/frontend-2/components/settings/shared/projects/index.vue
```diff
@@ -0,0 +1,145 @@
+<template>
+  <div>
+    <div class="flex flex-col-reverse md:justify-between md:flex-row md:gap-x-4">
+      <div class="relative w-full md:max-w-md mt-6 md:mt-0">
+        <FormTextInput
+          name="search"
+          :custom-icon="MagnifyingGlassIcon"
+          color="foundation"
+          search
+          :show-clear="!!search"
```
fabis94, line 10: showClear doesn't have to be conditional, it already disappears if there's no content

fabis94 (dismissed):

packages/frontend-2/components/settings/shared/projects/index.vue
Mikehrn: Thanks, adjusted this :)

andrewwallacespeckle (approved):

PR #2668, opened 2024-08-15
Andrew/web 1338 implement new navigation

andrewwallacespeckle:
## Description & motivation
- Implement new sidebar
- Change default layout so main scrolls independently of the page
- Implement new style of Top Bar
- General Tidy Ups


packages/frontend-2/components/dashboard/Sidebar.vue
```diff
@@ -0,0 +1,178 @@
+<!-- eslint-disable vuejs-accessibility/no-static-element-interactions -->
```
andrewwallacespeckle, line 1: On mobile, the sidebar comes in from the side, over the top of the content. It will not fill the full width, instead a div with backdrop-blur is added between the menu and the main content. On mobile, you would expect to be able to tap this and close the menu, which I have done. 

I added v-keyboard-clickable, but it didn't stop flagging the usual eslint complaints. This item isn't even visible on devices with keyboards, so i just added these.

gjedlicska:
I've done some testing on this, looks great, works well.

packages/frontend-2/pages/index.vue
```diff
@@ -1,20 +1,159 @@
 <template>
-  <ProjectsDashboard v-if="isLoggedIn" />
-  <div v-else class="mx-auto">
-    <CommonLoadingIcon />
+  <div>
+    <Portal to="navigation">
+      <HeaderNavLink :to="homeRoute" name="Dashboard" hide-chevron :separator="false" />
+    </Portal>
+    <PromoBannersWrapper v-if="promoBanners.length" :banners="promoBanners" />
```
fabis94, line 6: There's some conflicts w/ dashboard.vue, it has a different header now in `main`

fabis94 (approved):
do test this out w/ feature flags off as well

packages/frontend-2/components/global/icon/Connectors.vue
```diff
@@ -0,0 +1,16 @@
+<template>
```
Mikehrn, line 1: Should we update this in the user nav as well so they are the same icon?

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -169,9 +166,9 @@ const workspaceMenuItemClasses = (
     targetWorkspaceId.value === workspaceId &&
     !disabled
   ) {
-    return 'bg-highlight-2 hover:!bg-highlight-2'
+    return true
   }
-  return ''
+  return false
```
Mikehrn, line 171: Think we don't even need the if-statement anymore here, just

```   
const workspaceMenuItemClasses = (
  itemKey: string | number,
  workspaceId: string,
  disabled?: boolean
) =>
  targetMenuItem.value === itemKey &&
  targetWorkspaceId.value === workspaceId &&
  !disabled
```

packages/ui-components/src/components/layout/sidebar/menu/group/Item.vue
```diff
@@ -35,7 +34,20 @@
       ]"
       @click="toggleOpen"
     >
-      <ChevronDownIcon :class="[isOpen && 'rotate-180']" class="h-2.5 w-2.5" />
+      <svg
```
Mikehrn, line 37: Maybe create a component for this? Guess it's a custom chevron?

Mikehrn (changes requested):
Overall looks good :D Just some minor comments

packages/frontend-2/components/global/icon/Connectors.vue
andrewwallacespeckle: The icons were dropped from the user nav a while ago:
image

Unless you meant a different menu?

Mikehrn: Sorry you are right, in my mind it still had that cloud cloud icon haha

packages/frontend-2/pages/index.vue
```diff
@@ -1,20 +1,159 @@
...
+            :key="quickStartItem.title"
+            :title="quickStartItem.title"
+            :description="quickStartItem.description"
+            :buttons="quickStartItem.buttons"
+          />
+        </div>
+      </section>
+      <section>
+        <h2 class="text-heading-sm text-foreground-2">Recently updated projects</h2>
+        <div class="grid grid-cols-1 md:grid-cols-3 gap-3 pt-5">
```
Mikehrn, line 22: The empty state is missing here that was recently added, else it looks good 🚀

Mikehrn (changes requested):

Mikehrn (approved):

PR #2675, opened 2024-08-16
feat: accept & decline workspace invite as a registered member

fabis94:
This enables Accepting & Declining a workspace invite from the UI or email **for registered users only**. The flows for unregistered users are a separate task and I wanted to split it out to reduce the size of the PR.


packages/server/assets/serverinvites/typedefs/serverInvites.graphql
```diff
@@ -100,7 +100,7 @@ extend type Query {
   """
   Receive metadata about an invite by the invite token
   """
-  serverInviteByToken(token: String!): ServerInvite
+  serverInviteByToken(token: String): ServerInvite
```
fabis94, line 103: To allow for easier querying (both server & workspaceInvite w/ same query). Resolver just returns null if token empty.

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -512,6 +537,26 @@ export = FF_WORKSPACES_MODULE_ENABLED
 
           const invite = await ctx.loaders.invites.getInvite.load(inviteId)
           return invite?.token || null
+        },
+        email: async (parent, _args, ctx) => {
+          if (!parent.user) return parent.email
+
+          // TODO: Tests to check token & email access?
```
fabis94, line 544: Need to add tests for all of the changed backend logic, but considering the lack of time I'd like to approach this after

packages/frontend-2/components/workspace/invite/Banners.vue
```diff
@@ -0,0 +1,27 @@
...
+  <div class="flex flex-col divide-y divide-outline-3">
+    <WorkspaceInviteBanner v-for="item in items" :key="item.id" :invite="item" />
+  </div>
+</template>
+<script setup lang="ts">
+import { graphql } from '~~/lib/common/generated/gql'
+import type { WorkspaceInviteBanners_UserFragment } from '~~/lib/common/generated/gql/graphql'
+
+/**
+ * TODO: Add this to new dashboard page and remove from projects dashboard
```
fabis94, line 11: Currently all banners show up in projects page (current dashboard), but they should be moved to the new dashboard once its in

packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -11,10 +11,15 @@
...
       />
-      <ProjectsNewSpeckleBanner
-        v-if="showNewSpeckleBanner"
-        @dismissed="onDismissNewSpeckleBanner"
-      />
+      <WorkspaceInviteBanners
+        v-if="projectsPanelResult?.activeUser?.workspaceInvites?.length"
+        :invites="projectsPanelResult.activeUser"
+      >
+        <ProjectsNewSpeckleBanner
```
andrewwallacespeckle, line 18: Just checking that you meant to nest ProjectsNewSpeckleBanner in WorkspaceInviteBanner. 

ProjectsNewSpeckleBanner is removed in another branch anyway

packages/frontend-2/components/workspace/invite/Banners.vue
andrewwallacespeckle: The new dashboard is in at /dashboard/. But this can be moved in another ticket as you say

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -1,17 +1,34 @@
+import type { RouteLocationNormalized } from '#vue-router'
```
andrewwallacespeckle, line 1: Not a blocker, but what's the difference in this and simply `vue-router`?

andrewwallacespeckle (dismissed):
The FE2 changes look all good, just a few questions, but nothing blocking the merge.

cdriesler (approved):

packages/frontend-2/lib/workspaces/composables/management.ts
fabis94: @andrewwallacespeckle Auto-import imported it that way. The hashtag prefix means that there's an import map/alias defined there, maybe it resolves to different packages in CJS & ESM or Browser & Node. In this case it's not very important cause just `vue-router` works as well

packages/frontend-2/components/projects/Dashboard.vue
fabis94: fixed

andrewwallacespeckle (dismissed):

packages/frontend-2/components/workspace/invite/Banners.vue
fabis94: I've added the banners there as well

andrewwallacespeckle (approved):

PR #2695, opened 2024-08-19
Feat: add edit workspace avatar

Mikehrn:
Adds the input field for workspace avatar. Note: still falls back on the initials/avatar icon when no logo is set, we will address this behaviour when we add the space invader placeholders (basically it will never be empty, and an image will always be provided), but this allows us to at least adjust the workspace logo for now. 

Screenshot 2024-08-20 at 00 32 31


packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -55,6 +55,9 @@
               :title="workspaceItem.name"
               collapsible
             >
+              <template #title-icon>
+                <UserAvatar :logo="workspaceItem.logo" size="sm" />
```
Mikehrn, line 59: This component can be reused with one minor change to also work for workspace avatar, `UserAvatar` is a bit misleading in that sense, but also dont know if it's worth to update the naming across all components (and types/composables etc)

fabis94 (dismissed):

fabis94 (approved):

PR #2700, opened 2024-08-20
feat(fe2): workspace role select in project invite dialog

fabis94:
image

Role select only appears in workspace projects


andrewwallacespeckle (dismissed):
Looks good

PR #2712, opened 2024-08-20
chore(webhooks): update webhook function refactor

alemagio:
## Checklist:


packages/server/modules/webhooks/domain/operations.ts
```diff
@@ -3,12 +3,26 @@ import { Webhook } from '@/modules/webhooks/domain/types'
...
 }: Pick<Webhook, 'streamId'>) => Promise<number>
 
 export type GetWebhookById = ({ id }: Pick<Webhook, 'id'>) => Promise<Webhook | null>
+
+export type UpdateWebhook = ({
+  webhookId,
+  webhookInput
+}: {
+  webhookId: string
+  webhookInput: Pick<Webhook, 'updatedAt'> &
```
gjedlicska, line 21: updatedAt, is prob not needed here, it should be the service's responsibility to craft that date properly imho

gjedlicska (changes requested):

packages/server/assets/webhooks/typedefs/webhooks.graphql
```diff
@@ -84,7 +84,7 @@ input WebhookUpdateInput {
   description: String
   secret: String
   enabled: Boolean
-  triggers: [String]
+  triggers: [String!]
```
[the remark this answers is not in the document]
alemagio: By doing that we want always an array, we don't accept undefined or null
which I don't think it's the case for the current FE implementation right?

I mean, we accept:
- null
- undefined
- array with non null or undefined value

https://stackoverflow.com/a/46771890

packages/server/modules/webhooks/domain/operations.ts
alemagio: yes, that's why it is required.
I mean the type says that the `updatedAt` field is required.
Does it make sense?

packages/server/assets/webhooks/typedefs/webhooks.graphql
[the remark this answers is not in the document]
gjedlicska: disregard this, now i now, this is the update, we don't need to for this to be defined

packages/server/modules/webhooks/repositories/webhooks.ts
```diff
@@ -55,3 +56,21 @@ export const getWebhookByIdFactory =
 
     return { ...webhook, triggers: toTriggersArray(webhook.triggers) }
   }
+
+export const updateWebhookFactory =
+  ({ db }: { db: Knex }): UpdateWebhook =>
+  async ({ webhookId, webhookInput }) => {
+    const { triggers, ...update } = webhookInput
+    if (triggers) {
+      ;(
```
gjedlicska, line 65: this looks very confusing

gjedlicska (approved):

PR #2720, opened 2024-08-21
fix(graphql websocket): keep alive websocket connections if idle

iainsproat:
## Description & motivation

- if no data is being sent on a websocket connection then it was closed by nginx after 60s
- it is assumed to be better to keep alive the connection than have the overhead of closing and re-establishing the connection every 60s

## Changes:

- keepAlive parameter configured for websocket connections to graphql.

## Checklist:

### DIFF
--- a/packages/server/app.ts
+++ b/packages/server/app.ts
@@ -230,7 +230,8 @@ function buildApolloSubscriptionServer(
         }
 
         return baseParams
-      }
+      },
+      keepAlive: 30000 //milliseconds. Loadbalancers may close the connection after inactivity. e.g. nginx default is 60000ms.
     },
     {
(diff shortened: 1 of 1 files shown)


packages/server/app.ts
```diff
@@ -230,7 +230,8 @@ function buildApolloSubscriptionServer(
         }
 
         return baseParams
-      }
+      },
+      keepAlive: 30000 //milliseconds. Loadbalancers may close the connection after inactivity. e.g. nginx default is 60000ms.
```
gjedlicska, line 234: isn't the goal here, to keep the conn alive for longer than 60 secs, that is respected by nginx too?

PR #2724, opened 2024-08-21
feat(dui3): allows model cards to be added by url (CNX-235)

didimitrie:
closes off [CNX-235: Support adding models by URL](https://linear.app/speckle/issue/CNX-235/support-adding-models-by-url).

Allows users to input model urls (and model version urls) and this will be take them to the "last" step in the wizard process, depending on whether they're sending/receiving and/or have a version url. This is done intentionally so the user is in control and needs to "confirm" their actions with a click. 

ht to @oguzhankoral for the initial work on this! 

I have tested this, across send/receive, against: 
- invalid urls (non-speckle)
- valid urls, with no access to project
- valid urls

We will only display an error if 
- an invalid url
- there is no account found for that server
- there are no permissions to send to that ~~stream~~ project

@oguzhankoral for easy testing: wait for the deploy preview url, hack it into any existing connector with right click and the console, and let rip!


packages/dui3/components/send/Wizard.vue
```diff
@@ -98,6 +103,20 @@ const selectedModel = ref<ModelListModelItemFragment>()
...
+const { tryParseUrl, urlParsedData, urlParseError } = useAddByUrl()
+const updateSearchText = (text: string | undefined) => {
+  urlParseError.value = undefined
+  if (!text) return
+  tryParseUrl(text, 'sender')
+}
+
+watch(urlParsedData, (newVal) => {
+  if (!newVal) return
+  console.log(newVal)
```
oguzhankoral, line 115: probably cleanup forgotten

packages/dui3/components/wizard/ProjectSelector.vue
```diff
@@ -119,15 +121,19 @@ const props = withDefaults(
...
 const showNewProjectDialog = ref(false)
 const accountStore = useAccountStore()
 const { activeAccount } = storeToRefs(accountStore)
 
 const accountId = computed(() => activeAccount.value.accountInfo.id)
 const selectedAccountId = ref<string>()
 
+watch(searchText, () => {
+  newProjectName.value = searchText.value
+  console.log('adsf')
```
oguzhankoral, line 133: probably cleanup forgotten

oguzhankoral (changes requested):
minor logging cleanup needed. otherwise tested all potential scenarios and good to go

oguzhankoral (approved):

PR #2726, opened 2024-08-21
Feat: Added delete user from workspace button

Mikehrn:
Adds the button to delete a user from a workspace

### DIFF
--- a/packages/frontend-2/components/settings/shared/DeleteUserDialog.vue
+++ b/packages/frontend-2/components/settings/shared/DeleteUserDialog.vue
@@ -1,0 +1,40 @@
+<template>
+  <LayoutDialog v-model:open="open" max-width="sm" :buttons="dialogButtons">
+    <template #header>Remove user</template>
+    <div class="flex flex-col gap-4 text-body-xs text-foreground">
+      <p>Are you sure you want to remove the following user from the workspace?</p>
+      <p class="font-semibold">
+        {{ name }}
+      </p>
+    </div>
+  </LayoutDialog>
+</template>
+
+<script setup lang="ts">
+import type { LayoutDialogButton } from '@speckle/ui-components'
+
+const emit = defineEmits<{
+  (e: 'removeUser'): void
+}>()
+
+defineProps<{
+  name: string
+}>()
+const open = defineModel<boolean>('open', { required: true })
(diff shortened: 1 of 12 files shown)


packages/ui-components/src/components/global/ToastRenderer.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     aria-live="assertive"
-    class="pointer-events-none fixed top-0 right-0 left-0 bottom-0 flex items-end px-4 py-6 mt-10 sm:items-start sm:p-6 z-50"
+    class="pointer-events-none fixed top-0 right-0 left-0 bottom-0 flex items-end px-4 py-6 mt-10 sm:items-start sm:p-6 z-[60]"
```
Mikehrn, line 4: No huge fan of doing it this way, for it's the only way to have the toasts above the dialogs for now

packages/frontend-2/components/settings/shared/DeleteUserDialog.vue
```diff
@@ -0,0 +1,40 @@
+<template>
+  <LayoutDialog v-model:open="open" max-width="sm" :buttons="dialogButtons">
+    <template #header>Remove user</template>
+    <div class="flex flex-col gap-4 text-body-xs text-foreground">
+      <p>Are you sure you want to remove the following user from the workspace?</p>
+      <p class="font-semibold">
```
andrewwallacespeckle, line 6: font-medium

packages/ui-components/src/components/global/ToastRenderer.vue
andrewwallacespeckle: I don't have an issue with this. If there are meaningless z-index's it can get annoying, but we actually need all these levels of elevation, so this is fine for me.

andrewwallacespeckle (changes requested):

packages/frontend-2/components/settings/shared/DeleteUserDialog.vue
Mikehrn: Updated 👍 Also changed it in a couple other places in settings

andrewwallacespeckle (approved):

PR #2735, opened 2024-08-22
feat(fe2) Implement workspaces explainer page

andrewwallacespeckle:
## Description & motivation
The content is not finalised, but the page can go in


packages/frontend-2/components/workspace/CreateDialog.vue
```diff
@@ -39,6 +39,10 @@ import { isRequired, isStringOfLength } from '~~/lib/common/helpers/validation'
 
 type FormValues = { name: string; description: string }
 
+const props = defineProps<{
+  navigateOnSuccess?: boolean
```
andrewwallacespeckle, line 43: I made it so we have an option to navigate to the workspace when its created. Defaults to false

packages/frontend-2/components/workspaces/Promo/Card.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <div class="flex flex-col border border-outline-3 rounded-lg pt-6 px-8 pb-10 h-full">
```
Mikehrn, line 2: Wonder if we can reuse the QuickStart card from the dashboard here if we add a icon slot, I think they are basically the same? Maybe we just need to rename it

packages/frontend-2/components/workspaces/Promo/Banner.vue
```diff
@@ -0,0 +1,45 @@
...
+      <div
+        class="absolute top-0 right-0 h-full w-7/12 bg-gradient-to-r from-foundation-page"
+      >
+        <div class="h-full w-64 bg-gradient-to-r from-foundation-page" />
+      </div>
+    </div>
+
+    <div class="relative z-10 text-foreground max-w-lg">
+      <div class="flex gap-2 items-center">
+        <h3 class="text-heading-lg">Introducing Workspaces</h3>
```
Mikehrn, line 21: Small thing; `Introducing Workspaces` -> `Introducing workspaces`

packages/frontend-2/components/workspaces/Promo/Page.vue
```diff
@@ -0,0 +1,73 @@
+<template>
```
Mikehrn, line 1: I know we do this in other pages as well, but is there any specific reason why we have a separate component for pages, instead of use just using the page .vue file?

Mikehrn (changes requested):

packages/frontend-2/components/workspaces/Promo/Page.vue
andrewwallacespeckle: The only reason I did in in this branch is so we can move this easily. Currently, it lives at /workspaces/, but in future this will be a list of workspaces

packages/frontend-2/components/workspaces/Promo/Card.vue
andrewwallacespeckle: There are a few differences here. The padding is different, there is no CTA and the icon as you say. This is also likely to change as the content/design is not finalised. 

I also don't really like using components from other folders too much. At some point when either needs updated it causes more problems, because you don't expect QuickStartCard to be used elsewhere.

Mikehrn: I agree with the second point, however having multiple cards with is just a card with a border also isn't ideal, having icons/buttons optional isn't too bad, if we turn this into a more generic dashboard card (or whatever context) it doesn't have to cause problems :) We can leave it for now, but if we get more of these lets find a way to align them them (both design and FE)

Mikehrn (dismissed):

packages/frontend-2/components/workspaces/Promo/Card.vue
andrewwallacespeckle: Created CommonCard and used it here

Mikehrn (approved):

PR #2752, opened 2024-08-24
fix(workspaces): transact workspace role changes

cdriesler:
## Description & motivation

Currently, it's possible to get in a broken role state with workspace projects:

- User A creates a workspace
- User A creates a project in the workspace
  - User A then is given `owner` role on project
- User A invites User B to workspace as admin
  - User B then is given `owner` role on project
- User B changes User A to `reviewer` (or non-owner) on project
- User B leaves workspaces
  - User B workspace role successfully removed
  - User B project role removal throws (cannot remove last `owner` from project)

User B then in:

- Not a workspace member
- Still an `owner` on the workspace project

Something like:
```
      // User   Workspace Role    Project Role
      // A        Admin                  Reviewer
      // B        Admin                  Owner
      //
      // Action: `testMemberUser` leaves workspace
```

There's several ways to end up in this situation but the important part is that, because workspace roles and project roles are updated separately (with different rules) in one operation, it's possible to fail "in-between" either (or between project role changes, if there are multiple projects in a workspace).

## Changes:

- Prevents `grantStreamPermissions` from "promoting" the last admin out of a project
- Adds optional knex transaction parameter to workspace and project role changes
  - Pattern borrowed from other similar cases in code, very willing to do it in different shape/style
- Uses the transaction in workspace `deleteWorkspaceRole` and `updateWorkspaceRole` service functions

Behavior validated in new tests. This suite will be made more thorough with the upcoming (pre-teams) upgrades to how roles interact.


gjedlicska:
Lets discuss this with @alemagio, cause the implementation leaks `knex` into the domain, which should never happen

alemagio:
> Lets discuss this with @alemagio, cause the implementation leaks `knex` into the domain, which should never happen

If I do remember correctly we already discussed this situation.
The solution should be simple:
- you don't need to pass the transaction as an option
- you can simply pass the transaction itself instead of passing the db dependency

The transaction object has the same API as the knex one
https://knexjs.org/guide/transactions.html

cdriesler:
Roger, that's really nice will do that

cdriesler:
back at you gang @gjedlicska @alemagio @iainsproat 

the `withTransition` addition is neat but maybe this eventually gets upgraded to live in the gql `context` somehow?

packages/server/modules/workspaces/services/management.ts
```diff
@@ -249,6 +249,7 @@ export const deleteWorkspaceRoleFactory =
     // Perform delete
     const deletedRole = await deleteWorkspaceRole({ userId, workspaceId })
     if (!deletedRole) {
+      // await trx.commit()
```
gjedlicska, line 252: its this await needed here or is this just a leftover @cdriesler ?

gjedlicska:
In general i think I'm ok with the current implementation. Moving forward, i think we need to addopt a form of the UnitOfWork pattern, that should move these responsibilities out of the app context. This is more of a concern of domain correctness, so it should be part of the "business logic package" not the application itself.

cc @alemagio

gjedlicska (approved):

PR #2767, opened 2024-08-27
chore(errors): thrown errors should derive from BaseError and appropriate status code used

iainsproat:
## Description & motivation

This PR applies the following principles related to errors:

1. All errors classes extended from `BaseError` should define an http status code.

2. All thrown errors should instantiate from an error class derived from `BaseError`. This should be applied regardless of whether the error is used in handling an http request, as it allows the correct log level to be selected.

3. Errors related to user input should throw an error with a 4XX http status code. Otherwise, 5XX.

## Changes:

- status codes are added to all errors extended from `BaseError`.
- Errors where stream, projects, invites, commits etc. cannot be found now result in a X_NOT_FOUND error code. This makes it clearer as to how the client can resolve the issue.
- replaces `throw new Error` with a specific Error class. Again, making it clearer as to what the specific error is and how it might be resolved.
- Where error messages include dynamic context, these values are added to the `options.info` parameter when the error is constructed.

## Checklist:

### DIFF
--- a/packages/server/modules/comments/tests/comments.spec.ts
+++ b/packages/server/modules/comments/tests/comments.spec.ts
@@ -478,7 +478,9 @@ describe('Comments @comments', () => {
       .then(() => {
         throw new Error('This should have been rejected')
       })
-      .catch((error) => expect(error.message).to.be.equal('Object not found'))
+      .catch((error) =>
+        expect(error.message).to.be.equal('Object not found in the specified stream')
+      )
 
     // create a comment on streamA but commitB
     await createComment({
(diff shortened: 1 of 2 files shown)


packages/server/modules/workspaces/services/invites.ts
```diff
@@ -505,14 +509,14 @@ export const validateWorkspaceInviteBeforeFinalizationFactory =
       userId: finalizerUserId
     })
     if (!workspace) {
-      throw new InviteFinalizingError(
+      throw new WorkspaceNotFoundError(
```
fabis94, line 512: Some of these changed errors may break FE code that relies on a specific error code being returned, not sure, but if not needed I would hold off against changing already BaseApp-extending error types to different ones

fabis94: Or you should do a search of the code value inside `frontend-2` to see if its referenced anywhere

PR #2776, opened 2024-08-27
Add Plus CTA to SidebarGroup

andrewwallacespeckle:
## Description & motivation
Move the "Add Workspace" button to the LayoutSidebarGroup, as per the designs. 

image

### DIFF
--- a/packages/frontend-2/components/dashboard/Sidebar.vue
+++ b/packages/frontend-2/components/dashboard/Sidebar.vue
@@ -64,7 +64,7 @@
                     }
                   : undefined
               "
-              plus-text="Add workspace"
+              plus-text="Create workspace"
             >
               <NuxtLink :to="workspacesRoute">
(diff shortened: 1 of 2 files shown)


andrewwallacespeckle:
> Looks good :) We will also need to enable it in the settings sidebar tho

The designs showed the "+" in the main sidebar, but the button in the settings sidebar. I wasn't sure if this was a mistake in the Figma or not, so I'm going to check with Michal tomorrow.

Mikehrn:
> > Looks good :) We will also need to enable it in the settings sidebar tho
> 
> The designs showed the "+" in the main sidebar, but the button in the settings sidebar. I wasn't sure if this was a mistake in the Figma or not, so I'm going to check with Michal tomorrow.

In that case ignore my comment :) I think in settings it will be still nice to have also as a plus (the sidebar gets quite full), but we can follow up on that later if we wanna change that

PR #2784, opened 2024-08-28
feat(fe2): Update Content. Remove check for workspaces.length

andrewwallacespeckle:
Updating content provided by Dim for Workspace Explainer. 

Remove conditional from Sidebar to allow users to see Workspaces, even if they have 0 workspaces.

### DIFF
--- a/packages/frontend-2/components/workspaces/Promo/Banner.vue
+++ b/packages/frontend-2/components/workspaces/Promo/Banner.vue
@@ -29,7 +29,7 @@
         <FormButton :disabled="!isAdmin" color="outline" @click="$emit('create')">
           Create workspace
         </FormButton>
-        <FormButton color="subtle">Learn more</FormButton>
+        <!-- <FormButton color="subtle">Learn more</FormButton> -->
       </div>
     </div>
   </div>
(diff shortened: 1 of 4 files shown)


packages/frontend-2/components/workspaces/Promo/Banner.vue
```diff
@@ -29,7 +29,7 @@
         <FormButton :disabled="!isAdmin" color="outline" @click="$emit('create')">
           Create workspace
         </FormButton>
-        <FormButton color="subtle">Learn more</FormButton>
+        <!-- <FormButton color="subtle">Learn more</FormButton> -->
```
gjedlicska, line 32: is this button not needed any more @andrewwallacespeckle ?

Mikehrn (dismissed):

packages/frontend-2/components/workspaces/Promo/Banner.vue
andrewwallacespeckle: There is nowhere for it to link to yet that I know of. Rather than have a button that does nothing, I commented it out until we clarify what this should do. I can uncomment and add a destination in another branch rather than block this one.

Mikehrn (approved):

PR #2798, opened 2024-08-28
Feat: Add Mixpanel event for workspaces

Mikehrn:
More events will be added, but are dependent on other branches getting merged in. Added workspace_id where known to add it to the user group


gjedlicska (approved):
lgtm

benjaminvo (approved):
Nice! I like the event names

PR #2820, opened 2024-08-29
Feat: Add role label, version count, and update styling of project header

Mikehrn:
Updated the header styling of project and workspace to align with the design and display the label with the role. We can later refactor this a bit to turn this into a shared header component

Screenshot 2024-08-29 at 22 30 41


packages/frontend-2/pages/projects/[id]/index.vue
```diff
@@ -8,10 +8,19 @@
...
-        class="flex flex-col md:flex-row md:justify-between md:items-start gap-8 mb-6 mt-4 md:my-6"
+        class="flex flex-col md:flex-row md:justify-between md:items-center gap-4 my-2"
       >
         <ProjectPageHeader :project="project" />
-        <ProjectPageTeamBlock :project="project" class="w-full md:w-72 shrink-0" />
+        <div class="flex gap-x-4 items-center">
+          <div
+            class="text-body-3xs bg-foundation-2 text-foreground-2 rounded px-3 py-1 font-medium select-none whitespace-nowrap"
+          >
+            {{ project.modelCount.totalCount || 0 }} Project{{
```
Comment (andrewwallacespeckle) [inline packages/frontend-2/pages/projects/[id]/index.vue:18]: Shouldn't this be Model rather than Project

andrewwallacespeckle (changes requested):
All good apart from that 1 comment

packages/frontend-2/pages/projects/[id]/index.vue
Comment (Mikehrn) [reply packages/frontend-2/pages/projects/[id]/index.vue]: Yup nice catch thanks :) Also added the versions counter in this branch, had to add a bunch of flex classes and wrapper but it was the only way, as all the alignments are quite different on mobile/desktop and with/without counter

Screenshot 2024-08-30 at 00 42 55

andrewwallacespeckle (dismissed):

andrewwallacespeckle (approved):

PR #2823, opened 2024-08-30
feat(fe2): Workspace members table. Update Role

andrewwallacespeckle:
Moving the Role select to it's own dialog.


packages/frontend-2/components/settings/shared/ChangeRoleDialog.vue
```diff
@@ -2,38 +2,49 @@
...
-          <span>{{ getRoleLabel(newRole).title }}</span>
-        </div>
-      </div>
+      <p>
+        Select a new role for
+        <strong>{{ name }}</strong>
+        :
+      </p>
+      <FormSelectWorkspaceRoles
+        :disabled="!isWorkspaceAdmin"
```
Mikehrn, line 11: Maybe instead of disabling here we can disable it it on the parent?

packages/frontend-2/components/settings/shared/ChangeRoleDialog.vue
```diff
@@ -2,38 +2,49 @@
...
-  oldRole?: WorkspaceRoles
-  newRole?: WorkspaceRoles
+  oldRole: WorkspaceRoles
+  isWorkspaceAdmin: boolean
+  workspaceDomainPolicyCompliant: boolean
 }>()
+
 const open = defineModel<boolean>('open', { required: true })
 
+const localOldRole = ref(props.oldRole)
```
Mikehrn, line 40: I think if we have a model we can avoid setting it in here

packages/frontend-2/components/settings/workspaces/General/LeaveDialog.vue
```diff
@@ -41,7 +42,9 @@ graphql(`
 `)
 
 const props = defineProps<{
-  workspace: SettingsWorkspaceGeneralDeleteDialog_WorkspaceFragment
+  workspace:
```
Mikehrn, line 45: I think you can leave this prop as the current fragment, and then just include `SettingsWorkspaceGeneralDeleteDialog_WorkspaceFragment` in the members table

Mikehrn (changes requested):

Mikehrn (approved):

PR #2824, opened 2024-08-30
fix(server): auto-verify on invited server registration

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/auth/strategies/azureAd.ts
+++ b/packages/server/modules/auth/strategies/azureAd.ts
@@ -31,6 +31,7 @@ import {
 } from '@/modules/shared/helpers/envHelper'
 import type { Request } from 'express'
 import { ensureError, Optional } from '@speckle/shared'
+import { ServerInviteRecord } from '@/modules/serverinvites/domain/types'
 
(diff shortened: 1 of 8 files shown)


packages/server/modules/auth/strategies/azureAd.ts
```diff
@@ -128,50 +129,29 @@ const azureAdStrategyBuilder: AuthStrategyBuilder = async (
           return next()
         }
 
-        // if the server is not invite only, go ahead and log the user in.
```
fabis94, line 131: @gjedlicska It seems that there was a long-standing bug in SSO registration - if the server isn't inviteOnly, then the active invite wasn't used at all, the user was registered without it.

I fixed this in all SSO strategies, but let me know if you think that the way it was before was actually intentional

alemagio (approved):

gjedlicska (approved):

packages/server/modules/auth/strategies/azureAd.ts
gjedlicska: Lets test this on latest

PR #2826, opened 2024-08-30
fix: support accepting invites even if a workspace member already

fabis94:


Mikehrn (approved):
FE looks good :)

packages/server/modules/workspaces/services/invites.ts
```diff
@@ -492,11 +499,13 @@ export const validateWorkspaceInviteBeforeFinalizationFactory =
...
-          'Attempting to finalize invite to a workspace that the user already has access to'
-        )
-      }
+      // We now allow accepting an invite even if the user is already a member of the workspace
+      // (to get extra emails, for example)
+      // if (workspace.role) {
+      //   throw new InviteFinalizingError(
+      //     'Attempting to finalize invite to a workspace that the user already has access to'
+      //   )
+      // }
```
alemagio, line 508: leftover?

alemagio (approved):

packages/server/modules/workspaces/services/invites.ts
fabis94: My intention was to leave that in in case we want to uncomment it back in 🤔

gjedlicska (approved):

PR #2829, opened 2024-08-30
fix(fe2): sidebar flashing in post page load

fabis94:
caused by SSR-incompatible code


fabis94:
@Mikehrn would be great if you'd test this out locally, just to double check collapsing works as expected (seems that way to me)

Mikehrn:
> @Mikehrn would be great if you'd test this out locally, just to double check collapsing works as expected (seems that way to me)

Just tried it, works as expected!

Mikehrn (approved):

PR #2830, opened 2024-08-30
fix(fe2): Don't trigger shortcuts while typing comments

andrewwallacespeckle:
## Description & motivation
Currently, when typing a comment, you can accidentally trigger the sidebar shortcuts. 

Now, we check if the user is typing a comment before we execute the shortcut action.

### DIFF
--- a/packages/frontend-2/components/viewer/Controls.vue
+++ b/packages/frontend-2/components/viewer/Controls.vue
@@ -422,10 +422,14 @@ const measureShortcut = ref(
   `Measure mode ${getShortcutTitle(ViewerKeyboardActions.ToggleMeasurements)}`
 )
 
-const isTyping = computed(() => ui.threads.openThread.isTyping.value)
+const isTypingComment = computed(() => {
+  const isNewThreadEditorOpen = ui.threads.openThread.newThreadEditor.value
+  const isExistingThreadEditorOpen = !!ui.threads.openThread.thread.value
+  return isNewThreadEditorOpen || isExistingThreadEditorOpen
+})
 
 const handleKeyboardAction = (action: ViewerKeyboardActions) => {
-  if (isTyping.value) {
+  if (isTypingComment.value) {
     return
   }
   switch (action) {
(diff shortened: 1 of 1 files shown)


packages/frontend-2/components/viewer/Controls.vue
```diff
@@ -419,7 +422,12 @@ const measureShortcut = ref(
   `Measure mode ${getShortcutTitle(ViewerKeyboardActions.ToggleMeasurements)}`
 )
 
+const isTyping = computed(() => ui.threads.openThread.isTyping.value)
+
 const handleKeyboardAction = (action: ViewerKeyboardActions) => {
+  if (isTyping.value) {
```
fabis94, line 428: isTyping is not fully accurate - its got a debounce timeout of 7 seconds, so 7 seconds after you're done typing you're still shown as typing. Also you're shown as typing for the entire duration of the comment submission, so if you're on a slow internet connection that can block you as well. These kind of delays are bad UX imo and will cause users not to trust these shortcuts ("why tf isn't it working when it usually did??").

To properly fix this you'd probably need to introduce a new more accurate isTyping `isTypingRealtime` that has either no debounce at all or a very low one, and that only gets set/unset by typing, not by clicking submit in the comment editor.

While you're looking at this, also validate that `onKeyDownHandler` of `useIsTypingUpdateEmitter()` even works, I see that its attached to @keydown of `<ViewerCommentsEditor/>` but that component has no such event. So do double check what's going on there as well

fabis94 (changes requested):

packages/frontend-2/components/viewer/Controls.vue
andrewwallacespeckle: Thanks great feedback. 

A simpler solution is possibly blocking the shortcuts based on newThreadEditor rather than isTyping
image

Do you think this is an acceptable change to fix this, without the complications around debounce? I've testing and it's working, and I don't want to spend too long on this when I have some Workspaces work to do.

fabis94 (approved):

PR #2831, opened 2024-08-30
feat(fe2): greatly improved DX for apollo cache modification

fabis94:
(Built this when I had some downtime about a week ago, hope you like it :) )

There's 2 big parts to this:
1. A fully & automatically type-checked `modifyObjectField()` function. You no longer need to specify any generic arguments, it will figure out what part of the cache you're updating based on the cache reference and field name you pass into the function.
2. Improved and simplified helpers for common activities in `modifyObjectFIeld()`, most importantly introduced `createUpdatedValue` that updates the cached value at specific paths (e.g. `items.[0].name`) **ONLY** if such a value exists in the cache. This makes it easier to ensure we only update cached values, not introduce new ones (and get an invalid state that way - e.g. a totalCount of 0, when it was undefined before)

Demo of all of the TS auto typing you get out of the box in type name, field name, field value, variable value resolution
Code_pwLJH7zfr7

getCacheId() blocks invalid type names:
image

readField() allows reading fields of `Reference` objects, and its now fully typed - meaning it will understand what kind of reference are you specifying and only support fields that exist on that type
image
image

Reference objects are typed - you can no longer mismatch references, e.g. put a User reference where a Project reference is expected:
image
image


packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -143,33 +138,33 @@ export const useProcessWorkspaceInvite = () => {
...
               // Evict Query.workspace
-              modifyObjectField<Query['workspace'], QueryWorkspaceArgs>(
+              modifyObjectField(
                 cache,
                 ROOT_QUERY,
                 'workspace',
-                ({ variables, details: { DELETE } }) => {
-                  if (variables.id === workspaceId) return DELETE
+                ({ variables, helpers: { evict } }) => {
+                  if (variables.id === workspaceId) return evict()
```
fabis94, line 146: Return evict() replaces return DELETE

packages/frontend-2/tools/gqlCacheHelpersCodegenPlugin.js
```diff
@@ -0,0 +1,59 @@
+const { reduce } = require('lodash')
```
fabis94, line 1: Had to introduce a custom GQL codegen plugin to get this to work

packages/frontend-2/lib/common/helpers/graphql.ts
```diff
@@ -499,52 +529,182 @@ export const getDateCursorFromReference = (params: {
...
-    details: Parameters<Modifier<ModifyFnCacheData<FieldData>>>[1] & {
+    variables: Field extends keyof AllObjectFieldArgTypes[Type]
+      ? AllObjectFieldArgTypes[Type][Field]
+      : never
+    /**
+     * Value found in the cache. Read-only and should not be mutated directly. Use the
+     * createUpdatedValue() helper to build a new value with updated fields.
+     */
+    value: ReadonlyDeep<ModifyObjectFieldValue<Type, Field>>
+    helpers: {
```
fabis94, line 604: Access all of the helpers you need from the `helpers` object in the modifyObjectField() callback

packages/frontend-2/lib/common/helpers/graphql.ts
```diff
@@ -499,52 +529,182 @@ export const getDateCursorFromReference = (params: {
...
+      /**
+       * Build new value with the values at specific paths updated with the provided updater functions,
+       * ONLY if the paths exist in the cache.
+       *
+       * This function operates on a deeply cloned value that is safe to mutate
+       */
+      createUpdatedValue: (
```
fabis94, line 611: Simplifies this:
```ts
     modifyObjectField(
                cache,
                workspaceCacheId,
                'projects',
                ({ value, helpers: { ref } }) => {
                  const newItems = isUndefined(value?.items)
                    ? undefined
                    : [ref('Project', newProject.id), ...value.items]

                  const newTotalCount = isUndefined(value?.totalCount)
                    ? undefined
                    : (value.totalCount || 0) + 1

                  return {
                    ...value,
                    ...(isUndefined(newItems) ? {} : { items: newItems }),
                    ...(isUndefined(newTotalCount) ? {} : { totalCount: newTotalCount })
                  }
                }
              )
```

Into this:
```ts
modifyObjectField(
                cache,
                workspaceCacheId,
                'projects',
                ({ helpers: { ref, createUpdatedValue } }) => {
                  return createUpdatedValue(({ update }) => {
                    update('items', (items) => [
                      ref('Project', newProject.id),
                      ...items
                    ])
                    update('totalCount', (count) => count + 1)
                  })
                }
              )
```

update() calls are properly typed as well, and will only allow you to specify property paths that actually exist in `value`, and the property value in the callback is gonna be properly typed too
image
image

packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -170,35 +165,19 @@ onUserProjectsUpdate((res) => {
...
+      'projects',
+      ({ helpers: { ref, createUpdatedValue } }) =>
+        createUpdatedValue(({ update }) => {
+          update('items', (items) => [
+            ref('Project', incomingProject.id),
+            ...(items || [])
+          ])
+          update('totalCount', (count) => count + 1)
+        }),
+      { autoEvictFiltered: true }
```
fabis94, line 180: Option that automatically evicts all fields that have variables w/ filters in them. You usually want to do this when adding a new item to a list - it may be hard to tell if the item fits the filtered query, and it'll just be easier to evict that version of the query and refetch it 

cc @cdriesler

andrewwallacespeckle (approved):

Mikehrn (approved):
Nice 😄  Thanks for the visuals and clear explanations as well!

PR #2833, opened 2024-08-30
fix(workspaces): mutate on join workspace

cdriesler:
## Description & motivation

- Use new utils to also mutate on join

## Checklist:


packages/frontend-2/components/workspace/invite/DiscoverableWorkspaceBanner.vue
```diff
@@ -60,13 +65,31 @@ const processJoin = async (accept: boolean) => {
...
   const result = await apollo
     .mutate({
       mutation: DashboardJoinWorkspaceDocument,
       variables: {
         input: {
           workspaceId: props.workspace.id
         }
+      },
+      update(cache) {
+        modifyObjectField(
```
fabis94, line 80: You should always use the 2nd param of `update` - `data` - to get the actual data from the response and validate that the mutation did in fact go through successfully before you make any cache updates

fabis94 (changes requested):

packages/frontend-2/components/workspace/invite/DiscoverableWorkspaceBanner.vue
```diff
@@ -60,13 +65,31 @@ const processJoin = async (accept: boolean) => {
...
+          cache,
+          getCacheId('User', userId),
+          'workspaces',
+          ({ helpers: { createUpdatedValue, ref } }) => {
+            return createUpdatedValue(({ update }) => {
```
fabis94, line 85: @cdriesler User.workspaces has variables as well, you usually want to just evict all filtered (e.g. search specified) cache values cause there's no way to accurately tell if the new workspace fits that filter or not, and then make the cache update to the list only in non-filtered values

fabis94: Evict means - wherever that evicted value was referenced, e.g. in some kind of query living in a UI component, the query will understand that its now missing a part of its data and refetch it

cdriesler:
Ok, updated, but now it's pretty similar to the mutation when creating a workspace. Acceptable for now?

fabis94 (changes requested):
i think u need to merge main, there's conflicts and it shows my changes here

fabis94 (approved):

PR #2834, opened 2024-08-30
fix(workspaces): null billing for guests

cdriesler:
## Description & motivation

- Guests couldn't open workspace pages fully because `WorkspacePageQuery` requested `billing`, and `billing` field required a role of at least member or above

## Changes:

- Makes `billing` nullable
- Moves authz to resolver body
- Returns null if authz fails
- Also uses `codegen.yml` to assert return type of resolver instead of casting

## Checklist:


alemagio (approved):

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -276,7 +276,7 @@ type Workspace {
   """
   Billing data for Workspaces beta
   """
-  billing: WorkspaceBilling! @hasWorkspaceRole(role: MEMBER)
+  billing: WorkspaceBilling
```
fabis94, line 279: y

cdriesler: I could set it as requiring workspace guest, but the important bits are:

- Opting not to change frontend query based on role
- Returning `null` for billing if user is guest

We could also just return the billing information anyway for guests and continue not to show it on the frontend, but this seemed intentional to not expose that information to guests.

fabis94: Ok, that's fine.

As an FYI - you could've also set `errorPolicy: 'all'` in the FE to prevent the query from not loading all of the other fields because of the auth issue in regards to this specific field, this is also what we do for `invitedTeam`, another admin only field. The benefit there would be that the API would be clearer - instead of just returning null, it'd actually throw saying you don't have access, plus there's the directive in the typedef.

gjedlicska: If there is an easy schemantically correct frontend fix for this I'd like to do that. @fabis94 @cdriesler pls coordinate on this

cdriesler: @gjedlicska opting to do the `errorPolicy` thing

cdriesler:
Trying again in favor of simpler solution raised by @fabis94

PR #2835, opened 2024-08-30
fix(workspaces): guarantee that workspace admins are project owners

cdriesler:
## Description & motivation

- Because project roles can be changed independently of workspace roles, it's currently possible for a workspace admin to be removed from a workspace project.
- We think this is silly

## Changes:

- Route all workspace project changes through new `workspaceMutations.projects.updateRole` that has ✨extra protections✨
- Prohibit changing workspace admin roles here (i.e. cannot change their project admin status)
- Also disable the change dropdown on the frontend
- Add tooltip to give project owners a hint of why we're not letting them change roles here
- Bring workspace admins to top of project collaborators list (revertable change this just made sense to me at this brain state at this hour)


packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -623,6 +626,34 @@ export = FF_WORKSPACES_MODULE_ENABLED
           return ctx.loaders.workspaces!.getWorkspace.load(args.workspaceId)
         }
       },
+      WorkspaceProjectMutations: {
+        updateRole: async (_parent, args, context) => {
+          const { projectId, userId, role } = args.input
```
fabis94, line 631: 1. Why do we need a new resolver for project role updates?
2. If the old resolver would not work correctly in this scenario, what prevents an API consumer from using it anyway and introducing some kind of broken state?

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -623,6 +626,34 @@ export = FF_WORKSPACES_MODULE_ENABLED
...
+
+          const { workspaceId } = (await getStream({ streamId: projectId })) ?? {}
+
+          if (!workspaceId) {
+            // Project does not belong to a workspace
+            throw new Error()
```
fabis94, line 637: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#c680ec9cad8849baab5bc9016de08380

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -623,6 +626,34 @@ export = FF_WORKSPACES_MODULE_ENABLED
...
+          }
+
+          const currentRole = await getWorkspaceRoleForUserFactory({ db })({
+            workspaceId,
+            userId
+          })
+
+          if (currentRole?.role === Roles.Workspace.Admin) {
+            // User is workspace admin and cannot have their project roles changed
+            throw new Error()
```
fabis94, line 647: ditto

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: @gjedlicska If the new resolver is legitimate, then this is the 2nd time already where we need to duplicate a resolver just to inject some workspaces logic in non-workspaces code. The API surface and the implementations are getting complicated (and largely duplicated), so seems to me that figuring out a solution to the "how to dynamically invoke speckle module APIs and get returned values back" is becoming even more relevant.

packages/frontend-2/components/project/page/settings/collaborators/Collaborators.vue
```diff
@@ -111,10 +129,23 @@ const projectId = computed(() => route.params.id as string)
 const { result: pageResult } = useQuery(projectPageSettingsCollaboratorsQuery, () => ({
   projectId: projectId.value
 }))
+const { result: workspaceResult } = useQuery(
+  projectPageSettingsCollaboratorWorkspaceQuery,
+  () => ({
+    workspaceId: pageResult.value!.project.workspaceId!
+  }),
+  () => ({
+    enabled: isWorkspacesEnabled.value && !!pageResult.value?.project.workspaceId
```
fabis94, line 138: Have you validated what happens if workspaces are disabled? It should not block the role selector from working

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: I also want to add that the solution to that problem also would've made workspace invites logic simpler

packages/frontend-2/components/project/page/settings/collaborators/Collaborators.vue
```diff
@@ -125,11 +156,20 @@ const onCollaboratorRoleChange = async (
   if (collaborator.inviteId) return
 
   loading.value = true
-  await updateRole({
-    projectId: projectId.value,
-    userId: collaborator.id,
-    role: newRole
-  })
+  if (isWorkspacesEnabled.value && project.value?.workspaceId) {
```
fabis94, line 159: I'd abstract this away under a new composable, in case there are multiple places where project roles can be changed, so that future devs don't have to think about which mutation should be invoked

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
cdriesler: Part of the problem is that we don't know if the project is part of a workspace with just the `ProjectMutations.updateRole` mutation. So we can:

- Introduce these extra fetch/checks on every role update

or 

- Introduce this dedicated path

With Monday brain I'm feeling more okay with adding this overhead to all role changes, so that we can guarantee the protections are in place. Esp so if it helps invites work

fabis94: "Introduce these extra fetch/checks on every role update" not possible if we want to avoid workspace code in non-workspaces modules

cdriesler: bubbling up from our convo: investigating using the current event listener/emitter infrastructure for this

gjedlicska: In general, our graphql API should be the smallest possible wrapper on top of existing services. 

Many times, the usecase for a workspace specific action already has some more business logic to it, that is hard to reasonably inject into a non workspace specific resolver. Duplicating resolvers with re-using already working and sufficiently tested resolvers is not a big maintenance burden compared to how much clarity in implementation it adds both on the backend and the frontend.

gjedlicska (approved):

PR #2851, opened 2024-09-02
feat(fe2): fix mp.reset() usage to avoid too many distinct ids + utm tracking not working on homepage

fabis94:


andrewwallacespeckle (approved):
I didn't test the mixpanel stuff, but the middleware changes are working. Previously, login broke without the homepage middleware, but I have confirmed this is now working without this middleware. Nice work

PR #2867, opened 2024-09-03
fix(fe2): Fix scope display in developer settings

andrewwallacespeckle:
Reported by @oguzhankoral in discord:
image

### DIFF
--- a/packages/frontend-2/components/settings/user/Developer.vue
+++ b/packages/frontend-2/components/settings/user/Developer.vue
@@ -377,12 +377,15 @@ const goToExplorer = () => {
 }
 
 const getItemScopes = (item: TokenItem | ApplicationItem): string => {
+  if (!item.scopes || item.scopes.length === 0) return 'No scopes available'
+
   return item.scopes
-    ? item.scopes
-        .map(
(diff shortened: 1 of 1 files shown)


Mikehrn (approved):

packages/frontend-2/components/settings/user/Developer.vue
```diff
@@ -377,12 +377,15 @@ const goToExplorer = () => {
...
   return item.scopes
-    ? item.scopes
-        .map(
-          (event, index, array) => `"${event}"${index < array.length - 1 ? ',' : ''}`
-        )
-        .join(' ')
-    : 'No scopes available'
+    .map((scope) => {
+      if (typeof scope === 'string') return `"${scope}"`
+      if (typeof scope === 'object' && scope !== null) {
```
fabis94, line 385: if its typeof 'object' it already can't be null, can it?

fabis94 (approved):

packages/frontend-2/components/settings/user/Developer.vue
andrewwallacespeckle: Without this it gives a typescript error:
'scope' is possibly 'null'.ts-plugin(18047)

PR #2868, opened 2024-09-03
Fix: Show version count for all members

Mikehrn:


benjaminvo:
As commented on Discord: I would consider just outputting this to everyone, including reviewers. But we should at least output to members in addition to admins since members can also push versions to the workspace.

Mikehrn:
> As commented on Discord: I would consider just outputting this to everyone, including reviewers. But we should at least output to members in addition to admins since members can also push versions to the workspace.

It's possible members have 0 projects with write access, and that guests do have write access to project within a workspace, so I opted to show it for all

gjedlicska (approved):

PR #2878, opened 2024-09-04
fix(ui-components): Update switch to match new design system

andrewwallacespeckle:
Update to new style, as seen here:
https://www.figma.com/design/z9qYUyM3kLH534M19jvCGE/Michal's-library-WIP?node-id=2486-16333&node-type=FRAME&t=6H20EJPyYQ6itPTo-0

### DIFF
--- a/packages/frontend-2/components/project/page/settings/webhooks/Webhooks.vue
+++ b/packages/frontend-2/components/project/page/settings/webhooks/Webhooks.vue
@@ -49,10 +49,8 @@
         <template #enabled="{ item }">
           <FormSwitch
             :model-value="!!item.enabled"
-            icons
             :name="'switch-' + item.id"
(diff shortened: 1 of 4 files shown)


packages/frontend-2/components/project/page/settings/webhooks/Webhooks.vue
andrewwallacespeckle, line 1: Removing the icons prop, which is no longer used.

packages/tailwind-theme/src/plugin.ts
andrewwallacespeckle, line 1: Some components use the foreground-disabled colour, which is not part of the new design system. I've added these as 50% opacity variants, so we can fallback to this if the class is still used somewhere.

Mikehrn (approved):

PR #2879, opened 2024-09-04
feat(fe2): Open settings on click of users or version count

andrewwallacespeckle:
Added click interactions to the version count and user avatar group components. It allows users to quickly access specific sections of the settings dialog.


packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -41,15 +41,19 @@
       </div>
       <div class="flex items-center gap-x-3">
         <div v-if="workspaceInfo.billing" class="flex-1 md:flex-auto">
-          <WorkspacePageVersionCount
-            :versions-count="workspaceInfo.billing.versionsCount"
-          />
+          <button class="block" @click="openSettingsDialog('billing')">
```
Mikehrn, line 44: We should probably create a const for the menu item keys and use that (I also didnt do that for general), maybe something like 

```
const SettingMenuKeys = Object.freeze(<const>{
  Workspace: {
    General: 'workspaceGeneral',
    Billing: 'workspaceBilling'
    // etc
  },
  Server: {
    General: 'serverGeneral'
    // etc
  },
  User: {
    Details: 'userDetails'
    // etc
  },
})
```

And update the composable (`lib/settings/composables/menu.ts`) to use these keys 

```
  const workspaceMenuItems = shallowRef<SettingsMenuItems>({
    [SettingMenuKeys.Workspace.General]: {
      title: 'General',
      component: SettingsWorkspaceGeneral,
      permission: [Roles.Workspace.Admin, Roles.Workspace.Member, Roles.Workspace.Guest]
    },
   // etc
  })
  ```
  
  Then here we can do `@click="openSettingsDialog(SettingMenuKeys.Workspace.General)"` to assure it's an actual existing item (and avoid typos)

Mikehrn (changes requested):

Mikehrn (approved):

PR #2880, opened 2024-09-05
chore(server): apollo server v3 -> v4

fabis94:
there were many breaking changes in this one, the most annoying one being that the context builder function is no longer associated with the ApolloServer instance

also had to update graphql 15 -> 16

i tested out server (& tests), frontend 2 and frontend, tried out both normal operations and subscriptions - all seemed to work


packages/server/app.ts
```diff
@@ -21,17 +21,11 @@ import {
 import { errorLoggingMiddleware } from '@/logging/errorLogging'
 import prometheusClient from 'prom-client'
 
-import {
-  ApolloServer,
-  ForbiddenError,
-  ApolloServerExpressConfig,
-  ApolloError
```
fabis94, line 28: ApolloError no longer exists

packages/server/modules/accessrequests/tests/streamAccessRequests.spec.ts
```diff
@@ -55,7 +55,7 @@ const cleanup = async () => {
 }
 
 describe('Stream access requests', () => {
-  let apollo: ApolloServer
+  let apollo: ServerAndContext
```
fabis94, line 58: Biggest annoyance - Apollo Server instance is no longer associated w/ the context builder, its now fed into the express middleware instead. When running operations against it manually, the context value has to be passed into the execute() call

packages/server/modules/blobstorage/graph/resolvers/index.ts
```diff
@@ -13,9 +13,12 @@ import {
...
 } from '@/modules/core/graph/generated/graphql'
 import { StreamGraphQLReturn } from '@/modules/core/helpers/graphTypes'
-import { NotFoundError, ResourceMismatch } from '@/modules/shared/errors'
+import {
+  BadRequestError,
+  NotFoundError,
+  ResourceMismatch
+} from '@/modules/shared/errors'
 import { Nullable } from '@speckle/shared'
-import { UserInputError } from 'apollo-server-errors'
```
fabis94, line 18: UserInputError no longer exists

PR #2890, opened 2024-09-05
fix(fe2): Update cache after removing workspace members

andrewwallacespeckle:
Use the new modifyObjectField to update the cache, when a user has had their role set to null (removed).

### DIFF
--- a/packages/frontend-2/lib/workspaces/composables/management.ts
+++ b/packages/frontend-2/lib/workspaces/composables/management.ts
@@ -414,7 +414,22 @@ export const useWorkspaceUpdateRole = () => {
   const { triggerNotification } = useGlobalToast()
 
   return async (input: WorkspaceRoleUpdateInput) => {
-    const result = await mutate({ input }).catch(convertThrowIntoFetchResult)
+    const result = await mutate(
+      { input },
+      {
+        update: (cache) => {
+          if (!input.role) {
+            // If role is null, we're removing the user
+            modifyObjectField(
+              cache,
+              getCacheId('Workspace', input.workspaceId),
+              'team',
+              ({ helpers: { evict } }) => evict()
+            )
+          }
+        }
+      }
+    ).catch(convertThrowIntoFetchResult)
 
     if (result?.data) {
       triggerNotification({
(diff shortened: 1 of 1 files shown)


packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -414,7 +414,25 @@ export const useWorkspaceUpdateRole = () => {
...
+      { input },
+      {
+        update: (cache) => {
+          if (!input.role) {
+            modifyObjectField(
+              cache,
+              getCacheId('Workspace', input.workspaceId),
+              'team',
+              () => {
+                cache.evict({
```
fabis94, line 427: You're mixing up 2 things here. cache.evict() works by itself, there's no need to wrap it within modifyObjectField - it will just evict the object w/ that ID from the cache.

You'd only need modifyObjectField(workspace.team), if you want to edit that specific field (a workspace's team array). In this case I don't think you need it, cause if you cache.evict() the collaborator beforehand, the reference is gonna dissapear from any arrays its in automatically.

fabis94 (changes requested):

packages/frontend-2/lib/workspaces/composables/management.ts
fabis94: @andrewwallacespeckle Nvm, team isn't an array, its a collection. So yea, while it will disappear automatically from the collection's items array, you still need to update totalCount.

fabis94: Also - the field supports filters, so use the `autoEvictFiltered` option for modifyobjectfield

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -414,7 +414,35 @@ export const useWorkspaceUpdateRole = () => {
...
+
+            modifyObjectField(
+              cache,
+              getCacheId('Workspace', input.workspaceId),
+              'team',
+              ({ value }) => {
+                if (value.totalCount) {
+                  return {
+                    ...value,
+                    totalCount: value.totalCount - 1
```
fabis94, line 434:
```
            ({ helpers: { createUpdatedValue } }) => {
              return createUpdatedValue(({ update }) => {
                update('totalCount', (totalCount) => totalCount - 1)
              })
            },
```

fabis94 (changes requested):

fabis94 (approved):

PR #2894, opened 2024-09-05
fix(workspaces): fix leaky search

cdriesler:
## Changes:

- Fix `orWhere()` clause to limit workspace team searches to intended limit


gjedlicska (changes requested):
To be fully sure in tests, lest make assertion on the user IDs.

There is a way to do assert equals in any order

gjedlicska (approved):

PR #2904, opened 2024-09-09
Feat: Mixpanel workspace groups add properties

Mikehrn:
We're adding several properties to Mixpanel workspace groups, such as billing details, version count, and member data. Since many of these properties aren't directly tied to an action in the UI, I've opted to gather and send this information when the workspace dashboard is opened. I considered using the settings modal for this, but it has lower traffic and I think it wouldn't be updated often enough.


packages/frontend-2/pages/workspaces/[id]/index.vue
```diff
@@ -4,12 +4,37 @@
...
 const workspaceId = computed(() => route.params.id as string)
+const { workspaceMixpanelUpdateGroup } = useWorkspacesMixpanel()
+const pageFetchPolicy = usePageQueryStandardFetchPolicy()
 
 definePageMeta({
   middleware: ['requires-workspaces-enabled', 'require-valid-workspace']
 })
+
+const { result } = useQuery(
+  workspaceMixpanelUpdateGroupQuery,
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[id]/index.vue:22]: why not merge it with the main workspace page query that already exists in WorkspaceProjectList? this will unnecessarily lengthen page load times, especially since its quite a hefty query

packages/frontend-2/pages/workspaces/[id]/index.vue
```diff
@@ -4,12 +4,37 @@
...
+  () => ({
+    workspaceId: workspaceId.value
+  }),
+  () => ({
+    fetchPolicy: pageFetchPolicy.value
+  })
+)
+
+watch(
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[id]/index.vue:31]: Instead of (deep) watching, just run the updater once onResult. If there are any changes to the workspace (e.g. versionCount), you're not watching for them anyway (you only run updater if workspaceId changes) and you will get those changes the next time the page is reloaded.

packages/frontend-2/lib/workspaces/composables/mixpanel.ts
```diff
@@ -0,0 +1,87 @@
...
+`)
+
+graphql(`
+  fragment WorkspaceMixpanelUpdateGroup_Workspace on Workspace {
+    id
+    name
+    description
+    domainBasedMembershipProtectionEnabled
+    discoverabilityEnabled
+    id
```
fabis94, line 23: id twice

fabis94 (changes requested):

packages/frontend-2/pages/workspaces/[id]/index.vue
Comment (Mikehrn) [reply packages/frontend-2/pages/workspaces/[id]/index.vue:22]: I separated them so they get too intertwined, if subsequent queries fetch the same data they will leverage the cache or not?

Comment (fabis94) [reply packages/frontend-2/pages/workspaces/[id]/index.vue:22]: These both run at the same time, so there's nothing in the cache and they end up executing the same thing twice. Leveraging the cache would only be relevant if the data were already in the cache when the 2nd query starts, but here they both run concurrently.

Generally, for any kind of page load queries (not some query triggered async on button press or whatever) you just want everything to be loaded with as few queries as possible, because the cache is gonna be empty at that time (in SSR).

Comment (Mikehrn) [reply packages/frontend-2/pages/workspaces/[id]/index.vue]: Good point, I've updated it to be part of the same query

packages/frontend-2/lib/workspaces/graphql/queries.ts
```diff
@@ -54,3 +55,11 @@ export const workspaceInviteQuery = graphql(`
     }
   }
 `)
+
+export const workspaceMixpanelUpdateGroupQuery = graphql(`
```
fabis94, line 59: no longer needed?

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -239,4 +241,10 @@ const onShowSettingsDialog = (target: AvailableSettingsMenuKeys) => {
   showSettingsDialog.value = true
   settingsDialogTarget.value = target
 }
+
+onResult(() => {
```
fabis94, line 245: IIRC you can get the result from the arguments of this callback, no need to use initialQueryResult

fabis94 (changes requested):
found unused code

packages/frontend-2/lib/workspaces/graphql/queries.ts
Mikehrn: Unsaved file 😅

fabis94 (approved):

PR #2905, opened 2024-09-09
fix(fe2): Encode forward slashes in nested model URLs

andrewwallacespeckle:


packages/frontend-2/components/project/page/models/StructureItem.vue
```diff
@@ -380,7 +380,9 @@ const modelLink = computed(() => {
 
 const viewAllUrl = computed(() => {
   if (isPendingFileUpload(props.item)) return undefined
-  return modelRoute(props.project.id, `$${props.item.fullName}`)
+  const fullName = props.item.fullName
+  const encodedFullName = `$${fullName}`.replace(/\//g, '%2F')
```
andrewwallacespeckle, line 384: @fabis94 I used a regex here to handle multiple levels of nested folders

fabis94 (approved):

PR #2913, opened 2024-09-10
chore(server): emails IoC #1 - getPendingTokenFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/emails/repositories/index.ts
+++ b/packages/server/modules/emails/repositories/index.ts
@@ -1,7 +1,9 @@
 import { EmailVerifications } from '@/modules/core/dbSchema'
+import { GetPendingToken } from '@/modules/emails/domain/operations'
 import { InvalidArgumentError } from '@/modules/shared/errors'
 import cryptoRandomString from 'crypto-random-string'
 import dayjs from 'dayjs'
+import { Knex } from 'knex'
 
 export type EmailVerificationRecord = {
   id: string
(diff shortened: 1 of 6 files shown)


packages/server/modules/emails/repositories/index.ts
```diff
@@ -12,24 +14,25 @@ export type EmailVerificationRecord = {
...
+  async ({ token, email }) => {
+    if (!token && !email) throw new InvalidArgumentError('Token & email is empty')
 
-  const aWeekAgo = dayjs().subtract(1, 'week')
+    const aWeekAgo = dayjs().subtract(1, 'week')
 
-  const q = EmailVerifications.knex<EmailVerificationRecord>()
-    .where(EmailVerifications.col.createdAt, '>', aWeekAgo.toISOString())
-    .first()
+    const q = EmailVerifications.knex<EmailVerificationRecord>(deps.db)
```
alemagio, line 24: here we should use the injected db
I know that technically behind the scene it is using it
But the idea is to completely remove things like `EmailVerification.knex` from the codebase
We should always use `db(tableName)`

alemagio (changes requested):

packages/server/modules/emails/repositories/index.ts
fabis94: @alemagio It is using the injected DB, so why remove the useful helper?

PR #2914, opened 2024-09-10
chore(server): emails IoC 2 - deleteVerificationsFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/dbSchema.ts
+++ b/packages/server/modules/core/dbSchema.ts
@@ -137,7 +137,7 @@ const createBaseInnerSchemaConfigBuilder =
 
     return {
       name: aliasedTableName as T,
-      knex: (db?: Knex) => (knex || db)(aliasedTableName),
+      knex: (db?: Knex) => (db || knex)(aliasedTableName),
       col: reduce(
         columns,
         (prev, curr) => {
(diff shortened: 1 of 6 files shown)


packages/server/modules/emails/services/verification/request.ts
```diff
@@ -43,7 +43,9 @@ async function createNewVerification(
   if (user.verified)
     throw new EmailVerificationRequestError("User's email is already verified")
 
-  const verificationId = await deleteOldAndInsertNewVerification(user.email)
+  const verificationId = await deleteOldAndInsertNewVerificationFactory({ db })(
```
alemagio, line 46: I think this function should be injected, it should be instantiated in resolvers right?

alemagio (changes requested):

packages/server/modules/emails/services/verification/request.ts
alemagio: @fabis94 I think there was a misunderstanding.
I meant changing the function everywhere it is used so we don't need to change it again in the same work.
So, it is changing through all its usage paths.
But if you already made the work like that I'm fine merging it like this

alemagio (approved):

PR #2917, opened 2024-09-10
Feat: add filter for workspace members table

Mikehrn:
Add filter to workspace members table to filter on role

Screenshot 2024-09-10 at 10 00 18


andrewwallacespeckle (changes requested):
The code looks good, but there is a slight alignment bug:
image

The select seems to be 34px, but the search is only 32px.

andrewwallacespeckle:
Code looks

> The code looks good, but there is a slight alignment bug: image
> 
> The select seems to be 34px, but the search is only 32px.

Also a slight difference in the non-active or hovered border colour

andrewwallacespeckle (approved):

PR #2919, opened 2024-09-10
chore(server): emails IoC 6 - sendVerificationEmailFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/graph/resolvers/userEmails.ts
+++ b/packages/server/modules/core/graph/resolvers/userEmails.ts
@@ -18,12 +18,16 @@ import { validateAndCreateUserEmailFactory } from '@/modules/core/services/userE
 import { getUser } from '@/modules/core/repositories/users'
 import { getServerInfo } from '@/modules/core/services/generic'
 import { deleteOldAndInsertNewVerificationFactory } from '@/modules/emails/repositories'
+import { renderEmail } from '@/modules/emails/services/emailRendering'
+import { sendEmail } from '@/modules/emails/services/sending'
 
 const requestNewEmailVerification = requestNewEmailVerificationFactory({
   findEmail: findEmailFactory({ db }),
   getUser,
   getServerInfo,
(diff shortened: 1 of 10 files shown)


packages/server/modules/core/graph/resolvers/userEmails.ts
```diff
@@ -18,12 +18,16 @@ import { validateAndCreateUserEmailFactory } from '@/modules/core/services/userE
...
+import { sendEmail } from '@/modules/emails/services/sending'
 
 const requestNewEmailVerification = requestNewEmailVerificationFactory({
   findEmail: findEmailFactory({ db }),
   getUser,
   getServerInfo,
-  deleteOldAndInsertNewVerification: deleteOldAndInsertNewVerificationFactory({ db })
+  deleteOldAndInsertNewVerification: deleteOldAndInsertNewVerificationFactory({ db }),
+  renderEmail,
+  sendEmail
```
alemagio, line 30: If I understand correctly, here we are passing the deps that are needed for the function we are currently refactoring.
I think this might lead to changing the signature of this function multiple times.
WDYT?

fabis94: @alemagio Ultimately what matters in my approach is that sendVerificationEmailFactory is refactored and converted to be dependency injectable, that's done and I don't have to do that anymore. Having to come back and potentially update it is not the issue, if I'm forced to do it and I can't forget about it - and I can't forget about it if TypeScript is showing red errors that the function signature is no longer valid.

It's not having to come back and make updates that's the issue, its having to do so in a way where you can easily forget to do it/aren't forced to do it by the compiler.

fabis94: There shouldn't be any issues with this approach, if all of the PRs are merged you will see that all of the (relevant) functions are updated and converted as expected in the entire module.

alemagio (approved):

PR #2955, opened 2024-09-11
refactor(fe2): If description is line-clamped show read more and description dialog

andrewwallacespeckle:
image
image

I am having a strange reactivity problem on the read more, which appears on hover of a line-clamped description. On page load, it works as expected. If I navigate to a page with no description, or one too short to line clamp, it correctly removes the read more button. The problem comes when I navigate back to the original workspace, the read more is not visible on hover, and requires a refresh to work again.


andrewwallacespeckle:
From @Mikehrn:

I think the issue is that the dom is not finished rendering yet
If I do this 
const checkOverflow = () => {
  setTimeout(() => {
    if (descriptionRef.value) {
      console.log(descriptionRef.value.scrollHeight)
      console.log(descriptionRef.value.clientHeight)
      hasOverflow.value =
        descriptionRef.value.scrollHeight > descriptionRef.value.clientHeight
    }
  }, 2000)
}
It works
(timeout is not the solution, but just try to see if waiting for anything fixes it haha)

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -173,4 +193,28 @@ const onActionChosen = (params: { item: LayoutMenuItem; event: MouseEvent }) =>
...
+const isDescriptionDialogOpen = ref(false)
+
+const showDescriptionDialog = () => {
+  isDescriptionDialogOpen.value = true
+}
+
+watch(
+  () => props.workspaceInfo,
+  () => {
+    nextTick(checkOverflow)
```
fabis94, line 216: This is still imperative and still unreliable - whether or not the DOM node will be ready on next tick could change, and there's also the fact that window resizes should retrigger size calculations as well.

You should be using some kind of reactive vueuse DOM size tracking util instead, so that hasOverflow can be a declarative computed and you don't even have to worry about when and how to trigger recalculation

fabis94 (changes requested):

packages/frontend-2/components/workspace/header/Header.vue
andrewwallacespeckle: Thanks. There is no vueuse util that can read scroll height or client height. At least one that I could find. Any suggestions?

fabis94: @andrewwallacespeckle What about useElementSize or useElementBounding? Even if it doesn't give you those size values you're after, as long as it updates when the element size updates, you can at the very least retrigger overflow calculation when that happens (e.g. with a watcher)

andrewwallacespeckle: Thanks for your help on this! `useElementSize` worked perfectly

fabis94 (approved):

PR #2961, opened 2024-09-11
Fix: Associate user with mixpanel group on creation and dashboard open

Mikehrn:
This will ensure every user is added to the mixpanel group


Mikehrn:
Making this a draft until we decide to keep groups or change it in favor or user properties

Mikehrn:
Closing this, in favor of adding them as user properties instead

PR #2964, opened 2024-09-11
Feat: Allow users to create or select workspace on project creation

Mikehrn:
Without workspaces:
Screenshot 2024-09-11 at 17 11 00

With workspaces:
Screenshot 2024-09-11 at 17 10 36

Create new workspace:
Screenshot 2024-09-11 at 17 11 09


packages/frontend-2/components/projects/AddDialogNewWorkspace.vue
andrewwallacespeckle, line 1: All looks good, my only comment is on the naming of this file. Normally we would call it NewWorkspaceDialog, with the component type at the end. But maybe I'm missing some context on this.

andrewwallacespeckle (approved):
Just the one comment on naming, but if you have a reason for it, feel free to merge.

packages/frontend-2/components/projects/AddDialogNewWorkspace.vue
Mikehrn: @andrewwallacespeckle It's not a dialog on it's own, it's a subcomponent of ProjectsAddDialog, which if why I went with that prefix :)

andrewwallacespeckle (approved):

PR #2977, opened 2024-09-12
Fix: dont show workspace select on projects inside workspaces

Mikehrn:


cdriesler (approved):
✅ (greener than gergo's)

PR #2982, opened 2024-09-12
feat(workspaces): add workspace slug support

gjedlicska:
## Checklist:


packages/server/modules/workspaces/domain/operations.ts
```diff
@@ -32,11 +32,21 @@ export type GetWorkspace = (args: {
...
+  workspaceSlug: string
+  userId?: string
+}) => Promise<WorkspaceWithOptionalRole | null>
+
 export type GetWorkspaces = (args: {
   workspaceIds: string[]
   userId?: string
 }) => Promise<WorkspaceWithOptionalRole[]>
 
+export type GetWorkspacesBySlug = (args: {
```
cdriesler, line 45: Appears to be unused - if forward looking, `workspaceIds` => `slugs`?

cdriesler (approved):
👍👍

If I'm reading this correctly, it looks like the backend cannot provide a valid replacement slug if the default or user-provided slugs are invalid? It can only silently replace it on workspace create. (And say valid/invalid)

Maybe we plan to force a change on the frontend if the default is invalid? Or let this happen invisibly, for cases where the user didn't change the slug at all. We could share the slug generation/validation logic between the FE & BE, but _ever so strictly_ the FE can't validate the `${slug}-${crs}` hasn't already been taken.

gjedlicska:
Right, I meant to move the name generation logic to shared

PR #2984, opened 2024-09-13
Alessandro/web 1767 guest table should show what they have access to

alemagio:
## Checklist:


fabis94 (changes requested):

packages/server/modules/workspaces/repositories/workspaces.ts
```diff
@@ -408,3 +412,34 @@ export const countWorkspaceRoleWithOptionalProjectRoleFactory =
...
+
+export const getProjectRolesFactory =
+  ({ db }: { db: Knex }): GetProjectRoles =>
+  async ({ userIds, workspaceId }) => {
+    return await tables
+      .streamAcl(db)
+      .select<{ project: StreamRecord } & Pick<StreamAclRecord, 'role'>>(
+        StreamAcl.col.userId,
+        knex.raw(
+          `array_agg(json_build_object(
```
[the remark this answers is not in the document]
fabis94: Don't get me wrong, its nice that you used dataloaders already, its just that we don't need the actual retrieval of streams to be a part of that and we can simplify the whole thing

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -761,6 +761,12 @@ export = FF_WORKSPACES_MODULE_ENABLED
         },
         role: async (parent) => {
           return parent.workspaceRole
+        },
+        projectRoles: async (parent, _args, ctx) => {
+          return await ctx.loaders.workspaces!.getProjectRolesByWorkspaceId.load({
```
fabis94, line 766: Not sure if there's a point to this dataloader, what would be the scenario in which these calls could be batched together? Each entry is always gonna be its own workspace & user id pair and thus a separate query 🤔

packages/server/modules/workspaces/graph/dataloaders/workspaces.ts
```diff
@@ -35,7 +38,18 @@ const dataLoadersDefinition = defineRequestDataloaders(({ ctx, createLoader }) =
           )
           return ids.map((id) => results[id] || null)
         }
-      )
+      ),
+      getProjectRolesByWorkspaceId: createLoader<
+        { workspaceId: string; userId: string },
+        (Pick<StreamAclRecord, 'role'> & { project: StreamRecord })[] | null
+      >(async (keys) => {
+        const workspaceId = keys[0].workspaceId
```
fabis94, line 46: If we're ignoring all the other workspaceIds, then the key is misleading. The loader key makes one think that arbitrary pairs of workspaceId & user Id are all going to be batched as one query. In this case, if you're only supporting 1 workspaceId, but multiple user ids, then you might want to do something similar to `getStreamCommit` - create a separate loader for each workspaceId, so like for getStreamCommit we have `forStream(id)` you might want a `forWorkspace(id)` here

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: I just read the implementation and understand this better. If workspaceId is the same and multiple users can be batched together, then there is in fact a point to this dataloader. But yea, the API should be less misleading - the workspaceId cannot be different between dataloader load() requests

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -761,6 +762,18 @@ export = FF_WORKSPACES_MODULE_ENABLED
         },
         role: async (parent) => {
           return parent.workspaceRole
+        },
+        projectRoles: async (parent, _args, ctx) => {
+          const projectRoles = await getRolesByUserIdFactory({ db })({
+            userId: parent.id,
+            workspaceId: parent.workspaceId
+          })
+          return await Promise.all(
```
fabis94, line 771: The actual getStream.load() call should preferably be made in the ProjectRole.project resolver - otherwise we're prefetching all of these streams, even tho the requester might only be interested in roles

fabis94 (approved):

packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -333,13 +333,19 @@ type WorkspaceCollaboratorCollection {
   items: [WorkspaceCollaborator!]!
 }
 
+type ProjectRole {
```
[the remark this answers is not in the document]
alemagio: I don't see this very coupled with the workspace concept

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: This Promise.all() is no longer needed here

fabis94 (dismissed):

fabis94 (approved):

PR #2987, opened 2024-09-13
chore(server): notifications IoC 3 - digestNotificationEmailHandler

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/notifications/services/handlers/activityDigest.ts
+++ b/packages/server/modules/notifications/services/handlers/activityDigest.ts
@@ -28,43 +28,41 @@ import {
 import { getUserNotificationPreferencesFactory } from '@/modules/notifications/services/notificationPreferences'
 import { getSavedUserNotificationPreferencesFactory } from '@/modules/notifications/repositories'
 import { db } from '@/db/knex'
+import { GetUserNotificationPreferences } from '@/modules/notifications/domain/operations'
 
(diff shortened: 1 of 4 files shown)


packages/server/modules/notifications/services/handlers/activityDigest.ts
```diff
@@ -424,3 +425,25 @@ Here's a summary of what happened in the past week
...
+    getSavedUserNotificationPreferences: getSavedUserNotificationPreferencesFactory({
+      db
+    })
+  }),
+  createActivitySummary,
+  getServerInfo,
+  renderEmail
+})
+
+const handler: NotificationHandler<ActivityDigestMessage> = async (msg) => {
```
fabis94, line 440: @alemagio These notification handlers run on messages retrieved from a BullMQ messaging queue. So none of this logic happens inside a user's server request, but in the BG of the server process, and thus its probably not gonna conditionally switch the DB based on a user's request (since there is no user request). That said - I've still made the main handler dependency injectable, in case we do want the deps (db?) to be different between message queue messages

alemagio: I think you made it right making it injectable.
I don't think this will be problematic since I don't think we'll need to ensure data locality for notifications

alemagio (approved):

PR #2999, opened 2024-09-13
Fix: Change add_group to people.set for workspaces

Mikehrn:
This remove add_group (associating with workspace groups) and instead changes it to user properties, to declutter and give some more flexibility


packages/frontend-2/components/dashboard/Sidebar.vue
```diff
@@ -233,4 +237,13 @@ const openWorkspaceCreateDialog = () => {
     source: 'sidebar'
   })
 }
+
+onWorkspaceResult((result) => {
```
Mikehrn, line 241: When users open the app we fetch all workspaces to add them to the sidebar, this hooks into that

andrewwallacespeckle (approved):

teocomi (approved):

PR #3010, opened 2024-09-16
Alessandro/web 957 dispatch webhook event

alemagio:
## Checklist:


packages/server/modules/webhooks/services/webhooks.ts
```diff
@@ -0,0 +1,183 @@
...
+} from '@/modules/core/repositories/streams'
+import {
+  getUser as getUserFn,
+  UserWithOptionalRole
+} from '@/modules/core/repositories/users'
+import { Knex } from 'knex'
+
+const MAX_STREAM_WEBHOOKS = 100
+
+export const createWebhook =
```
fabis94, line 26: Its confusing that there's 2 operations, both called "create webhook". We sometimes have the same thing where there's the DB level creation operation and a service level operation that also adds extra validation on top (in this case this function probably makes more sense as a service level call). It just makes it more likely that someone coming in later will pick the wrong one and in the worst case scenario - avoid all validation.

I tend to solve this problem by forcing different naming. Maybe leaving the DB operation as "createX" and then having the service call be called "createValidatedX" or something like that. Ultimately you can name it all kinds of different things, but it should ideally have 1) its own operation type and 2) a different name.

packages/server/modules/webhooks/services/webhooks.ts
```diff
@@ -0,0 +1,183 @@
...
+  ({
+    createWebhookConfig,
+    countWebhooksByStreamId
```
fabis94, line 29: Kinda subjective so you don't have to change this, but I tend to avoid destructuring the deps here and leave it in a `deps` object, because:

1) Less busy function signature. Repeating the same dependency twice - first as an actual destructuring operation and then to specify it's time - looks busy/mesy IMO.
2) If you don't destructure the deps and invoke the dep from that object (e.g. `deps.invokeX()`) you can always be sure that you're in fact invoking the injected dependency, not a function with the same name from the outer scope, in case you forgot to destructure it. This is relevant when we have those `typeof X` injections for old operations that aren't yet DI supported. If you do `getUser: typeof getUser` and forget to destructure it from `deps`, you will actually use the one imported in the module not the injected version. I've already accidentally made this mistake in the past.

fabis94 (changes requested):

packages/server/modules/webhooks/services/webhooks.ts
alemagio: I think, at some point, we should agree on one way or the other.
But for now, I'll leave it there since, as you said, it is subjective.

packages/server/modules/webhooks/services/webhooks.ts
alemagio: I changed the repository functions names since, I believe, in the repository context the name should be related to the table name which is `WebhookConfig`

fabis94 (approved):

PR #3034, opened 2024-09-18
Feat: Allow selection of default workspace roles

Mikehrn:
Screenshot 2024-09-18 at 13 22 37


andrewwallacespeckle (approved):

cdriesler (commented):
Behaves well, looks sound! But some notes (maybe more just a sign that roles are getting complex):

- In the role description, I think it's also important to capture that this is the default role members will get in all projects when added to the workspace. Maybe something like:
  - "Role workspace members get when added to the workspace and in newly created projects." (v negotiable)
- I don't think we expose "collaborator" or "reviewer" anywhere else in the UI? On invites and project settings, these are expressed as "can edit" and "can view". Does it make sense to express the setting in the UI as something like "default access" or "default permissions", even if the implementation is strictly roles? If we do this, the previous comment should adjust a little too.

andrewwallacespeckle (approved):

PR #3094, opened 2024-09-24
Feat: Enable billing page for members

Mikehrn:
Screenshot 2024-09-24 at 12 54 42


packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
```diff
@@ -67,6 +67,7 @@
             @click="toggleMenu(item.id)"
           />
         </LayoutMenu>
+        <span v-else />
```
Mikehrn, line 70: Small fix to avoid "placeholder" showing from the UI component

andrewwallacespeckle (approved):

PR #3107, opened 2024-09-25
Feat(dui3): CNX-548 card settings backward and forward compatible

oguzhankoral:
We upgrade the sender card settings on the init.

Another issue was, we were not awaiting the init functions like `void refreshDocumentModelStore()` in the `hostAppStore` and it was causing empty `documentModelStore` whenever we call `tryToUpgradeSettings()`. I have wrapped them into `async` init function and await them all.

### DIFF
--- a/packages/dui3/store/hostApp.ts
+++ b/packages/dui3/store/hostApp.ts
@@ -523,7 +523,6 @@ export const useHostAppStore = defineStore('hostAppStore', () => {
     await refreshDocumentModelStore()
     await refreshSendFilters()
     await getSendSettings()
-
     await getHostAppName()
     await getHostAppVersion()
     await getConnectorVersion()
--- a/packages/dui3/store/hostApp.ts
+++ b/packages/dui3/store/hostApp.ts
@@ -448,10 +448,11 @@ export const useHostAppStore = defineStore('hostAppStore', () => {
 
   const getSendSettings = async () => {
     sendSettings.value = await app.$sendBinding.getSendSettings()
-    tryToUpgradeSendSettings()
+    // TODO: tryToUpgradeSendSettings()
   }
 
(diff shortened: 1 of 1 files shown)


packages/dui3/store/hostApp.ts
```diff
@@ -446,28 +446,89 @@ export const useHostAppStore = defineStore('hostAppStore', () => {
   const refreshSendFilters = async () =>
     (sendFilters.value = await app.$sendBinding?.getSendFilters())
 
-  const getSendSettings = async () =>
-    (sendSettings.value = await app.$sendBinding.getSendSettings())
+  const getSendSettings = async () => {
+    sendSettings.value = await app.$sendBinding.getSendSettings()
+    tryToUpgradeSettings('SenderModelCard')
```
didimitrie, line 451: This is a bit of a smell: 
- either rename to `getSendSettingsAndUpgradeCards()`, or
- leave as is, and call `tryToUpgradeSettings` at the end in `initializeApp` (this would be my preference)

didimitrie (changes requested):
minor change requested. another question i have is on enum/string values being different: what if the currently selected one is no longer present/valid in some way? should we reset it to the default?

packages/dui3/store/hostApp.ts
oguzhankoral: that sounds make sense, will do second

oguzhankoral:
> minor change requested. another question i have is on enum/string values being different: what if the currently selected one is no longer present/valid in some way? should we reset it to the default?

@didimitrie if we detect **_any_** chance on a setting, it resets it completely, so the value will be the current default even if the selected value exists in the new items. we could achieve it actually with more checks, WDYT?

PR #3115, opened 2024-09-25
chore(server): shared IoC 1 - index.js

fabis94:
## Checklist:


packages/server/modules/shared/index.ts
```diff
@@ -0,0 +1,34 @@
...
+} from '@/modules/shared/utils/subscriptions'
+
+export {
+  pubsub,
+  StreamSubscriptions as StreamPubsubEvents,
+  CommitSubscriptions as CommitPubsubEvents,
+  BranchSubscriptions as BranchPubsubEvents
+}
+
+export const validateScopes = validateScopesFactory()
```
fabis94, line 27: i converted these to factories, but left them as globally instantiated singletons cause that's what they are

packages/server/modules/core/tests/generic.spec.js
```diff
@@ -157,25 +157,23 @@ describe('Generic AuthN & AuthZ controller tests', () => {
       envHelperMock.resetMockedFunctions()
     })
     it('should allow stream:owners to be stream:owners', async () => {
-      const role = await authorizeResolver(
+      await authorizeResolver(
```
fabis94, line 160: Nothing outside of tests uses the return, and the return was very dynamic before - it could return server roles, stream roles, even fake roles if admin override was enabled. I decided to just scrap the otherwise `any` return, since nothing really needs it.

alemagio (approved):

PR #3124, opened 2024-09-25
Feat: Added modal for moving projects into workspace

Mikehrn:
Adds modal for moving projects into a workspace, until we have a proper bulk move option. Added some complexity to the `MoveProjectsDialog.vue` but I will follow up with a change on where we select the workspace always outside of the modal, which we will remove after related to workspace selection/creation

Screenshot 2024-09-25 at 16 24 20

Screenshot 2024-09-25 at 16 24 34


packages/frontend-2/components/projects/MoveToWorkspaceDialog.vue
```diff
@@ -2,41 +2,45 @@
...
-        <ProjectsWorkspaceSelect
-          v-if="hasWorkspaces"
-          v-model="selectedWorkspace"
-          :items="workspaces"
-          :disabled-roles="[Roles.Workspace.Member, Roles.Workspace.Guest]"
-          disabled-item-tooltip="Only workspace admins can move projects into a workspace."
-          label="Select a workspace"
-          help="Once a project is moved to a workspace, it cannot be moved out from it."
-          show-label
+      <template v-if="!workspace">
```
Mikehrn, line 5: This whole part will be remove from this modal soon, when workspace selection will happen outside of the modal and we will always pass a workspaceId as prop

andrewwallacespeckle:
Not sure what's happening, but when I go to "Move project to workspace", then hit Cancel in the dialog. It won't launch any of the other dialogs again. 
image

andrewwallacespeckle:
The new button seems to cause some overflow issues on mobile:
image

Maybe move this to the ... menu on mobile? It probably makes sense to always have it there, but maybe this was a design decision to have it as a button?

packages/frontend-2/components/workspace/header/Header.vue
andrewwallacespeckle, line 1: You're missing the case in onActionChosen for ActionTypes.MoveProjects

andrewwallacespeckle (changes requested):

andrewwallacespeckle (dismissed):

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -154,27 +164,41 @@ graphql(`
...
   (e: 'show-settings-dialog', v: AvailableSettingsMenuKeys): void
+  (e: 'show-move-projects-dialog'): void
 }>()
 
 const props = defineProps<{
   workspaceInfo: WorkspaceHeader_WorkspaceFragment
 }>()
 
 const menuId = useId()
+const breakpoints = useBreakpoints(TailwindBreakpoints)
```
fabis94, line 183: JS breakpoints are never gonna be calculated correctly in SSR (since there's no screen), so preferably we want to rely on CSS instead. In this case, if you don't see any problems (e.g. SSR rendered page looking weird in mobile or desktop sizes) its probably fine.

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -551,11 +552,30 @@ export function useMoveProjectToWorkspace() {
...
+    workspaceName: string,
+    eventSource?: string
+  ) => {
     const { data, errors } = await apollo
       .mutate({
         mutation: useMoveProjectToWorkspaceMutation,
-        variables: { projectId, workspaceId }
+        variables: { projectId, workspaceId },
+        update(cache) {
+          if (!workspaceId) return
```
fabis94, line 566: You should check if the update actually succeeded. Check the mutation response result (passed in as a param to update()).

packages/frontend-2/lib/projects/composables/projectManagement.ts
```diff
@@ -551,11 +552,30 @@ export function useMoveProjectToWorkspace() {
   const { triggerNotification } = useGlobalToast()
   const mixpanel = useMixpanel()
 
-  return async (projectId: string, workspaceId: string, workspaceName: string) => {
+  return async (
+    projectId: string,
```
fabis94, line 556: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#c79ca97569504d1dbd9d5d3b9978360a

packages/frontend-2/components/workspace/header/Header.vue
Mikehrn: I cannot really rely on CSS, because it's used in a computed to conditionally add different actions to a context menu, tested it on mobile/desktop and different ways of loading but seems to be working as expected :)

fabis94 (approved):

PR #3179, opened 2024-10-02
Feat: Add custom modal for user feedback

Mikehrn:
Screenshot 2024-10-02 at 12 28 22


packages/frontend-2/components/feedback/Dialog.vue
```diff
@@ -0,0 +1,80 @@
...
+  triggerNotification({
+    type: ToastNotificationType.Success,
+    title: 'Thank you for your feedback!'
+  })
+
+  mixpanel.track('Feedback Sent', {
+    feedback: feedback.value
+  })
+
+  await sendWebhook('https://hooks.zapier.com/hooks/catch/12120532/2m4okri/', {
```
fabis94, line 69: Is this OK being hardcoded? All of our deployments are going to use this

packages/frontend-2/components/feedback/Dialog.vue
```diff
@@ -0,0 +1,80 @@
...
+const { handleSubmit } = useForm<FormValues>()
+
+const feedback = ref('')
+
+const dialogButtons = computed((): LayoutDialogButton[] => [
+  {
+    text: 'Send',
+    props: { color: 'primary' },
+    onClick: () => {
+      onSubmit()
```
fabis94, line 50: Doesn't make sense that there's `useForm` and yet you're not actually using a `<form/>`. Use the on-submit prop of LayoutDialog to properly use form submission for this and convert the button to a submit button.

fabis94 (changes requested):

packages/frontend-2/components/feedback/Dialog.vue
Mikehrn: Good point, updated 👍

packages/frontend-2/components/feedback/Dialog.vue
Mikehrn: In this case it's a feedback form regarding Speckle as a product, I think in this case maybe no harm as we do want to gather feedback no matter what deployment this is on?

fabis94 (approved):

PR #3185, opened 2024-10-07
Feat: Remove workspace promobanner

Mikehrn:
This removes the temporary workspace promo banner. @benjaminvo also assigned this to you as you raised you would like to remove this


benjaminvo (changes requested):
We agreed to also include adding a new banner to the bottom of the sidebar

andrewwallacespeckle:
@Mikehrn I've left the old banner code for the SpeckleCon Banner, but set the expiry date to yesterday so it won't show. 

I don't want to remove this entirely, so that if we use this in future again, we can just update that content. 

image

andrewwallacespeckle:
@benjaminvo I've added the SpeckleCon promo block in this branch. Feel free to go in and update the copy, or post it here and I can do it. 

Also open to any feedback you have on design, but again you're welcome to do this yourself if you want. 

image

benjaminvo (approved):

PR #3186, opened 2024-10-07
Feat: Add SSO settings

Mikehrn:


cdriesler (approved):
does config correctly if everything is correct

packages/frontend-2/components/header/NavUserMenu.vue
```diff
@@ -208,6 +215,22 @@ onMounted(() => {
...
+
+      if (errorQuery && isString(errorQuery)) {
+        triggerNotification({
+          type: ToastNotificationType.Danger,
+          title: errorQuery
+        })
+      } else {
+        triggerNotification({
+          type: ToastNotificationType.Success,
+          title: 'Lekker bezig ah niffo'
```
Mikehrn, line 229: @cdriesler I wanted to change this ah niffo (🪦), but I'm not sure what is the correct message, is it something like "SSO settings successfully updated"?

cdriesler: "SSO settings successfully updated" is good 👍

cdriesler (approved):

PR #3201, opened 2024-10-08
chore(server): IoC 29 - getStream(s)Factory

fabis94:
scary! one of the most used repo fns probably


packages/server/modules/accessrequests/graph/resolvers/index.ts
```diff
@@ -20,13 +20,14 @@ import {
 import { Resolvers } from '@/modules/core/graph/generated/graphql'
```
fabis94, line 20: I tried to find all usages through the following regex: `getStreams?[^\w]`

packages/server/modules/core/services/streams.js
```diff
@@ -42,8 +42,6 @@ module.exports = {
     return id
   },
 
-  getStream,
```
fabis94, line 45: 100% duplicate

alemagio (approved):

PR #3211, opened 2024-10-09
refactor(fe2): Change workspace creation triggers

andrewwallacespeckle:
## Changes:
- Update settings dialog "Add Workspace" to "Create Workspace"
- Send user to the workspace explainer page when clicking the "+"
- Add a "Create workspace" tooltip to the "+" button in the "New project" dialog
- Add Common Confirm Dialog, which can be used in other cases in future
- Checked if form is dirty. If it is, open confirm dialog, if not, navigate to Workspace Explainer


andrewwallacespeckle:
@benjaminvo While i'm on this, the copy here also seems a bit clumsy:
image
I've updated this to:
"Workspaces offer better project management and higher data security."

But I can revert or update again if you can think of better.

packages/frontend-2/components/projects/AddDialog.vue
```diff
@@ -39,31 +39,32 @@
...
-            mixpanel-event-source="create-project-modal"
-            @cancel="isCreatingWorkspace = false"
-            @workspace-created="onWorkspaceCreated"
-          />
         </template>
       </div>
     </form>
+    <CommonConfirmDialog
+      v-model:open="showConfirmDialog"
+      text="You have unsaved changes. Are you sure you want to leave and create a new workspace?"
```
Mikehrn, line 65: Small thing, but maybe change the copy to `text="You have unsaved changes. Are you sure you want to leave?"` so we can also trigger it on cancel/close?

benjaminvo: I like this suggestion 👍

benjaminvo:
Very nice Andrew! I agree with Mike's copy suggestion above.

Also, we now have the annoying situation where the + action in the product sidebar doesn't do anything if you're already on the workspace explainer page 🥲. Could we in this situation trigger the create workspace modal? I hope that doesn't make the code too ugly..

CleanShot 2024-10-09 at 22 30 22@2x

packages/frontend-2/components/projects/AddDialog.vue
```diff
@@ -114,24 +125,21 @@ const emit = defineEmits<{
...
+const router = useRouter()
+const { handleSubmit, meta } = useForm<FormValues>()
 const { result: workspaceResult } = useQuery(projectWorkspaceSelectQuery, null, () => ({
   enabled: isWorkspacesEnabled.value
 }))
 
 const visibility = ref(ProjectVisibility.Unlisted)
 const selectedWorkspace = ref<ProjectsAddDialog_WorkspaceFragment>()
-const isCreatingWorkspace = ref<boolean>(false)
+const showConfirmDialog = ref(false)
```
Mikehrn, line 136: @andrewwallacespeckle What is the difference between these two? Can it be the same modal, with just a different action on `confirm`?

packages/frontend-2/components/dashboard/Sidebar.vue
```diff
@@ -273,4 +261,19 @@ const openFeedbackDialog = () => {
   showFeedbackDialog.value = true
   isOpenMobile.value = false
 }
+
+const openWorkspaceCreateDialog = () => {
+  showWorkspaceCreateDialog.value = true
+  mixpanel.track('Create Workspace Button Clicked', {
```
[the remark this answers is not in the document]
benjaminvo: Hmmm. Yeah not ideal event. But let's just keep it for now and we'll make sure to filter for the source when doing analysis.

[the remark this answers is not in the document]
benjaminvo: Or I'm not sure actually. You're probably right about changing this now so that we can properly track the before and after of changing this.

Links to the explainer page could instead trigger `Clicked Link to Workspace Explainer` and this should also be triggered when clicking the nav item in the sidebar.

packages/frontend-2/components/projects/AddDialog.vue
andrewwallacespeckle: I did it this way because we don't know what the user is confirming when there is 1 confirm modal. To me it made more sense having 2 separate instances of the confirm modal, as they are confirming different things (1 to close dialog, 1 to navigate to workspace explainer). I've updated this to use 1 modal, but had to introduce a new ref to track what they are confirming.

Mikehrn (approved):

PR #3212, opened 2024-10-09
fix(fe2): Border issues around invite banners

andrewwallacespeckle:
- Removed divide-y which was causing the unnecessary line between invites
- ProjectInviteBanner and WorkspaceInvitesBanner are 2 separate components. This causes a weird visual bug, where both have rounded borders, but there is no gap. I added a small gap to make this seem more intentional.
image
@benjaminvo The second part was out of scope of this ticket, so just want to run this past you before merging.


benjaminvo (approved):
Very nice! And I like the second change

PR #3217, opened 2024-10-09
Feat: Add role descriptions to select inputs

Mikehrn:


benjaminvo (changes requested):
A couple of things:

1) There's a pretty aggressive max height on these. Would be great to increase it a bit so we don't get the scrolling inside. Also, there's no checkmark for the selected item in this invite modal?
CleanShot 2024-10-09 at 21 52 35@2x

2) I see this in server settings:
CleanShot 2024-10-09 at 21 55 44@2x

3) Please remove the hover effect from the selected item. I personally don't like when it switches from a darker gray to a lighter gray on hover :).

4) Minor: Let's do `pr-8` instead of `pr-9` for the select items. It wraps the copy a bit better in some of the select inputs (I don't want to change the copy 😬).

Mikehrn:
Updated the styling and fixed the role display issue

Screenshot 2024-10-09 at 22 24 53

Screenshot 2024-10-09 at 22 36 39

Screenshot 2024-10-09 at 22 36 32

benjaminvo:
Works very nicely now! Minor thing we could do now that we're touching these:

Add a hover tooltip to the Guest server role when it's disabled saying "The Guest user role isn't enabled on the server". And the tooltip direction should probably be to the right or left so it doesn't block the text above or is positioned outside the viewport below.

CleanShot 2024-10-09 at 22 47 36@2x

Something similar could be done for when the Owner item is disabled when inviting to a project but I guess this will change. But here it could say "Guests cannot be project owners".

CleanShot 2024-10-09 at 22 45 44@2x

andrewwallacespeckle (dismissed):

benjaminvo (approved):

PR #3229, opened 2024-10-10
refactor(fe2): Changes to Move Projects button

andrewwallacespeckle:
- Moved the `Move projects` button down next to the `New projects` button and turned it into an actual button (white, similar to the invite button).
  - The button still becomes hidden in <md breakpoint, and instead is added to the ... menu
- Kept it visible to Members and Guests but disabled. On hover show tooltip with: "Only admins can move projects into the workspace".
  - Tooltips don't consistently work on buttons, so i wrapped the button in a div.
- Add "Move projects" button to Quickstart card

### DIFF
--- a/packages/frontend-2/components/workspace/ProjectList.vue
+++ b/packages/frontend-2/components/workspace/ProjectList.vue
@@ -32,9 +32,23 @@
             v-bind="bind"
             v-on="on"
           />
-          <FormButton v-if="!isWorkspaceGuest" @click="openNewProject = true">
-            New project
-          </FormButton>
+          <div class="flex gap-2">
+            <div
+              v-tippy="isWorkspaceAdmin ? undefined : 'You must be a workspace admin'"
+            >
+              <FormButton
+                :disabled="!isWorkspaceAdmin"
+                class="hidden md:block"
+                color="outline"
+                @click="showMoveProjectsDialog = true"
+              >
+                Move projects
+              </FormButton>
+            </div>
+            <FormButton v-if="!isWorkspaceGuest" @click="openNewProject = true">
+              New project
+            </FormButton>
+          </div>
         </div>
       </div>
 
@@ -184,6 +198,7 @@ const projects = computed(() => query.result.value?.workspaceBySlug?.projects)
 const workspaceInvite = computed(() => initialQueryResult.value?.workspaceInvite)
 const workspace = computed(() => initialQueryResult.value?.workspaceBySlug)
 const isWorkspaceGuest = computed(() => workspace.value?.role === Roles.Workspace.Guest)
+const isWorkspaceAdmin = computed(() => workspace.value?.role === Roles.Workspace.Admin)
 const showEmptyState = computed(() => {
   if (search.value) return false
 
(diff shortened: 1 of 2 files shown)


benjaminvo:
Maybe too early for feedback since this is in draft but I see the tooltip even though I'm an admin.

CleanShot 2024-10-10 at 15 58 57@2x

andrewwallacespeckle:
> Maybe too early for feedback since this is in draft but I see the tooltip even though I'm an admin.
> 
> CleanShot 2024-10-10 at 15 58 57@2x

This seems to be a reactivity bug. On reload, it works as expected. I'll look into it now.

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -32,9 +32,30 @@
             v-bind="bind"
             v-on="on"
           />
-          <FormButton v-if="!isWorkspaceGuest" @click="openNewProject = true">
-            New project
-          </FormButton>
+          <div class="flex gap-2">
```
andrewwallacespeckle, line 35: @Mikehrn It seems that v-tippy is not reactive. When you go from a workspace page that you are not an admin of, to one you are an admin, the buttons will correctly be enabled, but the tooltip remains.

Conditionally rendering the surrounding div seems to fix this, but open to suggestions if you have any other ideas.

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -189,12 +181,19 @@ const team = computed(() => props.workspaceInfo.team.items || [])
...
+const isWorkspaceGuest = computed(
+  () => props.workspaceInfo.role === Roles.Workspace.Guest
+)
 const actionsItems = computed<LayoutMenuItem[][]>(() => [
   [
     ...(isMobile.value
       ? [
-          { title: 'Move projects', id: ActionTypes.MoveProjects },
-          { title: 'Invite', id: ActionTypes.Invite }
+          ...(isWorkspaceAdmin.value
```
andrewwallacespeckle, line 191: Previously the mobile buttons where always shown, now they are hidden based on the same permissions as we use on desktop.

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -32,9 +32,30 @@
...
+            <!--- Conditionally apply tooltip only for non-admins and avoid v-tippy reactivity bug -->
+            <div v-if="!isWorkspaceAdmin" v-tippy="'You must be a workspace admin'">
```
andrewwallacespeckle, line 37: @benjaminvo This is different tooltip text to what you suggested in Linear. It seems our tooltips cant handle longer text right now. 

I've shortened it, but feel free to update this, or suggest something better. I can also look at improving the tooltips, and going with the original text if you think it is needed. 

image

andrewwallacespeckle:
> > Maybe too early for feedback since this is in draft but I see the tooltip even though I'm an admin.
> > CleanShot 2024-10-10 at 15 58 57@2x
> 
> This seems to be a reactivity bug. On reload, it works as expected. I'll look into it now.

This has been fixed

packages/frontend-2/components/workspace/ProjectList.vue
benjaminvo: I like it!

Mikehrn (approved):

PR #3242, opened 2024-10-10
feat(fe2): Show regions disabled menu item to all workspace users

andrewwallacespeckle:
- Show the "Regions" workspace menu item to all workspace members/guests/admin
- Keep it disabled for all users


andrewwallacespeckle:
@benjaminvo Can you just confirm that what I did is correct on this one? 

Now, all member roles can see the Regions menu item, but it is still disabled for all.

benjaminvo:
Slight change: let's not show to guests

andrewwallacespeckle:
> Slight change: let's not show to guests

Done :)

Mikehrn (approved):

PR #3243, opened 2024-10-10
Feat: Tooltips on avatar

Mikehrn:
Screenshot 2024-10-10 at 18 17 07


benjaminvo (changes requested):
I like the change! But it's not needed in places where the name is already outputted next to the avatars. There are maybe more but I just found these places in comments and member tables:
 
CleanShot 2024-10-10 at 22 44 01@2x

CleanShot 2024-10-10 at 22 42 22@2x

CleanShot 2024-10-10 at 22 38 21@2x

Mikehrn:
> I like the change! But it's not needed in places where the name is already outputted next to the avatars. There are maybe more but I just found these places in comments and member tables:

Good point, I did a couple but forgot to push it 😬 I disabled it for comment threads as well now, these are all the ones I could think of

benjaminvo (approved):
Nice!

PR #3245, opened 2024-10-10
Fix: Various bugfixes

Mikehrn:
Collection of small bugfixes

https://linear.app/speckle/issue/WEB-1981/swap-the-actions-in-the-delete-server-invite-modal
https://linear.app/speckle/issue/WEB-1991/multi-row-bug-in-select-input
https://linear.app/speckle/issue/WEB-1809/disable-view-all-in-3d-btn-when-0-models
https://linear.app/speckle/issue/WEB-1951/change-delete-version-copy-if-version-doesnt-have-a-version-message
https://linear.app/speckle/issue/WEB-1992/change-copy


benjaminvo (changes requested):
Really nice that you pick up these bug fixes! One thing I noticed: The view all in 3d button looks disabled but still has a hover effect and is clickable

Mikehrn:
> Really nice that you pick up these bug fixes! One thing I noticed: The view all in 3d button looks disabled but still has a hover effect and is clickable

Actually, seems that is the case for all/most disabled buttons across our app lol, I've updated the button component, this should fix it everywhere

packages/ui-components/src/components/form/Button.vue
```diff
@@ -1,24 +1,26 @@
...
-        ? `box-shadow: -1px 1px 4px 0px #0000000a inset; box-shadow: 0px 2px 2px 0px #0000000d;`
-        : ''
-    "
-    @click="onClick"
-  >
-    <Component :is="finalLeftIcon" v-if="finalLeftIcon" :class="iconClasses" />
-    <slot v-if="!hideText">Button</slot>
-    <Component :is="iconRight" v-if="iconRight || !loading" :class="iconClasses" />
-  </Component>
+  <div :class="isDisabled ? 'cursor-not-allowed' : ''">
```
Mikehrn, line 2: This is needed if we want to keep the `cursor-not-allowed` class, because by disabling all pointer events we also disable the option to set the cursor type

Mikehrn: (Oke this kills the tests)

benjaminvo:
> > Really nice that you pick up these bug fixes! One thing I noticed: The view all in 3d button looks disabled but still has a hover effect and is clickable
> 
> Actually, seems that is the case for all/most disabled buttons across our app lol, I've updated the button component, this should fix it everywhere

What, haha. Well disabled buttons are now actually disabled for me BUT aren't all hover effects gone from buttons?

Mikehrn:
> What, haha. Well disabled buttons are now actually disabled for me BUT aren't all hover effects gone from buttons?

I just reverted that, it actually broke a lot, but I will create a ticket to pick it up, because there are various "disabled" buttons now that are still clickable

benjaminvo (approved):

PR #3249, opened 2024-10-11
fix(CXPLA-100): use v1 for tag links

AlanRynne:


gjedlicska (approved):

gjedlicska:
This is fe1 which we're not running any more, but still nice catch

PR #3260, opened 2024-10-14
refactor(fe2): Make it easier to find workspace settings

andrewwallacespeckle:
- Project page - Make AvatarGroup clickable, taking you to manage users
- Project page - Remove "Manage" button
- Workspace page - Add settings button, hidden on mobile
- Workspace page - Make settings appear in context menu on mobile

- Added new prop to UserAvatarGroup - hideTooltips


andrewwallacespeckle:
@benjaminvo Couple of considerations on this before merging:
1. It's maybe getting quite busy along the top on Workspace pages?
image

2. See link to Linear comment here on Tooltips:
https://linear.app/speckle/issue/WEB-1965/make-it-easier-to-find-workspace-settings#comment-afd0e8ac

3. Do you think we should also add the more obvious "Settings" button to Projects header? Maybe the tab being prominent is enough?
image

benjaminvo:
@andrewwallacespeckle 

1. Yeah, very busy. I just pushed some changes that make it a bit less busy. It includes removing the ··· menu on bigger screen sizes by actually getting rid of the "Copy link" action. I don't really think it's needed but we can discuss this later.
2. I replied in Linear.
3. I don't think it's needed on project pages since the tab is there. Related comment: I often get confused if I'm looking at a workspace page or a project page since the designs are so similar – I want to figure out how to differentiate these more.

packages/frontend-2/pages/projects/[id]/index.vue
```diff
@@ -13,11 +13,6 @@
         <ProjectPageHeader :project="project" />
         <div class="flex gap-x-3 items-center justify-between">
           <div class="flex flex-row gap-x-3">
-            <CommonBadge rounded :color-classes="'text-foreground-2 bg-primary-muted'">
```
Comment (benjaminvo) [inline packages/frontend-2/pages/projects/[id]/index.vue:16]: Not really related to the initial goal of this PR but I also removed this model counter on the project page since there's already a count outputted in the tabbed nav.

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -66,15 +66,18 @@
...
-          <button
-            class="block"
-            @click="openSettingsDialog(SettingMenuKeys.Workspace.Members)"
-          >
-            <UserAvatarGroup
-              :users="team.map((teamMember) => teamMember.user)"
-              class="max-w-[104px]"
-            />
-          </button>
+          <div v-tippy="isWorkspaceAdmin ? 'Manage members' : 'View members'">
```
Mikehrn, line 69: Will this work correctly if you switch between workspaces? Sometimes to v-tippy reactive is a bit off

packages/frontend-2/pages/projects/[id]/index.vue
```diff
@@ -29,14 +24,16 @@
...
           <div class="flex flex-row gap-x-3">
-            <UserAvatarGroup :users="teamUsers" class="max-w-[104px]" />
-            <FormButton
-              v-if="canEdit"
-              color="outline"
-              :to="projectCollaboratorsRoute(project.id)"
-            >
-              Manage
-            </FormButton>
+            <div v-tippy="canEdit ? 'Manage collaborators' : 'View collaborators'">
```
Comment (Mikehrn) [inline packages/frontend-2/pages/projects/[id]/index.vue:27]: Same here with the tippy reactivity

packages/frontend-2/components/workspace/header/Header.vue
andrewwallacespeckle: Yeah, I tested this expecting that bug, but for some reason it worked fine in this scenario

Mikehrn:
@benjaminvo @andrewwallacespeckle Click on "View members" as a guest opens the settings dialog on the members page, however currently we hide this item for guests. I think we should either allow them to view it in settings, or make it not clickable

packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -189,17 +201,17 @@ const actionsItems = computed<LayoutMenuItem[][]>(() => [
   [
     ...(isMobile.value
       ? [
-          ...(isWorkspaceAdmin.value
-            ? [{ title: 'Move projects', id: ActionTypes.MoveProjects }]
-            : []),
+          { title: 'Settings', id: ActionTypes.Settings },
```
Mikehrn, line 204: I know it was already like this, but recently we've been adding ... after menu items that toggle a dialog, maybe add them here for all 3?

benjaminvo: Yeah I also went back and forth on this. But would you also add it in the account dropdown then? I don't think it feels natural to put "..." after Settings there. It feels more natural to add "..." when there's another step to an action to make it clear to the user that it doesn't trigger immediately (such as "Leave workspace..."). And opening the Settings page is not an "action".
CleanShot 2024-10-15 at 15 11 06@2x

I don't have a good way to explain this yet :D.

benjaminvo:
> @benjaminvo @andrewwallacespeckle Click on "View members" as a guest opens the settings dialog on the members page, however currently we hide this item for guests. I think we should either allow them to view it in settings, or make it not clickable

Hmm, I'd say maybe don't output this entire avatar group component to Guests then? There might be a situation where you don't want a Guest to see who all the other members are 🤔

Mikehrn:
> Hmm, I'd say maybe don't output this entire avatar group component to Guests then? There might be a situation where you don't want a Guest to see who all the other members are 🤔

I think that's a fair limit to not show it to guests

packages/frontend-2/components/workspace/header/Header.vue
benjaminvo: I think that could work

Mikehrn (approved):

Issue #61, opened 2020-12-16
🎋 Possible to accidentally delete the main branch 


izzylys:
Currently, it is possible to accidentally delete the main branch. I was able to do this on the front end when manipulating multiple branches. For the front end specifically, I think this was due to the fact that it doesn't always update on stream changes without your refreshing the page.

Deleting the main branch should not be possible as it will make your stream inaccessible! A check for this should be introduced in the server to block this from happening.

PR #3287, opened 2024-10-15
chore(server): core IoC #61 - changePassword & validateUserPassword

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/services/users.js
+++ b/packages/server/modules/core/services/users.js
@@ -1,15 +1,10 @@
 'use strict'
-const bcrypt = require('bcrypt')
 const knex = require('@/db/knex')
 const {
   ServerAcl: ServerAclSchema,
   Users: UsersSchema,
   UserEmails
 } = require('@/modules/core/dbSchema')
-const {
-  validateUserPassword,
-  MINIMUM_PASSWORD_LENGTH
-} = require('@/modules/core/services/users/management')
 
 const Users = () => UsersSchema.knex()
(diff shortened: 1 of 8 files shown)


packages/server/modules/core/services/users.js
```diff
@@ -51,16 +41,6 @@ module.exports = {
     return role
   },
 
-  /**
-   * @deprecated {Use changePassword()}
```
fabis94, line 55: So turns out the update & validate TS fns in `management.ts` are unused. So what I did was I converted those to fit the behavior & API of these existing ones, and replaced all usages of these old JS ones with the new ones.

While that's a bigger refactor than we usually attempt in these PRs, we have very good recently added test coverage for both local registration & login and also password reset

fabis94: Plus there's a god damn typo in the name! `validatePasssword`

packages/server/modules/core/services/users/management.ts
```diff
@@ -30,6 +31,10 @@ import { UsersEvents, UsersEventsEmitter } from '@/modules/core/events/usersEmit
 
 export const MINIMUM_PASSWORD_LENGTH = 8
 
+const createPasswordDigest = async (newPassword: string) => {
```
fabis94, line 34: moved this up for organizational purposes

packages/server/modules/core/services/users.js
alemagio: ok, I looked at the changes and they seems fine

PR #3294, opened 2024-10-15
Add otel tracing

adamhathcock:
adding otel packages, initialization and some comments


.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
+{
```
fabis94, line 1: why tho, is this so u can run the app through the GUI? or for debugging? cause for debugging you can just run any node process in VSCode's debug terminal

also - make sure u open the workspace through the workspace.code-workspace file for all of the configs to be properly applied, instead of just opening the folder

packages/server/bin/ts-www
```diff
@@ -7,6 +7,12 @@
  */
 
 require('ts-node/register')
+
+//need to initialize before everything else
+//knex is a singleton controlled by module so can't wait til app init
+const { initOpenTelemetry } = require('../dist/otel')
```
fabis94, line 13: maybe this can be done in app.ts or bootstrap.js to avoid duplication?

.vscode/launch.json
adamhathcock: Mainly because I like pressing a GUI button.   I can walk you through what I'm doing

packages/server/bin/ts-www
adamhathcock: it has be done before knex is created.  I guess I could put it at the top of app.ts.  Can talk about it.

PR #3308, opened 2024-10-16
Feat: New server invites

Mikehrn:
Fixes:
https://linear.app/speckle/issue/WEB-1936/implement-modal-for-server-invites
https://linear.app/speckle/issue/WEB-2001/input-focus-state-looks-off-in-server-invite

- Add dynamic lists of invitees, allowing the users to select the project and server role for each user
It's currently not possible to specify the project and server role per user when inviting in bulk, so instead it's sending out an invite per user. A very small amount of invites are actually 2 or more (~4% over the past year), so it was a trade-off we made
- Toasts are now an array and render as a list if they are multiple (instead of replacing the previous one)
Because one invite and fail and one can be successful we needed the option to show multiple toasts at the same time. When a new toast is add a timeout queue is created (if applicable) to keep track of which toast to dismiss. We can't rely purely on matching indexes so I had to add an ID, so we can match a toast to a timeout in the queue.
- Some small styling fixes and improvements to components

Screenshot 2024-10-17 at 10 54 36

Screenshot 2024-10-17 at 10 54 31


packages/frontend-2/components/settings/server/user/InviteDialog.vue
```diff
@@ -1,119 +1,161 @@
...
+        ])
+      : await inviteUserToServer([
+          {
+            email: invite.email,
+            serverRole: invite.serverRole
+          }
+        ])
+  })
+
+  mp.track('Invite Action', {
```
Mikehrn, line 149: @benjaminvo With the new invite modal we have to modify this event, but I'm not sure what makes most sense. Ideally we still want to track wether people are sending 1 or multiple invites out, so we keep `count`, but `hasProject` is now invite specific, but still useful to track. Any suggestion on how to modify/split up this event?

benjaminvo: Maybe it's best to do one event per invite? It also makes it easier to generate a graph of how many invites are sent.

So `count` is out. It's of course a little interesting to know if multiple invites were sent in the same "session" but I can live without this info.

Mikehrn: Cool 👍

packages/frontend-2/components/settings/server/user/InviteDialog.vue
```diff
@@ -1,119 +1,161 @@
...
+              class="w-full"
+              owned-only
+              show-optional
+              mount-menu-on-body
+              show-label
+              :name="`project-${index}`"
+              @update:model-value="(e) => onUpdateProject(e, index)"
+            />
+          </div>
+          <div class="relative w-4">
```
andrewwallacespeckle, line 50: v-if on CommonTextLink leaves the parent div taking up space
image

andrewwallacespeckle: Maybe this was intentional to stop a layout shift when the icon comes in?

andrewwallacespeckle (dismissed):

packages/frontend-2/lib/common/composables/toast.ts
```diff
@@ -68,7 +103,11 @@ export function useGlobalToast() {
    * Trigger a new toast notification
    */
   const triggerNotification = (notification: ToastNotification) => {
-    stateNotification.value = notification
+    const newNotification = { ...notification, id: uuidv4() }
```
Mikehrn, line 106: Another option would be to add the timeout to the `notification`, but that would mean both `triggerNotification` and `useGlobalToastManager` are concerned with timeout logic, open for suggestions

packages/frontend-2/components/settings/server/user/InviteDialog.vue
Mikehrn: It's indeed intended according to the design :)

packages/frontend-2/components/settings/server/user/InviteDialog.vue
Mikehrn: It's indeed intended according to the design :)

packages/ui-components/src/components/global/ToastRenderer.stories.ts
```diff
@@ -7,7 +7,7 @@ import { ToastNotificationType } from '~~/src/helpers/global/toast'
 import type { ToastNotification } from '~~/src/helpers/global/toast'
 import { useGlobalToast } from '~~/src/stories/composables/toast'
 
-type StoryType = StoryObj<{ notification: ToastNotification }>
+type StoryType = StoryObj<{ notifications: ToastNotification[] }>
```
fabis94, line 10: Add a story for rendering a batch of toasts

packages/ui-components/package.json
```diff
@@ -96,6 +97,7 @@
     "type-fest": "^2.13.1",
     "typescript": "^5.0.4",
     "unplugin-vue-macros": "^2.7.0",
+    "uuid": "^10.0.0",
```
fabis94, line 100: Why, when we already have nanoid?

packages/frontend-2/lib/common/composables/toast.ts
```diff
@@ -1,60 +1,95 @@
...
+      timeouts.value = timeouts.value.filter((t) => t.id !== id)
+    }
   }
 
-  const { start, stop } = useTimeoutFn(() => {
-    dismiss()
-  }, 4000)
+  // Create a timeout for a notification
+  const createTimeout = (notification: ToastNotification) => {
+    const { stop } = useTimeoutFn(() => {
```
fabis94, line 44: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#2a7cf25d251e403fac1ab53946887102

composable invoked in async scope (on mouse click, presumably)

packages/frontend-2/lib/common/composables/toast.ts
```diff
@@ -1,60 +1,95 @@
...
+    Array.isArray(stateNotification.value) ? stateNotification.value : []
+  )
+  const readOnlyNotification = computed(() => currentNotifications.value)
 
-  const dismiss = () => {
-    currentNotification.value = undefined
-    stateNotification.value = undefined
+  // Remove a specific notification from the state
+  const removeNotification = (id: string) => {
+    const index = currentNotifications.value.findIndex((n) => n.id === id)
```
fabis94, line 34: As long as there's only ever 1 reference/copy of the toast object, you don't need an ID, you can just compare object references.

```js
const removeNotification = (notif: Notification) => {
  const idx = currentNoficiations.value.findIndex(n => n === notif)
...
}
```

fabis94 (changes requested):

packages/frontend-2/lib/common/composables/toast.ts
Mikehrn: Good point, thanks, that should remove the need for all the ID logic :)

packages/frontend-2/lib/common/composables/toast.ts
Mikehrn: Yeah it's used on click and whenever the timeout is finished, do you have a suggestion regarding a better approach?

fabis94: Why not use real setTimeout (in CSR)?

PR #3320, opened 2024-10-17
chore(server): core IoC #82 - (most) branches resolvers to TS

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/graph/resolvers/branches.js
+++ b/packages/server/modules/core/graph/resolvers/branches.js
@@ -1,142 +1,18 @@
-'use strict'
-
 const { withFilter } = require('graphql-subscriptions')
-
 const {
   pubsub,
   BranchSubscriptions: BranchPubsubEvents
 } = require('@/modules/shared/utils/subscriptions')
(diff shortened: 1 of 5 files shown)


packages/server/modules/core/graph/resolvers/branches.js
```diff
@@ -1,142 +1,18 @@
...
-        args.branch.streamId,
-        Roles.Stream.Contributor,
-        context.resourceAccessRules
-      )
-
-      const deleted = await deleteBranchAndNotify(args.branch, context.userId)
-      return deleted
-    }
-  },
   Subscription: {
```
fabis94, line 16: Subscription resolvers can't be moved to TS 1:1, they require a different syntax and new type definitions which can only be resolved by using the debugger and looking at real runtime values. Will clean these up later

alemagio: I like that done in a dedicated PR
BTW, if you find this is too much work, we can skip it.
Moving to TS is outside of our current scope

alemagio (approved):

PR #3327, opened 2024-10-17
feat(fe2): Add forum message to feedback dialog

andrewwallacespeckle:
image

Try and stop users submitting product support requests through this, direct them to the forum instead.


andrewwallacespeckle:
@benjaminvo Feel free to edit the copy in this one. 

One thing to raise, and it's maybe personal preference, but I find the line height of the text-body-xs class to be quite gappy. If you agree I can bring it up with Michal when he returns.

Mikehrn:
> @benjaminvo Feel free to edit the copy in this one.
> 
> One thing to raise, and it's maybe personal preference, but I find the line height of the text-body-xs class to be quite gappy. If you agree I can bring it up with Michal when he returns.

I've had the same thought, in a few instances I had to override the line-height because it was too big to fit

benjaminvo:
Yeah I also agree. Let's just decrease the line-height now and let MIchal set a different value if he disagrees :).

And I think the copy is good!

Mikehrn (approved):

PR #3328, opened 2024-10-17
chore(server): core IoC #88 - createObjectFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -1,3 +1,14 @@
-import { ObjectRecord } from '@/modules/core/helpers/types'
+import { ObjectChildrenClosureRecord, ObjectRecord } from '@/modules/core/helpers/types'
+import { Nullable } from '@speckle/shared'
+
+export type SpeckleObjectClosureEntry = ObjectChildrenClosureRecord
 
 export type SpeckleObject = ObjectRecord
+
+export type RawSpeckleObject = Record<string, unknown> & {
+  id: string
+  speckle_type: string
+  totalChildrenCount: number
+  hash?: string
+  __closure: Nullable<Record<string, number>>
+}
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -1,9 +1,21 @@
 import { ObjectChildrenClosureRecord, ObjectRecord } from '@/modules/core/helpers/types'
 import { Nullable } from '@speckle/shared'
+import { OverrideProperties, SetOptional } from 'type-fest'
 
 export type SpeckleObjectClosureEntry = ObjectChildrenClosureRecord
 
 export type SpeckleObject = ObjectRecord
+
+/**
+ * We preemptively serialize a couple of fields (usually knex does it)
+ */
+export type InsertableSpeckleObject = OverrideProperties<
+  SetOptional<SpeckleObject, 'createdAt'>,
+  {
+    data: string
+    totalChildrenCountByDepth: string
+  }
+>
 
 export type RawSpeckleObject = Record<string, unknown> & {
   id: string
--- a/packages/server/modules/core/services/objects.js
+++ b/packages/server/modules/core/services/objects.js
@@ -1,68 +1,15 @@
-'use strict'
 const { performance } = require('perf_hooks')
-const crypto = require('crypto')
 const { set, get, chunk } = require('lodash')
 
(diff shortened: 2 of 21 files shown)


packages/server/modules/core/services/objects.js
```diff
@@ -1,68 +1,15 @@
...
-          minDepth: object.__closure[prop]
-        })
-
-        if (totalChildrenCountByDepth[object.__closure[prop].toString()])
-          totalChildrenCountByDepth[object.__closure[prop].toString()]++
-        else totalChildrenCountByDepth[object.__closure[prop].toString()] = 1
-      }
-    }
-
-    delete insertionObject.__tree
```
fabis94, line 43: @didimitrie @gjedlicska removed these 2 lines cause there was no way for these props to ever be there - prepInsertionObject doesn't return them

packages/server/modules/core/services/objects/management.ts
```diff
@@ -0,0 +1,105 @@
...
+    data: stringifiedObj as unknown as Record<string, unknown>,
+    streamId,
+    id: obj.id,
+    // YEAH, this has been broken forever...
+    // speckleType: obj.speckleType
+    speckleType: obj.speckle_type
+  }
+}
+
+export const createObjectFactory =
```
fabis94, line 51: here's the new factory

packages/server/modules/core/services/objects/management.ts
```diff
@@ -0,0 +1,105 @@
...
+  const objectByteSize = estimateStringMegabyteSize(stringifiedObj)
+  if (objectByteSize > MAX_OBJECT_SIZE_MB) {
+    throw new ObjectHandlingError(
+      `Object too large. Object ID: ${obj.id}. (${objectByteSize} MB is > than limit, ${MAX_OBJECT_SIZE_MB} MB)`
+    )
+  }
+
+  return {
+    // Not sure why we stringify it, cause knex will do that, but leaving it as is for now
+    data: stringifiedObj as unknown as Record<string, unknown>,
```
fabis94, line 42: @didimitrie @gjedlicska Any clues on why we do this (see comment)? Usually for Jsonb columns the TS/JS data type is an object, cause Knex knows how to handle the objects and convert them to serialized JSON when inserting, and parsing the JSON when retrieving from the DB.

Is this some kind of preemptive optimization or?

fabis94: Same goes for totalChildrenCountByDepth below

fabis94: i'm seeing this used for obj size calculations later on, so i've just adjusted the insertion type to support strings

packages/server/modules/core/domain/objects/types.ts
```diff
@@ -1,3 +1,26 @@
...
+import { OverrideProperties, SetOptional } from 'type-fest'
+
+export type SpeckleObjectClosureEntry = ObjectChildrenClosureRecord
 
 export type SpeckleObject = ObjectRecord
+
+/**
+ * We preemptively serialize a couple of fields (usually knex does it)
+ */
+export type InsertableSpeckleObject = OverrideProperties<
```
fabis94, line 12: @didimitrie @gjedlicska makes sense?

packages/server/modules/core/domain/objects/types.ts
```diff
@@ -1,3 +1,26 @@
...
+  SetOptional<SpeckleObject, 'createdAt'>,
+  {
+    data: string
+    totalChildrenCountByDepth: string
+  }
+>
+
+export type RawSpeckleObject = Record<string, unknown> & {
```
fabis94, line 20: @didimitrie @gjedlicska Does this make sense for the type of the raw object sent into the API (not the final object DB record)? I'm not sure about 'totalChildrenCount' and '__closure' - maybe they should be optional? I was thrown off by a `__closure !== null` check which made me think `undefined` is not a supported value there

fabis94: I think `id` could be optional as well, it gets generated by prepInsertionObject. speckle_type seems to be the only actually required prop

fabis94: See final versions of InsertableSpeckleObject & RawSPeckleObject here - https://github.com/specklesystems/speckle-server/pull/3335/files#diff-d35068824944d012d2120242ee1ed42bfc11ca8d6a257455e3039c48a3f0776d

gjedlicska (approved):

PR #3330, opened 2024-10-17
refactor(fe2): Survicate improvements

andrewwallacespeckle:
Survicate now allow a delay in the triggering of the first survey, so a lot of the complications in this can be removed. We now don't need to track the onboarding date as a cookie. 

This would require the following changes to the NPS survey in survicate:
image

### DIFF
--- a/packages/frontend-2/plugins/survicate.client.ts
+++ b/packages/frontend-2/plugins/survicate.client.ts
@@ -1,26 +1,9 @@
-import { useOnAuthStateChange } from '~/lib/auth/composables/auth'
-import { useActiveUser } from '~~/lib/auth/composables/activeUser'
-import { useSynchronizedCookie } from '~/lib/common/composables/reactiveCookie'
-import dayjs from 'dayjs'
 import type { Survicate } from '@survicate/survicate-web-surveys-wrapper'
 import type { Nullable } from '@speckle/shared'
-import { useRoute } from 'vue-router'
+import { useOnAuthStateChange } from '~/lib/auth/composables/auth'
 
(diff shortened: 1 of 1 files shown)


andrewwallacespeckle:
@benjaminvo @Mikehrn 
I've tidied up the Survicate plugin, which should solve a lot of the frustrations with survicate. 

It is now a lot more simple, as they allow us to delay the occurrence of the first NPS survey now.

We'd need to update the targeting of the NPS survey in survicate, but I did some testing and it will work fine. I've attached the suggested targeting settings.

Mikehrn:
Nice :) I wonder, but maybe you already tried this, if this async way of loading Survicate is still needed, or if by now they have fixed it?

andrewwallacespeckle:
> Nice :) I wonder, but maybe you already tried this, if this async way of loading Survicate is still needed, or if by now they have fixed it?

I didn't try once I saw that the survicate-wrapper package is still on the same version as when we first installed it. 

When we first set this live, the problem only happened with certain ad blockers, so I don't want to remove this, incase it works fine for me, but breaks the full web app for people with different ad blockers.

Mikehrn:
> > Nice :) I wonder, but maybe you already tried this, if this async way of loading Survicate is still needed, or if by now they have fixed it?
> 
> I didn't try once I saw that the survicate-wrapper package is still on the same version as when we first installed it.
> 
> When we first set this live, the problem only happened with certain ad blockers, so I don't want to remove this, incase it works fine for me, but breaks the full web app for people with different ad blockers.

Not a very active package 😅

Mikehrn (approved):

PR #3333, opened 2024-10-17
feat(fe2): Update Section Box controls to include visibility

andrewwallacespeckle:
image

(User request)

Allow user to hide the section box including the gizmo after making the "cut".

Now: Section box is reset when you deselect in the section box tool.

Future: Save the current cut when clicking away from the section box tool. Instead show the "Reset filters" button in the bottom so users can reset the model from there.

This is already supported per the [viewer docs](https://speckle.guide/viewer/section-tool-api.html#visible).


packages/viewer/src/modules/LegacyViewer.ts
```diff
@@ -180,6 +180,10 @@ export class LegacyViewer extends Viewer {
     this.sections.enabled = true
   }
 
+  public setSectionBoxVisibility(visible: boolean) {
```
andrewwallacespeckle, line 183: @AlexandruPopovici I made this change to test my frontend work, it works so I've left it in, but can you check incase I've done something wrong here?

One problem is that when the frontend updates the section box visibility, I need to slightly move the camera for the change to be visible. EG. Turn off section box visibility. Section box will remain, until I slightly move the camera. Is this something that should be fixed in the viewer? I can probably fix in frontend, but it would be a bit hacky. 

Don't think there's any great urgency in this task.

AlexandruPopovici: @andrewwallacespeckle 
The viewer employs on-demand rendering. This means that it is not constantly rendering, but rather it only does so when it needs to, or when it's told to. We're doing this mostly for mobile platform's sake where rendering each frame would drain battery pointlessly.
That's why whenever the client does something that needs to trigger a render, it needs to call `requestRender`, otherwise no change can be seen since there is no rendering happening. Moving the camera triggers a render request implicitly, that's why you were seeing that behavior.

I've added the calls to `requestRender` in the places that were missing it. Also, I've added a better approach on showing/hiding the section box that does not go through the `LegacyViewer`. We shouldn't add functionality in there, we actually need to retire it.

Let me know if you're having other issues with this

andrewwallacespeckle: @AlexandruPopovici, that's working perfectly! Thanks for the explanation too, I'll know better for next time.

andrewwallacespeckle:
I have made this so the toggle in the left hand menu toggle visibility. This allows the user to toggle it using shift+b. 

On load, section box is both disabled, and invisible. On first click, it is enabled and set as visible. This will also trigger the "Reset Section Box" button. 

Any further clicks to the sidebar button will simply toggle visibility. A click of reset will return everything to it's original state.

@benjaminvo Can you check the UI and see if you are happy with this implementation? If so I'll get it in today and reply to the user on the forum who requested it.

Mikehrn (approved):

benjaminvo:
@andrewwallacespeckle I like it! Only issue is that the "Reset section box" button appears as soon as you enter the section box tool. It should ideally only appear if the section box has been edited.

If this somehow shows to be difficult you can also just merge this now and create a follow up PR.

andrewwallacespeckle:
> @andrewwallacespeckle I like it! Only issue is that the "Reset section box" button appears as soon as you enter the section box tool. It should ideally only appear if the section box has been edited.
> 
> If this somehow shows to be difficult you can also just merge this now and create a follow up PR.

This change has been made and it is now ready for a final review

benjaminvo (changes requested):
I discovered two bugs: https://www.loom.com/share/8d49f2a3c4224b549736b6ae9c0f963f

1. The section box' outlines are still visible when you deselect the tool. I don't think we want this?
2. If you leave a comment with an active section box, then the section box tool is not active when you click the comment again. The section box is correctly visible in the viewer, but the tool's button is not selected on the left. I'm sure this was confusing – watch the recording...

cc @AlexandruPopovici

AlexandruPopovici:
> 1. The section box' outlines are still visible when you deselect the tool. I don't think we want this?

@benjaminvo  From my point of view the outlines should stay visible while sectioning is active. Conceptually they are not part of the section tool's box (that gets hidden and shown), they are there to highlight the result of sectioning the geometry. And the geometry still being sectioned, they should still be visible. 

If you still want Andrew to hide them, [this](https://speckle.guide/viewer/section-tool-outlines-api.html) is how.

benjaminvo:
Ah, broken Loom link, I've updated it.

Alright, let's keep the outlines while sectioning is active. @AlexandruPopovici But as you can see in the Loom at [1:20](https://www.loom.com/share/8d49f2a3c4224b549736b6ae9c0f963f?t=80&sid=2f635b0c-e2b5-41db-83e1-9a28108b6815) we don't show outlines if you view a comment that was made on sectioned model. Should we be consistent and also show outlines here?

@andrewwallacespeckle This discussion doesn't block this PR. But point 2 from [this comment](https://github.com/specklesystems/speckle-server/pull/3333#pullrequestreview-2378935955) is still relevant.

AlexandruPopovici:
> But as you can see in the Loom at [1:20](https://www.loom.com/share/8d49f2a3c4224b549736b6ae9c0f963f?t=80&sid=2f635b0c-e2b5-41db-83e1-9a28108b6815) we don't show outlines if you view a comment that was made on sectioned model

@benjaminvo Yeah, outlines should be visible there

benjaminvo:
@andrewwallacespeckle I just tested – is it just me or does the section box now reset when you toggle the tool?

fabis94 (approved):

benjaminvo (approved):

PR #3334, opened 2024-10-18
chore(server): core IoC #90 - createObjectsBatchedAndNoClosuresFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/domain/objects/operations.ts
+++ b/packages/server/modules/core/domain/objects/operations.ts
@@ -54,3 +54,9 @@ export type CreateObjectsBatched = (params: {
   objects: RawSpeckleObject[]
   logger?: Logger
 }) => Promise<boolean>
+
+export type CreateObjectsBatchedAndNoClosures = (params: {
+  streamId: string
+  objects: RawSpeckleObject[]
+  logger?: Logger
+}) => Promise<string[]>
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -1,5 +1,5 @@
 import { ObjectChildrenClosureRecord, ObjectRecord } from '@/modules/core/helpers/types'
-import { Nullable } from '@speckle/shared'
+import { Nullable, NullableKeysToOptional } from '@speckle/shared'
 import { OverrideProperties, SetOptional } from 'type-fest'
 
 export type SpeckleObjectClosureEntry = ObjectChildrenClosureRecord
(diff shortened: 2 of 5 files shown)


packages/server/modules/core/domain/objects/types.ts
```diff
@@ -9,12 +9,14 @@ export type SpeckleObject = ObjectRecord
 /**
  * We preemptively serialize a couple of fields (usually knex does it)
  */
-export type InsertableSpeckleObject = OverrideProperties<
-  SetOptional<SpeckleObject, 'createdAt'>,
-  {
-    data: string
-    totalChildrenCountByDepth: string
-  }
+export type InsertableSpeckleObject = NullableKeysToOptional<
```
fabis94, line 12: Cause totalChildrenCount and totalChildrenCountByDepth are actually optional (nullable in DB)

didimitrie: yep, they are. we probably don't need 'em even (@gjedlicska please sanity check this)

packages/server/modules/core/domain/objects/operations.ts
```diff
@@ -54,3 +54,9 @@ export type CreateObjectsBatched = (params: {
   objects: RawSpeckleObject[]
   logger?: Logger
 }) => Promise<boolean>
+
+export type CreateObjectsBatchedAndNoClosures = (params: {
```
gjedlicska, line 58: This interface is not needed, both functions should implement the `CreateObjectsBatched` interface, that is the point, that he no closures is a drop in replacement from the caller's perspective

packages/server/modules/core/domain/objects/types.ts
gjedlicska: all this typing wizardry usually fries my mind

i'd like to simplify this into a common params, insert model and query model 3 type definition instead of this, which took me 5 mins to understand, cause of all the jumping around between shared, typefest and our own type definitions

gjedlicska (changes requested):
functionally i think we're good, some type changes suggested

fabis94:
> all this typing wizardry usually fries my mind

@gjedlicska  I disagree, I think its better to use a utility wrapper than to duplicate all of the keys from the source type (WET) and then forget to update them once the source type changes. This way I've outlined what the actual functional difference from the source type is, without duplicating it again. Not only that, but these are common utility types exported out of type-fest and @speckle/shared, so once you use them a few times its easy to remember what they mean, kind of like lodash utilities.

packages/server/modules/core/domain/objects/operations.ts
fabis94: I checked, they are actually functionally different right now - one returns bool, the other returns string[]

fabis94:
I've added some extra comments to the InsertableSpeckleObject. It is a bit more complicated than usual, but 1) I don't want to type out/duplicate the same properties 2) the utility types used are very common 3) I feel like this outlines what the difference is between it and the source type way better than just a plain duplicated list of properties

NullableKeysToOptional - very common for DB insertion models, converts Nullable keys to Optional, cause Nullable in SQL means that you can effectively omit that prop altogether.
OverrideProperties - The same as Merge<>, except it also validates that the properties being overridden actually exist
SetOptional - another very common one, just sets specified props as optional

gjedlicska (dismissed):

alemagio (approved):

Issue #91, opened 2021-02-02
Viewer should inform the user when there are no visible elements


teocomi:
Small enhancement suggestion: the viewer should inform the user when there are no visible elements in a commit, or maybe the viewer could be disabled alltogether.

teocomi:
Maybe points should be considered as visible elements?

![image](https://user-images.githubusercontent.com/2679513/114425209-c1a07580-9bb0-11eb-8e72-50fcc817d657.png)

didimitrie:
Yes. The current implementation is buggy - it doesn't work for gh based streams (or, I guess, dynamo ones?). I need to do more testing and fix it.

PR #3335, opened 2024-10-18
chore(server): core IoC #91 - createObjectsFactory

fabis94:
## Checklist:

### DIFF
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -7,6 +7,11 @@ export type SpeckleObjectClosureEntry = ObjectChildrenClosureRecord
 export type SpeckleObject = ObjectRecord
 
 /**
+ * TODO: The types are a bit too lax right now, we don't really want to allow missing id, speckle_type, __closure props, so we should revisit
+ * the code in the future and make the necessary changes to ensure that these props are always present
+ */
+
+/**
  * We preemptively serialize a couple of fields (usually knex does it)
  */
 export type InsertableSpeckleObject = NullableKeysToOptional<
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -11,10 +11,8 @@ export type SpeckleObject = ObjectRecord
  * the code in the future and make the necessary changes to ensure that these props are always present
  */
 
-/**
- * We preemptively serialize a couple of fields (usually knex does it)
- */
 export type InsertableSpeckleObject = NullableKeysToOptional<
+  // data & totalChildrenCountByDepth are inserted as strings, which differs from the original type (because of preemptive serialization)
   OverrideProperties<
     // Both createdAt & speckleType have defaults upon insertion
     SetOptional<SpeckleObject, 'createdAt' | 'speckleType'>,
--- a/packages/server/modules/core/domain/objects/types.ts
+++ b/packages/server/modules/core/domain/objects/types.ts
@@ -11,7 +11,8 @@ export type SpeckleObject = ObjectRecord
  */
 export type InsertableSpeckleObject = NullableKeysToOptional<
   OverrideProperties<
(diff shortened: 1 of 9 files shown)


packages/server/modules/core/domain/objects/types.ts
```diff
@@ -11,7 +11,8 @@ export type SpeckleObject = ObjectRecord
  */
 export type InsertableSpeckleObject = NullableKeysToOptional<
   OverrideProperties<
-    SetOptional<SpeckleObject, 'createdAt'>,
+    // Both createdAt & speckleType have defaults upon insertion
```
fabis94, line 14: @didimitrie @gjedlicska Ended up making these changes, I think pretty much all of RawSpeckleObject is optional

packages/server/modules/core/services/objects.js
```diff
@@ -1,100 +1,10 @@
...
-            else totalChildrenCountByDepth[obj.__closure[prop].toString()] = 1
-          }
-        }
-
-        insertionObject.totalChildrenCount = totalChildrenCountGlobal
-        insertionObject.totalChildrenCountByDepth = JSON.stringify(
-          totalChildrenCountByDepth
-        )
-
-        delete insertionObject.__tree
```
fabis94, line 63: deleted these 2 yet again

packages/server/modules/core/graph/resolvers/objects.ts
```diff
@@ -71,7 +79,7 @@ export = {
 
       const ids = await createObjects({
         streamId: args.objectInput.streamId,
-        objects: args.objectInput.objects
+        objects: args.objectInput.objects.filter(isNonNullable)
```
fabis94, line 82: GQL typedefs allow for Array<null> here

packages/server/test/speckle-helpers/commitHelper.ts
```diff
@@ -90,7 +89,7 @@ export type BasicTestCommit = {
 export async function createTestObject(params: { projectId: string }) {
   return await createObject({
     streamId: params.projectId,
-    object: { foo: 'bar' } as unknown as RawSpeckleObject
+    object: { foo: 'bar' }
```
fabis94, line 92: no longer needed since RawSpeckleObject has all optional keys

packages/server/modules/core/domain/objects/types.ts
```diff
@@ -20,9 +21,9 @@ export type InsertableSpeckleObject = NullableKeysToOptional<
 >
 
 export type RawSpeckleObject = Record<string, unknown> & {
-  id: string
-  speckle_type: string
-  totalChildrenCount: number
+  id?: string
```
gjedlicska, line 24: this is infact the case of the insertion object, but i do not want to encourage this.

we're so heavily relying on the `id` and the `speckle_type` and `__closure` property, that it doesn't make sense to allow those to be undefined in our type system.

packages/server/modules/core/domain/objects/types.ts
gjedlicska: see comment below, I understand, this is inline with the DB schema, but the default for `speckleType` has bit us in the past.
There was a bug in the code, where we were specifying a value for `speckle_type` but not for `speckleType` and everything defaulted to `Base`
 TBF i'd like to remove the default and make it mandatory.

gjedlicska (changes requested):
Some type comments

packages/server/modules/core/domain/objects/types.ts
fabis94: @gjedlicska I don't want to go into refactoring mode here, I'm only trying to describe how it works right now. The fact is that through the GQL API and possibly even the REST API you can submit any kind of arbitrary object and we submit all kinds of junk objects through tests.

If we want to change this after, I'm open to it, but not within these multi region PRs. If you want I can leave a comment next to the types outlining your concerns

gjedlicska (approved):

PR #3342, opened 2024-10-18
refactor(fe2): Use webflow api for dashboard stories

andrewwallacespeckle:
Swapping out the Ghost API to Webflow for Dashboard stories.

I had to create a server-side API route because Webflow's API doesn't support client-side requests due to CORS.

I haven't done this before in Nuxt, so let me know if I did anything wrong. I tried to base it on the only other api route we have in fe2. 

Couple of Quirks:
- Webflow api doesn't allow to sort by createdOn dates, only published. This causes old article to come to the top when a small change is made. We filter out items that are older than 6 months old to fix this. 
- No read time available, so I calculated it based on word count.
- The provided Webflow API URL is for Blog, not tutorials and news. This can be updated easily in future when it is fixed in WebFlow. I updated the title to reflect this. 

**I also installed NuxtImg here, to deal with the potentially large images coming from WebFlow**

**Requires new API key to be added to live environments**


packages/frontend-2/components/dashboard/Webflow/Wrapper.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+  <section v-if="!error">
+    <h2 class="text-heading-sm text-foreground-2 mb-4">Blog</h2>
+    <div class="grid grid-cols-1 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4 gap-4">
+      <DashboardWebflowCard
```
Mikehrn, line 5: Maybe we should keep this naming a bit more ambiguous, so we can reuse/repurpose them?

packages/frontend-2/components/dashboard/Webflow/Wrapper.vue
```diff
@@ -0,0 +1,29 @@
...
+        v-for="webflowItem in webflowItems"
+        :key="webflowItem.id"
+        :webflow-item="webflowItem"
+      />
+    </div>
+  </section>
+  <section v-else />
```
Mikehrn, line 12: We should probably have an error/empty state in here, in case we couldn't fetch the posts

Mikehrn (changes requested):
Nice getting this one in quickly :) Some small suggestion, and we will need to change a few things for the env var to work, but I can help you with that on Monday

packages/frontend-2/components/dashboard/Webflow/Wrapper.vue
andrewwallacespeckle: I did consider this, but thought showing the "Blog" title and then saying "Could not fetch" or "No posts available" was a worse UX.

I don't have a strong opinion on this though. What do you think @benjaminvo ?

packages/frontend-2/.env.example
```diff
@@ -42,3 +42,5 @@ NUXT_PUBLIC_ENABLE_DIRECT_PREVIEWS=true
 
 # Ghost API
 NUXT_PUBLIC_GHOST_API_KEY=
+
+NUXT_WEBFLOW_API_TOKEN=your_webflow_api_token_here
```
fabis94, line 46: How is this supposed to work locally? Is everyone supposed to ask for a token? Or should we have a demo one for everyone to use?

packages/frontend-2/components/dashboard/Blog/Card.vue
```diff
@@ -0,0 +1,49 @@
+<template>
+  <NuxtLink :to="webflowItem.url" target="_blank">
+    <div
+      class="bg-foundation border border-outline-3 rounded-xl flex flex-col overflow-hidden hover:border-outline-5 transition"
+    >
+      <NuxtImg
```
fabis94, line 6: What's NuxtImg do for us in this scenario? Does it somehow optimize the links?

packages/frontend-2/components/dashboard/Blog/Wrapper.vue
```diff
@@ -0,0 +1,29 @@
...
+  </section>
+  <section v-else />
+</template>
+
+<script setup lang="ts">
+import type { WebflowItem } from '~/lib/dashboard/helpers/types'
+
+const logger = useLogger()
+
+const { data: webflowData, error } = await useFetch<{
```
fabis94, line 20: Shouldn't you use useAsyncData instead, for proper caching and stuff? useFetch works, but may be too low level, it probably refetches the API every time this component initializes

packages/frontend-2/server/api/webflow.ts
```diff
@@ -0,0 +1,89 @@
...
+// Used to filter to last 6 months' articles to prevent old,
+// recently edited posts from appearing at the top
+const getSixMonthsAgo = (): Date => {
+  const date = new Date()
+  date.setMonth(date.getMonth() - 6)
+  return date
+}
+
+export default defineEventHandler(async (): Promise<{ items: WebflowItem[] }> => {
+  const { webflowApiToken } = useRuntimeConfig()
```
fabis94, line 35: What happens if there's no token set? Do we handle that gracefully?

fabis94 (changes requested):
looks good for the most part

packages/frontend-2/.env.example
andrewwallacespeckle: I've updated the env example to include the real key. This key (although stored in env) does not need to remain private, and has already been hard coded into the Connectors. It only has read access to Webflow.

packages/frontend-2/components/dashboard/Blog/Card.vue
fabis94: @andrewwallacespeckle

>NuxtImg creates variants of the image, and will serve the best one, depending on device and connection quality.

How and when does it do this?

packages/frontend-2/server/api/webflow.ts
andrewwallacespeckle: I adde an error throw if token missing which should prevent invalid requests. Is this enough here do you think?

fabis94: @andrewwallacespeckle Graceful handling would involve ensuring the UI doesn't break - that outside of those article cards everything else looks normal. I don't think we necessarily need to throw/log any errors, altho we can for informational purposes if it doesn't break the app

andrewwallacespeckle: It doesn't break the app. The whole Blog section will be hidden if there is an error (including no api key)
https://github.com/specklesystems/speckle-server/pull/3342

packages/frontend-2/components/dashboard/Blog/Card.vue
andrewwallacespeckle: My mistake, it doesn't create variants, it optimises on-the-fly when they're requested, applying resizing and compression based on the device and network. 

Some googling suggests this happens server-side when the image is first accessed, with results cached for subsequent requests.

I can use the plain old img tag here if you want?

andrewwallacespeckle:
@iainsproat I added a new environment variable here. Can you check I did everything ok?

packages/frontend-2/components/dashboard/Blog/Card.vue
fabis94: @andrewwallacespeckle I'm just curious how it actually works, I do think its a beneficial plugin, but it has to be configured correctly. That said, its probably better than nothing so lets see how this performs live.

fabis94 (approved):

Mikehrn (approved):

packages/frontend-2/.env.example
```diff
@@ -42,3 +42,5 @@ NUXT_PUBLIC_ENABLE_DIRECT_PREVIEWS=true
 
 # Ghost API
 NUXT_PUBLIC_GHOST_API_KEY=
+
+NUXT_WEBFLOW_API_TOKEN=8c9bea4c120742a21ff308cda0bea73f13e89ffe26dcc886990bba353549b652
```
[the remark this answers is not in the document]
andrewwallacespeckle: Thanks for your comments. In this case, it shouldn't be NUXT_PUBLIC_ because it's a server-side API route (it's in fe2/server/api). Server routes can access all environment variables directly.

On the first point, this is not meant to be secret. It only has read access to blog stories, and it has already been added to the Connectors repo in plain text

PR #3362, opened 2024-10-21
fix(workspacs): workspace soft delete

cdriesler:
## Description & motivation

- Raised some flags that it's somewhat dangerous to immediately delete everything associated with a workspace.
- A user-facing way to message and manage this is high-investment
- We can still do the soft delete (for recovery reasons) and handle the other bits later

## Changes:

- Introduces `deleteAfter` on workspace records, set to 30 days after `deleteWorkspace` is called
- Filters out workspaces with `deleteAfter` from all workspace query repo functions

Lots to discuss, fronted and backend, with how we want to surface these soft deleted workspaces...


benjaminvo:
The only frontend I would do + hide workspace as if it has already been deleted.

CleanShot 2024-10-22 at 09 37 14@2x

gjedlicska:
This becomes a bit more complex for paid workspaces, so i think we should not focus on this right now.

PR #3363, opened 2024-10-21
fix(sso): revoke all sso sessions on logout

cdriesler:
## Description & motivation

- SSO sessions are currently too sticky - they should be revoked when the user signs out

## Changes:

- Add `RevokeUserSsoSessions`
- Add workspace sso rest endpoint `/api/v1/auth/logout`, which calls the above repo function
- Makes request during logout, if Sso is enabled

### DIFF
--- a/packages/frontend-2/lib/auth/composables/auth.ts
+++ b/packages/frontend-2/lib/auth/composables/auth.ts
@@ -190,6 +190,7 @@ export const useAuthManager = (
   const getMixpanel = useDeferredMixpanel()
   const postAuthRedirect = usePostAuthRedirect()
   const { markLoggedOut } = useJustLoggedOutTracking()
+  const isSsoEnabled = useIsWorkspacesSsoEnabled()
 
   /**
    * Invite token, if any
@@ -396,6 +397,17 @@ export const useAuthManager = (
       forceFullReload: boolean
     }>
   ) => {
+    if (isSsoEnabled.value) {
+      await fetch(new URL('/api/v1/auth/logout', apiOrigin).toString(), {
+        method: 'POST',
+        headers: authToken.value
+          ? {
+              Authorization: authToken.value
+            }
+          : undefined
+      })
+    }
+
     await saveNewToken(undefined, { skipRedirect: true })
 
     if (!options?.skipToast) {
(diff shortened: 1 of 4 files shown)


packages/frontend-2/lib/auth/composables/auth.ts
```diff
@@ -396,6 +397,17 @@ export const useAuthManager = (
       forceFullReload: boolean
     }>
   ) => {
+    if (isSsoEnabled.value) {
+      await fetch(new URL('/api/v1/auth/logout', apiOrigin).toString(), {
```
fabis94, line 401: @cdriesler  Y tho? Isn't deleting the token from the client enough? Surely if its enough for our main auth, its also enough for SSO?

packages/server/modules/workspaces/rest/sso.ts
```diff
@@ -119,6 +120,18 @@ router.get(
   }
 )
 
+router.post('/api/v1/auth/logout', async ({ context, res }) => {
```
fabis94, line 123: This file already existed before with the same pattern, but you should generally avoid this kind of heavy sideffects in the module root scope, cause all of this will be invoked just by virtue of importing this file, which could theoretically break stuff or cause logic to be invoked way earlier than you anticipate it to.

Generally better to avoid all of this and export a function instead that you then invoke explicitly on Speckle module init

packages/frontend-2/lib/auth/composables/auth.ts
cdriesler: It's enough to sign you out, but not enough to clear the session. We are not linking "SSO sessions" to the web app sign-in because of how we want them to work with PATs:

- A PAT represents a given user
- A PAT is used in connectors (or other things) to access project data
- An SSO session is created when a user signs in successfully with SSO
- If a the user a PAT is associated with has a valid SSO session for the workspace being accessed, allow
  - If not, disallow, redirect to SSO flow on app

The provision here for clearing the SSO session is to make "sign out" literally mean sign in for SSO access as well.

packages/server/modules/workspaces/rest/sso.ts
cdriesler: Roger, do you mind if I submit this change to you in another PR? There's an unfortunate clash with this and the big SSO PR atm.

cdriesler:
Closed pending future discussions about how "SSO sessions" relate to the lifetime of app/pat tokens

PR #3375, opened 2024-10-22
feat(fe2): Add change role dialog to guests table

andrewwallacespeckle:
Adding SettingsSharedChangeRoleDialog to guests table, triggered by the actions menu. 

Also adding a prop to this, to allow us to hide items from the role select, in this case, hide guests role when editing a guest.

### DIFF
--- a/packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
+++ b/packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
@@ -87,8 +87,6 @@
 
     <SettingsSharedChangeRoleDialog
       v-model:open="showChangeUserRoleDialog"
-      :name="userToModify?.user.name ?? ''"
-      :is-workspace-admin="isWorkspaceAdmin"
       :workspace-domain-policy-compliant="
         userToModify?.user.workspaceDomainPolicyCompliant
       "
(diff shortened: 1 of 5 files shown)


packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
```diff
@@ -84,6 +84,17 @@
       :user="userToModify"
       :workspace-id="workspaceId"
     />
+
+    <SettingsSharedChangeRoleDialog
+      v-model:open="showChangeUserRoleDialog"
+      :name="userToModify?.user.name ?? ''"
```
Mikehrn, line 90: I think this prop is actually redundant, seems it's not used in this component anymore, let's remove it

packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
```diff
@@ -84,6 +84,17 @@
...
+      :is-workspace-admin="isWorkspaceAdmin"
```
Mikehrn, line 91: Same with this one, it doesn't exist in `SettingsSharedChangeRoleDialog` however we probably should add some logic for this, because you can only make someone an admin when you are one

packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
```diff
@@ -231,4 +251,20 @@ const onRemoveUser = async () => {
...
+const onUpdateRole = async (newRoleValue: WorkspaceRoles) => {
+  if (!userToModify.value || !newRoleValue) return
+
+  await updateUserRole({
+    userId: userToModify.value.id,
+    role: newRoleValue,
+    workspaceId: props.workspaceId
+  })
+
+  mixpanel.track('Workspace User Role Updated', {
```
Mikehrn, line 264: I think this event will also be sent if `updateUserRole` was executed but not succes, probably need to check the result of this function and see wether it was successful or not

Mikehrn (changes requested):

packages/frontend-2/components/settings/workspaces/members/GuestsTable.vue
andrewwallacespeckle: Good point, I moved the mixpanel events to management.ts, so we could accurately track this.

Mikehrn (dismissed):

Mikehrn (approved):

Mikehrn (approved):

PR #3378, opened 2024-10-22
webhook-service  refactor multiregion

alemagio:
I made this a single PR since it is very small.
I know that to be precise we should also migrate dependencies that are not `db` related (like the logger), I just ignored them for now since I would really like to move quickly to the next step of the project.

cc @gjedlicska

### DIFF
--- a/packages/webhook-service/src/main.js
+++ b/packages/webhook-service/src/main.js
@@ -12,10 +12,12 @@ const HEALTHCHECK_FILE_PATH = '/tmp/last_successful_query'
 const { makeNetworkRequest } = require('./webhookCaller')
 const WebhookError = require('./errors')
 
-async function startTask() {
-  const { rows } = await knex.raw(`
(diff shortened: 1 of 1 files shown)


fabis94 (approved):
Looks OK, t hankfully a very tiny service. It does looks like it would be pretty easy to introduce TS here, maybe worth bringing up with the rest of the team?

Also - did you manage testing it somehow? I don't think I ever have, so it would be useful to add some testing instructions to the readme

alemagio:
> Looks OK, t hankfully a very tiny service. It does looks like it would be pretty easy to introduce TS here, maybe worth bringing up with the rest of the team?
> 
> Also - did you manage testing it somehow? I don't think I ever have, so it would be useful to add some testing instructions to the readme

For TS I created tickets.
Testing yes, I tested it locally and webhooks are fired properly

PR #3382, opened 2024-10-23
fix(server): on sign up send out only 1 verification email

fabis94:


packages/server/modules/emails/services/verification/request.ts
```diff
@@ -184,25 +183,11 @@ export const requestEmailVerificationFactory =
...
-/**
- * Listen for user:created events and trigger email verification initialization
- */
-export const initializeVerificationOnRegistrationFactory =
-  (deps: {
-    userEmitterListener: typeof UsersEmitter.listen
-    requestEmailVerification: RequestEmailVerification
-  }) =>
-  () => {
-    return deps.userEmitterListener(UsersEvents.Created, async ({ user }) => {
```
fabis94, line 196: This was the redundant one. Upon adding multiple emails we added another way of triggering email verification directly in the user creation service and both of these were firing.

packages/server/modules/emails/tests/verifications.spec.ts
```diff
@@ -70,10 +70,15 @@ describe('Email verifications @emails', () => {
...
+  afterEach(async () => {
+    mailerMock.resetMockedFunctions()
+  })
+
+  it('sends out 1 verification email immediately after new account creation', async () => {
     const sendEmailInvocations = mailerMock.hijackFunction(
       'sendEmail',
-      async () => true
+      async () => true,
+      { times: 2 }
```
fabis94, line 81: Improved tests to catch this in the future

packages/shared/src/core/helpers/tracking.ts
```diff
@@ -7,3 +7,5 @@ export function resolveMixpanelUserId(email: string): string {
 export function resolveMixpanelServerId(serverHostname: string): string {
   return md5(serverHostname.toLowerCase()).toUpperCase()
 }
+
+export const resolveDistinctId = resolveMixpanelUserId
```
fabis94, line 11: the original name is a bit unintuitive, I can never remember what it's called

PR #3404, opened 2024-10-28
refactor(fe2): Send user to dashboard after tour - skip checklist

andrewwallacespeckle:
Now: After completing the "model tour" onboarding we send the user to their first sample project and show the 4-step guide.

Change to: Send the user to the Dashboard instead after completing the model tour. It's a better experience and we don't want this outdated 4-step guide.

** Note: I haven't been able to test this properly, my local onboarding flow is broken. I'll fix this and test properly tomorrow. Should not be merged until it has been confirmed working. **


benjaminvo:
@andrewwallacespeckle

1. Are you skipping the entire onboarding? When I visit `localhost:8081/onboarding?force=true` I'm just sent straight to the new project's page. I'd like to keep the guided journey through a 3D model but when you reach the end go to the Dashboard instead of the viewer.

2. As mentioned above, let's just send the user to the Dashboard instead of their new project.

Mikehrn:
I just tested this locally, and for me it works as described, I get the 3D tour (either with a new account or using the force flag) and then get send to my dashboard when completed. Only the last step feels a bit jumpy because it shows the model viewer page for 1-2 secs and then jumps to the dashboard


https://github.com/user-attachments/assets/88b7a014-44ce-4c04-8e81-9545ba3fd874

andrewwallacespeckle:
@benjaminvo I got my local onboarding working again if you want to get yours working again. 

In `/packages/server/.env`, there is a variable called `ONBOARDING_STREAM_CACHE_BUST_NUMBER`. Increasing this to 2 and restarting server and fe2 fixed my issue. 

image

benjaminvo:
@andrewwallacespeckle I like it! Only thought is if we should add a hover tooltip to the Finish button? "First add another model" or something. I'm worried they don't why it's disabled and close the window.

packages/frontend-2/components/tour/Comment.vue
```diff
@@ -83,6 +99,7 @@ defineEmits(['skip', 'previous', 'next'])
 const props = defineProps<{
   index: number
   item: SlideshowItem
+  hasAddedOverlay: boolean
```
Mikehrn, line 102: Should this be a more generic `disableNext`? Because `hasAddedOverlay` is only relevant voor 1 specific TourComment

packages/frontend-2/components/tour/Onboarding.vue
```diff
@@ -4,41 +4,20 @@
   >
     <TourSegmentation v-if="showSegmentation && step === 0" @next="step++" />
     <TourSlideshow v-if="step === 1" @next="step++" />
```
Mikehrn, line 6: I think the step number maybe can be removed? I think one boolean should be enough? `showSlideshow` or something after `TourSegmentation` emitted `next`

Mikehrn (changes requested):

Mikehrn (approved):

PR #3406, opened 2024-10-29
feat(server): multi region module base w/ a test (mocked) mutation

fabis94:


packages/server/codegen.yml
```diff
@@ -25,7 +25,6 @@ generates:
         User: '@/modules/core/helpers/graphTypes#UserGraphQLReturn'
         ActiveUserMutations: '@/modules/core/helpers/graphTypes#MutationsObjectGraphQLReturn'
         UserEmailMutations: '@/modules/core/helpers/graphTypes#MutationsObjectGraphQLReturn'
-        UserWorkspaceMutations: '@/modules/core/helpers/graphTypes#MutationsObjectGraphQLReturn'
```
fabis94, line 28: GQL type no longer exists

packages/server/modules/index.ts
```diff
@@ -77,7 +78,8 @@ const getEnabledModuleNames = () => {
     'serverinvites',
     'stats',
     'webhooks',
-    'workspacesCore'
+    'workspacesCore',
+    'multiregion'
```
fabis94, line 82: As mentioned before, the module itself is enabled always (for simplicity), but then functionality checks inside of it can use the feature flag

packages/server/modules/index.ts
```diff
@@ -93,7 +95,13 @@ async function getSpeckleModules() {
   const moduleNames = getEnabledModuleNames()
 
   for (const dir of moduleNames) {
-    loadedModules.push(require(`./${dir}`))
+    const moduleIndex = await import(`./${dir}/index`)
+    const moduleDefinition = 'init' in moduleIndex ? moduleIndex : moduleIndex.default
```
fabis94, line 99: Finally supporting `export default {}` in module index files, no longer having to do any of this `export = {}` bullshit

packages/server/modules/index.ts
```diff
@@ -185,7 +193,10 @@ const graphComponents = (): Pick<ApolloServerOptions<any>, 'resolvers'> & {
     // first pass load of resolvers
     const resolversPath = path.join(fullPath, 'graph', 'resolvers')
     if (fs.existsSync(resolversPath)) {
-      resolverObjs = [...resolverObjs, ...values(autoloadFromDirectory(resolversPath))]
+      const newResolverObjs = values(autoloadFromDirectory(resolversPath)).map((o) =>
```
fabis94, line 196: Similarly supporting `export default {}` in GQL resolver modules

packages/server/modules/multiregion/graph/resolvers/index.ts
```diff
@@ -0,0 +1,5 @@
+import { Resolvers } from '@/modules/core/graph/generated/graphql'
+
+// TODO: Real implementation? Need to discuss questions regarding data model & region config first
```
fabis94, line 3: I think I got my answers earlier today, so I should be unblocked for this

PR #3410, opened 2024-10-29
fix(fe2): Add workspace name to workspace page title

andrewwallacespeckle:
Added the workspace name to the page title on the workspace page. 

Set it once we have the name available.


packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -277,6 +277,9 @@ const onShowSettingsDialog = (target: AvailableSettingsMenuKeys) => {
 onResult((queryResult) => {
   if (queryResult.data?.workspaceBySlug) {
     workspaceMixpanelUpdateGroup(queryResult.data.workspaceBySlug)
+    useHead({
```
Mikehrn, line 280: Not sure if we sanitize user data, but might be worth to check `useHeadSafe` because workspace name is user input based; https://nuxt.com/docs/api/composables/use-head-safe

PR #3376, opened 2024-10-22
SSO

cdriesler:
```
      ___           ___           ___     
     /\__\         /\__\         /\  \    
    /:/ _/_       /:/ _/_       /::\  \   
   /:/ /\  \     /:/ /\  \     /:/\:\  \  
  /:/ /::\  \   /:/ /::\  \   /:/  \:\  \ 
 /:/_/:/\:\__\ /:/_/:/\:\__\ /:/__/ \:\__\
 \:\/:/ /:/  / \:\/:/ /:/  / \:\  \ /:/  /
  \::/ /:/  /   \::/ /:/  /   \:\  /:/  / 
   \/_/:/  /     \/_/:/  /     \:\/:/  /  
     /:/  /        /:/  /       \::/  /   
     \/__/         \/__/         \/__/    
```

PR #3420, opened 2024-10-31
feat(sso): list sso associations by user email

cdriesler:
Depends on #3376 

## Description & motivation

- We are adding "continue with SSO" to the sign in flow, which will prompt users to sign in with SSO after they provide their email
- To do so, we need to expose limited information (at least slug) about valid SSO targets to continue to

## Changes:

- Adds repo function for listing, given a user id:
  - Workspaces they are in
  - That also have SSO configured

- Exposes it via public rest endpoint (via service that limited info emitted):

image

The repo function will be used for incoming work around protecting workspace assets.


cdriesler:
Working via gql:

image

gjedlicska (approved):

PR #3437, opened 2024-11-05
feat: set default workspace region

fabis94:
did as much as I could without real multi region DBs


packages/server/modules/workspaces/repositories/regions.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+export const deleteAllRegionAssignmentsFactory =
+  (deps: { db: Knex }): DeleteAllRegionAssignments =>
+  async (params) => {
+    const { workspaceId } = params
+    await tables.workspaceRegions(deps.db).where({ workspaceId }).delete()
+  }
+
+export const getDefaultRegionFactory =
```
gjedlicska, line 42: lets add integration tests to this function

packages/server/modules/workspacesCore/domain/types.ts
```diff
@@ -55,3 +56,5 @@ export type WorkspaceAcl = {
   workspaceId: string
   createdAt: Date
 }
+
+export type WorkspaceRegionAssignment = WorkspaceRegionRecord
```
gjedlicska, line 60: Domain never relies on database implementation

packages/server/modules/workspaces/services/regions.ts
```diff
@@ -0,0 +1,39 @@
...
+  GetAvailableRegions,
+  UpsertRegionAssignment,
+  DeleteAllRegionAssignments
+} from '@/modules/workspaces/domain/operations'
+import { WorkspaceRegionAssignmentError } from '@/modules/workspaces/errors/regions'
+
+export const getAvailableRegionsFactory =
+  (deps: { getRegions: GetRegions }): GetAvailableRegions =>
+  async () => {
+    // TODO: Gatekeeper checks here? For now just returning all server regions
```
gjedlicska, line 13: Yup, lets inject, the ` canWorkspaceUseRegions` function, and if it returns false, lets throw.

gjedlicska (changes requested):

packages/server/modules/workspacesCore/migrations/20241105070219_create_workspace_regions_table.ts
```diff
@@ -0,0 +1,32 @@
...
+    table
+      .text('workspaceId')
+      .references('id')
+      .inTable(WORKSPACES_TABLE)
+      .onDelete('cascade')
+      .notNullable()
+    table
+      .string('regionKey')
+      .references('key')
+      .inTable(REGIONS_TABLE)
```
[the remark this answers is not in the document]
fabis94: @gjedlicska Why do we need workspace_regions to be replicated in the regional shard?

packages/server/modules/workspaces/services/regions.ts
fabis94: @gjedlicska Ok, but there also has to be a way to get the specific regions accessible to the workspace. Some can be already bought, others may not be.

gjedlicska: for now, that doesn't matter. You are not buying access on a per region basis

fabis94: @gjedlicska it does matter in the context of this upcoming task https://linear.app/speckle/issue/WEB-1837/list-available-regions-on-the-server-in-workspace-settings

packages/server/modules/workspaces/services/regions.ts
```diff
@@ -0,0 +1,39 @@
...
+        'Specified region not available for workspace',
+        {
+          info: { params }
+        }
+      )
+    }
+
+    // For now, only 1 region - delete old assignments
+    await deps.deleteAllRegionAssignments({ workspaceId })
+    await deps.upsertRegionAssignment({ workspaceId, regionKey })
```
gjedlicska, line 38: In here, we need to copy the workspace record into the region DB, so that projects in the region DB, can use the `workspaceId` foreign key

packages/server/modules/workspacesCore/migrations/20241105070219_create_workspace_regions_table.ts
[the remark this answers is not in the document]
gjedlicska: stream.workspaceId is an FK to workspace.id, if we write the stream to a region, we need the workspace in the region too, if workspace.regionKey is an FK to region.key, we'd need the regions in the regional db too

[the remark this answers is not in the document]
gjedlicska: For now, this feels like an OK cut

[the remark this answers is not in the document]
gjedlicska: @fabis94 I'm wrong, this is not the workspaces table. The FK here makes sense

packages/server/modules/workspaces/services/regions.ts
fabis94: @gjedlicska Also, 1) I'm not able to find "canWorkspaceUseRegions" anywhere in the codebase 2) Why do you think this should throw instead of returning an empty array?

fabis94:
gonna be merged w/ gergos pr

PR #3455, opened 2024-11-07
feat(server): multi region support in dataloaders

fabis94:
So all of the preexisting dataloaders still exist but only use the main DB. If you want to spawn a set of loaders for your specific region, you have to do `ctx.loaders.forRegion({ db })` and then this set will use that new region DB.

Loaders of course won't be shared across regions. So if you do `ctx.loaders.streams.getStream.load()` and `ctx.loaders.forRegion({ db: regionDb }).streams.getStream.load()` these 2 will be 2 different queries in 2 different databases.

### DIFF
--- a/packages/server/modules/core/loaders.ts
+++ b/packages/server/modules/core/loaders.ts
@@ -1,145 +1,42 @@
 import DataLoader from 'dataloader'
-import {
-  getStreamsFactory,
-  getCommitStreamsFactory,
-  getBatchUserFavoriteDataFactory,
-  getBatchStreamFavoritesCountsFactory,
-  getOwnedFavoritesCountByUserIdsFactory,
-  getStreamRolesFactory,
-  getUserStreamCountsFactory,
-  getStreamsSourceAppsFactory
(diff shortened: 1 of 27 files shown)


packages/server/modules/core/loaders.ts
```diff
@@ -167,531 +63,67 @@ const buildDataLoaderCreator = (selfClearing = false) => {
...
   const createLoader = buildDataLoaderCreator(options?.cleanLoadersEarly || false)
   const modulesLoaders = graphDataloadersBuilders()
 
-  const loaders = {
-    ...(Object.assign(
-      {},
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-return
-      ...modulesLoaders.map((l) => l({ ctx, createLoader }))
-    ) as Record<string, unknown>),
-    streams: {
```
fabis94, line 185: Moved all of the `core` dataloaders to their own module so that this file is only about dataloader bootstrapping

alemagio (approved):

PR #3464, opened 2024-11-08
feat(fe2): Frontend SSO Integration

andrewwallacespeckle:
Adds SSO to frontend. 

- New middleware
- New Login Page
- New slug/sso error page
- Updated styles

Still some things not done:
- Some discussions ongoing with backend over how (or if) we end an sso session
- Server change to remove projects from Project page when no access
- Update server so it acts on Newsletter Subscribe checkbox for SSO registraions
- Add Delete config functionality once its in backend
- We need the backend to include an access_code in the registration redirect URL, just like it does for email login.


packages/frontend-2/components/settings/workspaces/security/sso/Form.vue
```diff
@@ -0,0 +1,142 @@
...
+  providerName: props.initialData?.providerName ?? '',
+  clientId: props.initialData?.clientId ?? '',
+  clientSecret: props.initialData?.clientSecret ?? '',
+  issuerUrl: props.initialData?.issuerUrl ?? ''
+})
+
+const { handleSubmit } = useForm<SsoFormValues>()
+
+const handleCreate = () => {
+  const baseUrl = `${apiOrigin}/api/v1/workspaces/${props.workspaceSlug}/sso/oidc/validate`
```
fabis94, line 101: Why not use `new URL()` with its utils for setting query params (URL.searchParams), instead of doing all of this string concatenation?

packages/frontend-2/components/settings/workspaces/security/sso/Wrapper.vue
```diff
@@ -0,0 +1,190 @@
...
+  return `${apiOrigin}/api/v1/workspaces/${props.workspace.slug}/sso/oidc/callback?validate=true`
+})
+
+const formInitialData = computed(() => {
+  if (!isEditing.value || !provider.value) return undefined
+
+  return {
+    providerName: provider.value.name ?? '',
+    clientId: provider.value.clientId ?? '',
+    clientSecret: '••••••••••••••••••',
```
fabis94, line 186: Do we make sure this won't get saved and override the actual secret in downstream components?

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -417,6 +423,76 @@ export function useCreateWorkspace() {
   }
 }
 
+export const useWorkspaceSso = (params: { workspaceSlug: string }) => {
```
fabis94, line 426: Can we keep all of these in their own file? management.ts is already at ~400 LOC, SSO composables can have their own file

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -417,6 +423,76 @@ export function useCreateWorkspace() {
...
+    hasSsoEnabled,
+    isSsoAuthenticated,
+    needsSsoLogin,
+    provider,
+    loading,
+    error
+  }
+}
+
+export function useWorkspaceSsoPublic(workspaceSlug: string) {
```
fabis94, line 456: Some of these names could be improved. I've no idea what this composable does based on its name. And why is the loading only done on mounted, not in SSR?

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -522,3 +598,71 @@ export const useSetDefaultWorkspaceRegion = () => {
...
+
+export function useSsoAuth(workspaceSlug: string) {
+  const logger = useLogger()
+  const ssoError = ref<string | null>(null)
+
+  const { hasSsoEnabled, needsSsoLogin, error } = useWorkspaceSso({
+    workspaceSlug
+  })
+
+  if (error.value) {
```
fabis94, line 610: None of this is reactive. If useWorkspaceSso recalculates, these downstream values (ssoError) will remain out of date

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -522,3 +598,71 @@ export const useSetDefaultWorkspaceRegion = () => {
...
+
+export function useWorkspaceSsoDelete() {
+  const apiOrigin = useApiOrigin()
+  const { triggerNotification } = useGlobalToast()
+  const mixpanel = useMixpanel()
+
+  const deleteSsoProvider = async (workspaceSlug: string) => {
+    try {
+      const res = await fetch(
+        new URL(`/api/v1/workspaces/${workspaceSlug}/sso`, apiOrigin),
```
fabis94, line 628: @cdriesler Why are a lot of these REST endpoints? Feels like a few of these could just be GQL queries/mutations

packages/frontend-2/lib/workspaces/graphql/queries.ts
```diff
@@ -78,3 +78,36 @@ export const validateWorkspaceSlugQuery = graphql(`
     validateWorkspaceSlug(slug: $slug)
   }
 `)
+
+export const workspaceSsoByEmailQuery = graphql(`
+  query WorkspaceSsoByEmail($email: String!) {
+    workspaceSsoByEmail(email: $email) {
+      id
```
fabis94, line 85: Shouldn't these be fragment based? Where are these requirements coming from?

fabis94 (changes requested):

packages/frontend-2/pages/workspaces/[slug]/sso/index.vue
```diff
@@ -1,50 +1,117 @@
...
+const workspaceSlug = computed(() => route.params.slug as string)
+const { isSsoAuthenticated } = useWorkspaceSso({ workspaceSlug: workspaceSlug.value })
+
+const { result } = useQuery(authRegisterPanelQuery, {
+  token: route.query.token as string
+})
+
+const { workspace, loading, error } = useWorkspaceSsoPublic(workspaceSlug.value)
+
+if (error.value) {
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[slug]/sso/index.vue:85]: not reactive https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#8a1d83812bf24d7ca4fabfa6d521cfd9

packages/frontend-2/components/auth/sso/Login.vue
```diff
@@ -0,0 +1,187 @@
...
+import { workspaceSsoByEmailQuery } from '~/lib/workspaces/graphql/queries'
+import { useAuthManager, useLoginOrRegisterUtils } from '~/lib/auth/composables/auth'
+import { useDebounceFn } from '@vueuse/core'
+import type { WorkspaceSsoByEmailQuery } from '~/lib/common/generated/gql/graphql'
+
+type Workspace = WorkspaceSsoByEmailQuery['workspaceSsoByEmail'][number]
+
+const loading = ref(false)
+const email = ref('')
+const emailCheckState = ref<'idle' | 'checking' | 'checked'>('idle')
```
Mikehrn, line 83: Does check also means valid, or only that it has been checked?

packages/frontend-2/components/auth/sso/Register.vue
```diff
@@ -0,0 +1,75 @@
...
+const serverInfo = computed(() => result.value?.serverInfo)
+const workspaceSlug = computed(() => route.params.slug?.toString() || '')
+const { workspace } = useWorkspaceSsoPublic(workspaceSlug.value)
+
+const handleContinue = () => {
+  if (!workspaceSlug.value) return
+
+  loading.value = true
+  try {
+    mixpanel.track('Workspace SSO Register Attempt', {
```
Mikehrn, line 60: The Mixpanel panels events are in past tense -> `Workspace SSO Register Attempted`. Maybe also add the ones for `Failed` and `Successful` in the `catch` `finally`

packages/frontend-2/components/auth/third-party/LoginButtonBase.vue
```diff
@@ -1,21 +1,14 @@
...
-      'rounded shadow bg-foundation- text-foreground dark:text-foreground-on-primary',
-      'border border-outline-2 hover:border-outline-3',
-      noVerticalPadding ? '' : 'py-2'
-    ]"
-  >
-    <div class="flex items-center justify-center">
-      <slot />
-    </div>
+  <NuxtLink :to="to">
+    <FormButton color="outline" size="lg" full-width class="group">
```
Mikehrn, line 3: If you pass `to:` to the `FormButton` it should render as a NuxtLink component as well

packages/frontend-2/components/settings/workspaces/Security.vue
```diff
@@ -69,17 +77,14 @@
...
         <div class="flex flex-col space-y-8">
           <div class="flex items-center">
             <div class="flex-1 flex-col pr-6 gap-y-1">
               <p class="text-body-xs font-medium text-foreground">Domain protection</p>
-              <p class="text-body-xs text-foreground-2 leading-5 max-w-md">
-                Admins won't be able to add users as members (or admins) to a workspace
-                unless the one of the users email matches one of the workspace's
-                verified email domains.
+              <p class="text-body-2xs text-foreground-2 leading-5 max-w-md">
+                Members won’t be able to add users as members (or admins) to a workspace
```
Mikehrn, line 86: Members already can't invite new people to the workspace, maybe stick with `Admin` in this wording

packages/frontend-2/components/settings/workspaces/security/sso/DeleteDialog.vue
```diff
@@ -0,0 +1,60 @@
...
+const emit = defineEmits<{
+  (e: 'deleted'): void
+}>()
+
+const isOpen = defineModel<boolean>('open', { required: true })
+const { deleteSsoProvider } = useWorkspaceSsoDelete()
+
+const handleRemove = async () => {
+  const success = await deleteSsoProvider(props.workspaceSlug)
+  if (success) {
```
Mikehrn, line 38: What happens when it doesn't succeed? Maybe trigger a toast on both succes and error

packages/frontend-2/lib/workspaces/composables/management.ts
```diff
@@ -417,6 +423,76 @@ export function useCreateWorkspace() {
...
 }
 
+export const useWorkspaceSso = (params: { workspaceSlug: string }) => {
+  const { result, loading, error } = useQuery<WorkspaceSsoCheckQuery>(
+    workspaceSsoCheckQuery,
+    { slug: params.workspaceSlug }
+  )
+
+  const hasSsoEnabled = computed(() => !!result.value?.workspaceBySlug.sso?.provider)
+  const provider = computed(() => result.value?.workspaceBySlug.sso?.provider ?? null)
```
Mikehrn, line 433: Does it need to explicitly `null`? `result.value?.workspaceBySlug.sso?.provider` will be `undefined` if the chaining fails, so you can also use it to check for falsey values

packages/frontend-2/middleware/requireValidProject.ts
```diff
@@ -19,17 +19,46 @@ export default defineNuxtRouteMiddleware(async (to) => {
...
+  if (
+    data?.project?.visibility === ProjectVisibility.Public ||
+    data?.project?.visibility === ProjectVisibility.Unlisted
+  ) {
+    return
+  }
+
+  // Check for SSO session error
+  const ssoSessionError = (errors || []).find(
+    (e) => e.extensions?.['code'] === 'SSO_SESSION_MISSING_OR_EXPIRED_ERROR'
```
Mikehrn, line 37: Is there a constant or enum you can use to get these codes? Now if you are not familiar with the code it seems like a random string

packages/frontend-2/pages/authn/sso/index.vue
```diff
@@ -0,0 +1,9 @@
+<template>
+  <AuthSsoLogin />
+</template>
+
+<script setup lang="ts">
+useHead({
+  title: 'Log in'
```
Mikehrn, line 7: Maybe check with Michal or Benjamin, but we are mixing `Log in` and `Sign in`, maybe good to be consistent :)

packages/frontend-2/pages/workspaces/[slug]/sso/index.vue
```diff
@@ -1,50 +1,117 @@
...
+        </div>
+
+        <!-- Already Authenticated Message -->
+        <div v-if="isSsoAuthenticated" class="border border-outline-3 rounded p-4 mb-2">
+          <p class="text-body-xs text-foreground">
+            You already have a valid SSO session for this workspace.
+          </p>
+        </div>
+
+        <!-- SSO Login Button -->
```
Comment (Mikehrn) [inline packages/frontend-2/pages/workspaces/[slug]/sso/index.vue:37]: imo this is already quite clear without the comments

packages/frontend-2/pages/workspaces/[slug]/sso/index.vue
```diff
@@ -1,50 +1,117 @@
...
+  // Convert URL-friendly error to readable message
+  return decodeURIComponent(error).replace(/\+/g, ' ').trim()
+})
+
+const isAuthenticating = computed(() => {
+  return !!route.query.access_code
+})
 
 const handleContinue = () => {
+  mixpanel.track('Workspace SSO Login Attempt', {
```
Comment (Mikehrn) [inline packages/frontend-2/pages/workspaces/[slug]/sso/index.vue:106]: Same here, -> `Workspace SSO Login Attempted`

packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue
```diff
@@ -0,0 +1,63 @@
...
+    </div>
+    <template v-else>
+      <div v-if="workspace?.logo" class="w-16 h-16">
+        <img
+          :src="workspace.logo"
+          :alt="workspace?.name"
+          class="w-full h-full object-contain rounded-full"
+        />
+      </div>
+      <h1 class="text-heading-xl">SSO Required for {{ workspace?.name }}</h1>
```
Comment (Mikehrn) [inline packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue:14]: `SSO Required` -> `SSO is required`

packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue
```diff
@@ -0,0 +1,63 @@
...
+
+const workspaceSlug = route.params.slug as string
+const { workspace, loading, error } = useWorkspaceSsoPublic(workspaceSlug)
+
+if (error.value) {
+  logger.error('Failed to fetch workspace data:', error.value)
+}
+
+const handleSsoLogin = () => {
+  mixpanel.track('Workspace SSO Session Error Redirect', {
```
Comment (Mikehrn) [inline packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue:51]: `Workspace SSO Session Error Redirected`

Mikehrn (changes requested):

packages/frontend-2/components/settings/workspaces/security/sso/Wrapper.vue
andrewwallacespeckle: I've removed all edit functionality in this PR. It's not possible in backend yet, and I'll need to discuss with Chuck how to handle it.

packages/frontend-2/lib/workspaces/composables/management.ts
cdriesler: I think this is (should be) the only one left that needs to be converted to gql. Predates your flag, but will happen. 👍

packages/frontend-2/components/auth/sso/Login.vue
```diff
@@ -0,0 +1,187 @@
...
+        show-label
+      >
+        <template #option="{ item }">
+          <div class="flex items-center gap-2">
+            <NuxtImg v-if="item.logo" :src="item.logo" class="h-6 w-6 rounded-full" />
+            <span>{{ item.name }}</span>
+          </div>
+        </template>
+        <template #something-selected="{ value }">
+          <div v-if="isWorkspace(value)" class="flex items-center gap-2">
```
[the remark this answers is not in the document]
andrewwallacespeckle: I think the type guard is needed because the component doesn't properly type the value based on multiple: false, so TypeScript thinks it could still be an array.

packages/frontend-2/components/auth/third-party/LoginButtonBase.vue
andrewwallacespeckle: I tried this, but it broke the functionality of these buttons. I can't test this very easily, and it's out of scope for this ticket, so I kept it as it was and just used the FormButton for the styles.

packages/frontend-2/components/settings/workspaces/security/sso/DeleteDialog.vue
andrewwallacespeckle: The toasts are triggered in useWorkspaceSsoDelete

packages/frontend-2/middleware/requireValidProject.ts
andrewwallacespeckle: No enum existed from the backend, but I've added one in types.ts that covers all the SSO error codes we receive from the API.

packages/frontend-2/components/auth/sso/Login.vue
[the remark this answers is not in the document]
fabis94: @andrewwallacespeckle That's what the `isArrayValue`/`isMultiitemArrayValue` utils off `useFormSelectChildInternals()` are for

packages/frontend-2/components/auth/third-party/LoginButtonBase.vue
fabis94: @andrewwallacespeckle The bigger issue here is that you're breaking HTML spec by nesting a button inside of an anchor, this should cause a hydration mismatch

andrewwallacespeckle: There was no hydration issue, but it does break HTML spec. 

I got this working using the `linkComponent` prop of FormButton 👍

packages/frontend-2/components/auth/third-party/LoginButtonBase.vue
```diff
@@ -1,21 +1,22 @@
...
+    full-width
+    class="group"
     :to="to"
-    :class="[
-      'transition grow px-3 inline-flex justify-center items-center outline-none h6 font-medium leading-7',
-      'rounded shadow bg-foundation- text-foreground dark:text-foreground-on-primary',
-      'border border-outline-2 hover:border-outline-3',
-      noVerticalPadding ? '' : 'py-2'
-    ]"
+    :link-component="NuxtLink"
```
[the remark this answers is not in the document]
andrewwallacespeckle: I don't think this is nested though. We have a FormButton that will render to a NuxtLink, but there is no parent button in the parent component or anything. Really all this change is, is to remove the style definition from class, and swap it for the styled buttons we use everywhere else.

packages/frontend-2/lib/workspaces/helpers/types.ts
```diff
@@ -0,0 +1,26 @@
+export type SsoFormValues = {
+  providerName: string
+  clientId: string
+  clientSecret: string
+  issuerUrl: string
+}
+
+export type WorkspaceSsoProviderPublic = {
```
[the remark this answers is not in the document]
andrewwallacespeckle: This one is for the REST API - should change in future

packages/frontend-2/middleware/requireSsoEnabled.ts
```diff
@@ -0,0 +1,19 @@
+import { workspaceRoute } from '~/lib/common/helpers/route'
+import { useWorkspacePublicSsoCheck } from '~/lib/workspaces/composables/sso'
+
+/**
+ * Used to validate that the workspace has SSO enabled, redirects to workspace page if not
+ */
+export default defineNuxtRouteMiddleware(async (to) => {
+  if (to.query.access_code) {
```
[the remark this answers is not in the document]
andrewwallacespeckle: The page has dual purposes - it's both a login page and the OAuth callback URL. Without this check, the middleware would redirect before we can process the access code from the SSO provider.

Let me know if you see a better way to handle this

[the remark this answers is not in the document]
fabis94: ok that's fine, it would be nice to have this as a comment in code too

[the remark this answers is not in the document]
andrewwallacespeckle: I've also added a comment here explaining this.

packages/frontend-2/lib/workspaces/composables/sso.ts
```diff
@@ -0,0 +1,181 @@
...
+    hasSsoEnabled
+  }
+}
+
+/**
+ * Checks if a workspace requires SSO authentication and the current user's SSO status.
+ * Used to enforce SSO login requirements for workspace access.
+ */
+export const useWorkspaceSsoStatus = (params: { workspaceSlug: Ref<string> }) => {
+  graphql(`
```
fabis94, line 51: never thought of putting these constants inside a function body, I suppose that does make it cleaner/more encapsulated 👍

packages/frontend-2/lib/workspaces/composables/sso.ts
```diff
@@ -0,0 +1,181 @@
...
+      expiredSsoSessions {
+        id
+        slug
+      }
+    }
+  `)
+
+  const { result, loading, error } = useQuery<WorkspaceSsoCheckQuery>(
+    workspaceSsoCheckQuery,
+    { slug: params.workspaceSlug }
```
fabis94, line 76: i'm not sure if this is reactive, seems like its a non-reactive object w/ the slug being a ref. maybe vue-apollo is smart enough to unwrap that, but I'd do the entire variables object as a ref: `computed(() => ({slug: params.workspaceSlug.value))

packages/frontend-2/pages/workspaces/[slug]/sso/index.vue
```diff
@@ -1,50 +1,127 @@
...
+const workspaceSlug = computed(() => route.params.slug as string)
+const { isSsoAuthenticated } = useWorkspaceSsoStatus({
+  workspaceSlug: computed(() => workspaceSlug.value)
+})
+
+const { result } = useQuery(authRegisterPanelQuery, {
+  token: route.query.token as string
+})
+
+const { workspace, loading, error } = useWorkspacePublicSsoCheck(workspaceSlug.value)
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[slug]/sso/index.vue:82]: not reactive, you're passing in a non-reactive string

packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue
```diff
@@ -0,0 +1,62 @@
...
+import { useMixpanel } from '~/lib/core/composables/mp'
+import { useAuthManager, useLoginOrRegisterUtils } from '~/lib/auth/composables/auth'
+
+const route = useRoute()
+const logger = useLogger()
+const mixpanel = useMixpanel()
+const { signInOrSignUpWithSso } = useAuthManager()
+const { challenge } = useLoginOrRegisterUtils()
+
+const workspaceSlug = route.params.slug as string
```
Comment (fabis94) [inline packages/frontend-2/pages/workspaces/[slug]/sso/session-error.vue:42]: non-reactive, wrap this with a computed

fabis94 (changes requested):
some remaining comments, besides the ones i left in DMs

packages/frontend-2/components/auth/sso/Login.vue
```diff
@@ -0,0 +1,187 @@
...
+
+const onEmailChange = (value: string) => {
+  email.value = value
+  emailCheckState.value = EmailCheckState.Idle
+  setFieldValue('workspace', undefined)
+  debouncedCheckEmail(value)
+}
+
+const onSubmit = handleSubmit((values: FormValues) => {
+  if (!values.workspace) return
```
fabis94, line 157: y not use a required rule? 🤔 ideally there shouldnt be any validation in this callback, vee-validate should've already handled it

fabis94 (dismissed):

packages/frontend-2/components/auth/sso/Login.vue
fabis94: oh and also - y the `values: FormValues`? `values` should already have the correct type from the useForm generic

fabis94: @andrewwallacespeckle well, yea, that's what you've typed the form values as. the field has the isRequired rule on it, so it doesn't appear that it can actually be undefined and you can update the type

fabis94 (approved):

fabis94 (approved):

Mikehrn (approved):

PR #3489, opened 2024-11-13
Chore(dui3): improve existing model card error

oguzhankoral:
image

image


didimitrie (approved):
it's not perfect, and it never will be - approved :D

PR #3494, opened 2024-11-14
fix(fe2): Only show "New Discussion" button when an object is selected

andrewwallacespeckle:
[LINEAR TICKET](https://linear.app/speckle/issue/WEB-2087/new-discussion-button-doesnt-do-anything)

Now, we only show the button when an object is selected. If not, we show a message telling them to select something.

### DIFF
--- a/packages/frontend-2/components/viewer/comments/Comments.vue
+++ b/packages/frontend-2/components/viewer/comments/Comments.vue
@@ -62,7 +62,10 @@
       <div v-if="commentThreads.length === 0" class="pb-4">
         <ProjectPageLatestItemsCommentsEmptyState
           small
-          :show-button="canPostComment"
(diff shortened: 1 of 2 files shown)


packages/frontend-2/components/viewer/comments/Comments.vue
```diff
@@ -62,7 +62,10 @@
       <div v-if="commentThreads.length === 0" class="pb-4">
         <ProjectPageLatestItemsCommentsEmptyState
           small
-          :show-button="canPostComment"
+          :show-button="canPostComment && hasSelectedObjects"
+          :text="
+            hasSelectedObjects ? undefined : 'Select an object to start collaborating'
```
andrewwallacespeckle, line 67: image
@benjaminvo feel free to update this copy

Mikehrn (approved):

PR #3503, opened 2024-11-15
fix(automate): drop specklecon function templates

cdriesler:
I miss u already SpeckleCon


gjedlicska (approved):

jsdbroughton:
Ahh I told a few people they'd be there for the coming week. Never mind.

PR #3515, opened 2024-11-19
fix(server): remove flaky dblink_slot_create_or_drop calls in tests

fabis94:
this keeps causing flaky multiregion test breakages https://app.circleci.com/pipelines/github/specklesystems/speckle-server/14136/workflows/adbe28b7-c2bc-4967-ba05-f3aee4d682a0/jobs/231690/tests

maybe we don't need it?

### DIFF
--- a/packages/server/test/hooks.ts
+++ b/packages/server/test/hooks.ts
@@ -151,9 +151,11 @@ export const resetPubSubFactory = (deps: { db: Knex }) => async () => {
     await deps.db.raw(
       `SELECT * FROM aiven_extras.pg_drop_subscription('${sub.subname}');`
     )
-    await deps.db.raw(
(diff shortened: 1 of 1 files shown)


gjedlicska:
Will the DB in test and dev envs have enough replication slots to not delete the subs?
@fabis94 @iainsproat ?

fabis94:
@gjedlicska we're still deleting the subs, we're just not explicitly deleting the slots after. it's not clear if that's even necessary, we added that while troubleshooting the original multiregion test issues, but there's been no signs so far that we actually need it - I mean you would think that deleting a subscription would also delete its slot.

If NOT, then the only alternative we have is to introduce delays/waits into the mix, which I did not want to do until we're sure this doesn't work

fabis94:
lets merge and monitor this first before we remove that line altogether

PR #3516, opened 2024-11-19
Fix: Billing fail/success toasts

Mikehrn:
- Trigger toasts on the correct location
- Fixes cancelCheckoutSessionMutation
- Remove SSO related toasts (will be re-added in follow up PR but are currently broken)


packages/frontend-2/lib/billing/composables/actions.ts
```diff
@@ -40,9 +45,35 @@ export const useBillingActions = () => {
     })
   }
 
+  const validateCheckoutSession = (workspaceId: string) => {
```
andrewwallacespeckle, line 48: In my SSO branch, Fabians told me to always make composables reactive by using refs/maybeRef

`this is why composables that accept a component's data through fn params should always accept a Ref or a MaybeRef so that it can react to changes to that value`

andrewwallacespeckle (approved):
1 minor comment, but all good apart from that. Feel free to ignore if you think it's unneccessary

andrewwallacespeckle (approved):

andrewwallacespeckle (approved):

PR #3523, opened 2024-11-20
feat(file import): model ID is passed to IFC parser

iainsproat:
## Description & motivation

Due to multi-region changes, we now need a model ID for some graphql calls.

The model ID, if it exists, is known by the file import service service. This should be passed to the downstream parsers.

This PR also aligns the command line arguments passed to all parsers to remain the same, even if the parser does not require the additional values at this time.

## Validation of changes:

 - tested when running yarn dev locally, IFC files imported ok.

## Checklist:

### DIFF
--- a/packages/fileimport-service/src/daemon.js
+++ b/packages/fileimport-service/src/daemon.js
@@ -155,7 +155,10 @@ async function doTask(mainDb, regionName, taskDb, task) {
           info.userId,
           info.streamId,
           info.branchName,
-          `File upload: ${info.fileName}`
+          `File upload: ${info.fileName}`,
+          info.id,
+          existingBranch?.id,
+          regionName
         ],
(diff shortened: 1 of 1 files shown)


adamhathcock (approved):

packages/fileimport-service/src/daemon.js
```diff
@@ -81,7 +81,8 @@ async function doTask(mainDb, regionName, taskDb, task) {
       fileSize: fileSizeForMetric,
       userId: info.userId,
       streamId: info.streamId,
-      branchName: info.branchName
+      branchName: info.branchName,
+      branchId: info.branchId
```
gjedlicska, line 85: this will always be undefined, the info does not have a branchId field, cause the DB does not have that column

PR #3531, opened 2024-11-20
fix(automate): automate module multi region

cdriesler:
## Description & motivation

- Move (almost all) automate module endpoints to multiregion logic

## Changes:

- rest, gql, and event listeners
- pending:
  - how to count automations a function is used in (requires counting in all regions) (or just deleting this capability?)
  - how to find project based on function run id (status reporting from automate)
    - I have some thoughts but would like to do this in a separate PR because it requires changes beyond a simple refactor


packages/server/modules/automate/rest/logStream.ts
```diff
@@ -21,27 +19,30 @@ import { Application } from 'express'
 
 export default (app: Application) => {
   app.get(
-    '/api/automate/automations/:automationId/runs/:runId/logs',
+    '/api/v1/projects/:projectId/automations/:automationId/runs/:runId/logs',
```
alemagio, line 22: is it fine breaking this API?

packages/server/package.json
```diff
@@ -22,7 +22,7 @@
     "build:clean": "rimraf ./dist && yarn build",
     "dev:clean": "yarn build:clean && yarn dev",
     "dev:server:test": "cross-env DISABLE_NOTIFICATIONS_CONSUMPTION=true NODE_ENV=test LOG_LEVEL=silent LOG_PRETTY=true node ./bin/ts-www",
-    "test": "cross-env NODE_ENV=test LOG_LEVEL=silent LOG_PRETTY=true mocha",
+    "test": "cross-env NODE_ENV=test LOG_PRETTY=true mocha",
```
alemagio, line 25: did you remove the LOG_LEVEL=silent?

alemagio (changes requested):

packages/server/modules/automate/rest/logStream.ts
gjedlicska: yes, its only the frontend, that uses this api

cdriesler: yes has been adjusted to reflect this 👍

packages/server/package.json
cdriesler: not on purpose tyty

cdriesler:
Ready for review after testing under multiregion config locally @alemagio @gjedlicska . One corner I'll raise during standup.

alemagio (dismissed):

PR #3537, opened 2024-11-21
fix(automate): support success object result level

cdriesler:
## Description & motivation

- We're adding a `SUCCESS` result level for objects and the frontend needs to support it. 

## Changes:

- Looks like this:

Screen Shot 2024-11-21 at 11 32 15


jsdbroughton:
Questionable use-case example in the image. Running state should not have a result at all.

cdriesler:
> Questionable use-case example in the image. Running state should not have a result at all.

100%, this is using the mocks just to make sure it shows up as expected

Mikehrn (approved):

PR #3547, opened 2024-11-25
Feat: Disable SSO if not in plan

Mikehrn:
Screenshot 2024-11-25 at 13 51 23


packages/frontend-2/components/settings/workspaces/security/sso/Wrapper.vue
```diff
@@ -126,6 +131,7 @@ graphql(`
         issuerUrl
       }
     }
+    hasAccessToSSO: hasAccessToFeature(featureName: OidcSso)
```
fabis94, line 134: As we do more of these we should turn these into fragments so that they always have the same prefix and thus rely on the same value in the cache. Currently there's only this and my "hasMultiRegionAccess" one afaik, but we should keep this in mind as we keep adding these fields.

fabis94 (approved):

packages/frontend-2/components/settings/workspaces/Security.vue
```diff
@@ -318,7 +318,8 @@ const addDomain = async () => {
     },
     result.value?.workspace.domains ?? [],
     result.value?.workspace.discoverabilityEnabled,
-    result.value?.workspace.domainBasedMembershipProtectionEnabled
+    result.value?.workspace.domainBasedMembershipProtectionEnabled,
```
fabis94, line 321: https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#c79ca97569504d1dbd9d5d3b9978360a

packages/frontend-2/lib/settings/composables/management.ts
```diff
@@ -89,7 +90,8 @@ export function useAddWorkspaceDomain() {
                 ],
                 discoverabilityEnabled:
                   domains.length === 0 ? true : discoverabilityEnabled,
-                domainBasedMembershipProtectionEnabled
+                domainBasedMembershipProtectionEnabled,
```
fabis94, line 93: This whole optimisticResponse block looks very hacky, with the `as Workspace` bit and readFragment(). Why do we need an optimisticResponse here at all?

fabis94 (approved):
there's some messy bits that should be improved on after

packages/frontend-2/lib/settings/composables/management.ts
Mikehrn: @fabis94 I think there were two parts of this, one part was added because in the UI we needed to reflect the list of updated domains, but it was not available, the other was added later on during SSO but I didn't work it that. You have any suggestions for a better approach?

fabis94: @Mikehrn Its still not clear to me why optimisticResponse was needed. All it does is return a fake response before the real one comes back. But surely the real one comes back in very short time like all other GQL mutations, so why not just use that?

optimisticResponse is used for things like like buttons where you want the button to toggle immediately, instead of having to wait for the API call to finish

PR #3550, opened 2024-11-25
Fix: Clean up some more automate styling

Mikehrn:
Bit more cleaning up of automate UI. Again, can be improved further, but just a start to align things a bit more with the other styles

Screenshot 2024-11-25 at 20 29 56

Screenshot 2024-11-25 at 20 29 41

Screenshot 2024-11-25 at 20 27 41

Screenshot 2024-11-25 at 20 15 47

### DIFF
--- a/packages/frontend-2/components/automate/function/page/ParametersDialog.vue
+++ b/packages/frontend-2/components/automate/function/page/ParametersDialog.vue
@@ -1,19 +1,19 @@
 <template>
-  <LayoutDialog v-model:open="open" max-width="md" title="Function Parameters">
+  <LayoutDialog
+    v-model:open="open"
+    :buttons="dialogButtons"
+    max-width="md"
+    title="Function parameters"
+  >
     <div class="flex flex-col space-y-4">
       <template v-if="finalParams">
         <FormJsonForm :schema="finalParams" />
-        <LayoutDialogSection
-          title="Parameter schema"
-          :icon="BeakerIcon"
-          border-t
-          border-b
-        >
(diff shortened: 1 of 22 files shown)


packages/frontend-2/components/automate/function/page/ParametersDialog.vue
```diff
@@ -1,19 +1,19 @@
...
-          border-t
-          border-b
-        >
+        <LayoutDialogSection title="Parameter schema" border-t border-b>
           <FormTextArea
             name="actionYaml"
             readonly
             :model-value="JSON.stringify(finalParams, null, 2)"
-            class="text-sm text-primary"
+            textarea-classes="!bg-foundation !border border-outline-2 p-2 rounded-lg font-mono !font-body-2xs"
```
benjaminvo, line 16: I haven't seen `textarea-classes` before 🤔. Is this a real prop?

benjaminvo (approved):
All looks very good! One comment – but I've approved anyway in case it is as it should be. Otherwise fix and I'll reapprove.

packages/frontend-2/components/automate/function/page/ParametersDialog.vue
Mikehrn: Yup it is, it's used only once or twice I think, but inside `FormTextArea` is a nested `textarea` element, this prop allows you to set the classes on that element :)

benjaminvo: Ah!

PR #3566, opened 2024-11-27
feat(fe2): Update isSsoAuthenticated composable

andrewwallacespeckle:
Update composable to check if sso session is still valid.


packages/server/assets/workspacesCore/typedefs/workspaces.graphql
```diff
@@ -299,6 +299,7 @@ type WorkspaceSsoProvider {
 }
 
 type WorkspaceSsoSession {
+  id: ID!
```
andrewwallacespeckle, line 302: @cdriesler Is this change ok? workspaceBySlug wasn't returning the id, which causes some apollo cache warnings. Am I safe returning ID here?

Mikehrn (approved):

PR #3567, opened 2024-11-27
fix(fe2): workspace region settings broken if non-admin + Workspace.subscription errors

fabis94:


packages/frontend-2/lib/core/configs/apollo.ts
```diff
@@ -334,7 +334,7 @@ function createWsClient(params: {
 
 const coreShouldSkipLoggingErrors = (err: ErrorResponse): boolean => {
   // These fields have special auth requirements and will often throw errors that we don't want to log
-  const specialAuthFields = ['invitedTeam', 'billing', 'domains']
+  const specialAuthFields = ['invitedTeam', 'billing', 'domains', 'subscription']
```
fabis94, line 337: @Mikehrn We have errorPolicy: 'all' so its fine that the Workspace.subscription field errors out and returns null for non-admins, but we should also remember to update this array (or the skipLoggingErrors param of the specific query) to ensure we don't log these as errors in the console and scare devs (which is how I was told to investigate this hehe)

packages/server/modules/gatekeeper/graph/resolvers/index.ts
```diff
@@ -37,24 +37,12 @@ export = FF_GATEKEEPER_MODULE_ENABLED
         plan: async (parent) => {
           return await getWorkspacePlanFactory({ db })({ workspaceId: parent.id })
         },
-        subscription: async (parent, _, ctx) => {
+        subscription: async (parent) => {
```
fabis94, line 40: just moved these over to directives

packages/server/assets/gatekeeper/typedefs/gatekeeper.graphql
```diff
@@ -81,10 +81,15 @@ extend type Workspace {
...
+  subscription: WorkspaceSubscription
+    @hasScope(scope: "workspace:billing")
+    @hasWorkspaceRole(role: ADMIN)
   # this can only be created if there is an active subscription
-  customerPortalUrl: String @hasScope(scope: "workspace:billing")
+  customerPortalUrl: String
+    @hasScope(scope: "workspace:billing")
+    @hasWorkspaceRole(role: ADMIN)
   hasAccessToFeature(featureName: WorkspaceFeatureName!): Boolean!
+    @hasWorkspaceRole(role: GUEST)
```
fabis94, line 92: @gjedlicska downgraded this to allow guest access, cause this is kind of a core part of the frontend that resolves which pages to load and how. is that ok?

Mikehrn (dismissed):

Mikehrn (approved):

Mikehrn (approved):

PR #3568, opened 2024-11-27
fix(fe2): Update delete sso provider composable

andrewwallacespeckle:
Updated delete sso provider mutation to use the gql mutation.
Also used the Badge component for the sso status. 

Locally, all my workspaces are trials. I haven't been able to test this, as backend throws an error when you dont have the correct billing. Will likely be a follow up to this branch, I just need to get it on latest to test it properly.


packages/frontend-2/components/settings/workspaces/security/sso/Wrapper.vue
```diff
@@ -25,22 +25,29 @@
         <FormButton v-else @click="goToBilling">Upgrade to Plus</FormButton>
       </div>
 
+      <div
```
Mikehrn, line 28: You can use `CommonCard`, if it's just for now for testing purposes no need to change it

packages/frontend-2/components/settings/workspaces/security/sso/Wrapper.vue
```diff
@@ -16,7 +16,7 @@
         </div>
         <FormButton
           v-if="workspace.hasAccessToSSO"
-          :disabled="isFormVisible || !!provider"
+          :disabled="!isWorkspaceAdmin || isFormVisible || !!provider"
```
Mikehrn, line 19: Probably should add a tooltip here, something like "Ask your workspace admin to enable SSO" or something like that

Mikehrn (changes requested):

Mikehrn (approved):

PR #3579, opened 2024-11-28
fix(sso): include both redirects

cdriesler:
Gotta include both:

Screen Shot 2024-11-28 at 17 35 52


Mikehrn (approved):
I think you just need to merge back main and and regen gql

cdriesler:
jk

PR #3583, opened 2024-11-29
Feat: Disable WS delete on active plans

Mikehrn:


packages/frontend-2/components/billing/Alert.vue
```diff
@@ -1,11 +1,14 @@
+<!-- eslint-disable vuejs-accessibility/heading-has-content -->
```
Mikehrn, line 1: There is actually content in the titel, but because it's in a nested component linter doesnt see it

andrewwallacespeckle (dismissed):

cdriesler (approved):

PR #3587, opened 2024-12-01
chore(automate): drop specklecon templates

cdriesler:
## Description & motivation

- Bringing templates back to just C#/python in preparation for public beta
- @jsdbroughton requesting your thumbs up that now is a good time to do so


jsdbroughton (approved):
its time!

PR #3589, opened 2024-12-01
Feat: disable settings access if active SSO session is required

Mikehrn:


packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -228,6 +236,11 @@ const workspaceMenuItemClasses = (
   targetWorkspaceId.value === workspaceId &&
   !disabled
 
+const needsSsoSession = (workspace: SettingsMenu_WorkspaceFragment, key: string) => {
```
Mikehrn, line 239: Not perfect, but this can be improved when moving to pages

andrewwallacespeckle (approved):

PR #3596, opened 2024-12-02
Fix: Billing switch

Mikehrn:
Only enable billing switch if:
In trial period
OR can upgrade to plan
OR (matches current plan AND the active billing interval is monthly)

@benjaminvo Could you double-check if this is how you would like it to work?


benjaminvo (approved):
It is!

PR #3615, opened 2024-12-03
feat(fe2): Billing - Add tooltips to disabled buttons

andrewwallacespeckle:
Let the user know why certain actions are disabled in Settings>Workspace>Billing.

I created some new computed for some of these conditionals and used them where appropriate.

### DIFF
--- a/packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
+++ b/packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
@@ -31,8 +31,9 @@
         20% off
       </CommonBadge>
     </div>
-    <div v-if="workspaceId" v-tippy="buttonTooltip" class="w-full mt-4">
+    <div v-if="workspaceId" class="w-full mt-4">
       <FormButton
+        v-if="isMonthlyToAnnual"
         :color="buttonColor"
         :disabled="!isSelectable"
         full-width
@@ -40,6 +41,16 @@
       >
         {{ buttonText }}
       </FormButton>
+      <div v-else v-tippy="buttonTooltip">
+        <FormButton
+          :color="buttonColor"
+          :disabled="!isSelectable"
+          full-width
+          @click="onCtaClick"
+        >
+          {{ buttonText }}
+        </FormButton>
+      </div>
     </div>
     <ul class="flex flex-col gap-y-2 mt-4 pt-3 border-t border-outline-3">
       <li
@@ -137,6 +148,12 @@ const statusIsTrial = computed(
   () => props.currentPlan?.status === WorkspacePlanStatuses.Trial || !props.currentPlan
 )
 
+const isMatchingInterval = computed(
+  () =>
+    props.activeBillingInterval ===
+    (props.yearlyIntervalSelected ? BillingInterval.Yearly : BillingInterval.Monthly)
+)
+
 const isDowngrade = computed(() => {
   return !canUpgradeToPlan.value && props.currentPlan?.name !== props.plan.name
 })
@@ -156,12 +173,6 @@ const isMonthlyToAnnual = computed(() => {
     props.yearlyIntervalSelected
   )
 })
-
-const isMatchingInterval = computed(
-  () =>
-    props.activeBillingInterval ===
-    (props.yearlyIntervalSelected ? BillingInterval.Yearly : BillingInterval.Monthly)
-)
 
 const isCurrentPlan = computed(
   () => isMatchingInterval.value && props.currentPlan?.name === props.plan.name
--- a/packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
+++ b/packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
@@ -33,7 +33,7 @@
     </div>
     <div v-if="workspaceId" class="w-full mt-4">
       <FormButton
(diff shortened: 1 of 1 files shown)


packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
```diff
@@ -34,13 +33,24 @@
     </div>
     <div v-if="workspaceId" class="w-full mt-4">
       <FormButton
+        v-if="isMonthlyToAnnual || canUpgradeToPlan"
```
andrewwallacespeckle, line 36: Had to do this hacky fix to the v-tippy reactivity bug

benjaminvo (changes requested):
Tested and found various problems. Don't mind the slightly transparent tooltips in the images – my screenshotting tool is a bit glitchy.

1. This tooltip appears on the starter plan while you're on a trial:

CleanShot 2024-12-03 at 16 00 47@2x

2. The downgrading tooltip appears on the starter plan while the button is in the "current plan" state
CleanShot 2024-12-03 at 16 01 39@2x

3. No tooltips on the disabled Plus and Business plan – in this case because I'm on an annual Starter plan and trying to upgrade to monthly Plus or Business plan. 
CleanShot 2024-12-03 at 16 02 41@2x

4. The annual->monthly tooltip appears on the current plan here after I toggle the billing switch on and off.
CleanShot 2024-12-03 at 16 04 04@2x

packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
```diff
@@ -172,36 +200,50 @@ const isSelectable = computed(() => {
   // Allow upgrades to higher tier plans
   return canUpgradeToPlan.value
 })
-const toggleEnabled = computed(() => {
```
Mikehrn, line 175: There is a bit of a difference here, if there current interval is monthly, it should be enable (even if the selected state is yearly)

packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
```diff
@@ -36,15 +35,35 @@
     </div>
     <div v-if="workspaceId || hasCta" class="w-full mt-4">
       <slot name="cta" />
+      <div v-if="!isAdmin" v-tippy="`You must be a workspace admin`">
```
Mikehrn, line 38: - It renders both on the wizard, maybe do `!hasCta` on the `<div>`
- Because the div checks `v-if="!isAdmin"` it will not render at all so also no tooltip, you probably need `<div v-if="!hasCta" v-tippy="isAdmin ? undefined : You must be a workspace admin">` and then we can get rid of the button duplication

Screenshot 2024-12-05 at 16 35 27

Mikehrn (changes requested):

packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
andrewwallacespeckle: But there is an v-else-if below, so this will just show the other button if there is no cta defined. 

I also can't do the v-tippy ternary because that causes the reactivity bug. But the current logic makes sense, the button with the tooltip will only be shown to non-admins. Telling them why its disabled. 

I wrapped the slot in a v-if for hasCta, then did a v-else-if="!isAdmin", so this should solve the button appearing in the Wizard. 

I think this works in every scenario, but let me know if I missed something you can see if your local dev .

Mikehrn (approved):

andrewwallacespeckle:
@benjaminvo This one has been approved, but I can't merge until you also approve. Thanks

benjaminvo (approved):

benjaminvo:
I just tested this on Latest. I'm still experiencing problem 3 and 4 from [this message](https://github.com/specklesystems/speckle-server/pull/3615#pullrequestreview-2475986281) 🤔. No tooltips on the disabled upgrade buttons (3) and a wrong tooltip on the "current plan" button after switching the billing switch on and off (4).

andrewwallacespeckle:
@B

> I just tested this on Latest. I'm still experiencing problem 3 and 4 from [this message](https://github.com/specklesystems/speckle-server/pull/3615#pullrequestreview-2475986281) 🤔. No tooltips on the disabled upgrade buttons (3) and a wrong tooltip on the "current plan" button after switching the billing switch on and off (4).

I have just opened a PR for point 3 [here](https://github.com/specklesystems/speckle-server/pull/3650).

I haven't been able to recreate point 4, was there anything else you did to get it? 🤔

benjaminvo:
> @B
> 
> 
> 
> > I just tested this on Latest. I'm still experiencing problem 3 and 4 from [this message](https://github.com/specklesystems/speckle-server/pull/3615#pullrequestreview-2475986281) 🤔. No tooltips on the disabled upgrade buttons (3) and a wrong tooltip on the "current plan" button after switching the billing switch on and off (4).
> 
> 
> 
> I have just opened a PR for point 3 [here](https://github.com/specklesystems/speckle-server/pull/3650).
> 
> 
> 
> I haven't been able to recreate point 4, was there anything else you did to get it? 🤔

Not really I think. It's whenever my current plan is a yearly plan -> I switch to monthly billing -> hover the disabled "change to monthly" button -> switch back to yearly billing -> hover the disabled "current plan" button.

PR #3618, opened 2024-12-03
fix(fe2): Allow server:admin access to all workspace settings

andrewwallacespeckle:
Bug reported by @iainsproat in discord:

as a server admin, I'm able to see all workspace settings. (expected?)
The tooltip and menu on the left suggest that I need to sign in with SSO (bug)
I'm able to click on the greyed-out link anyway and it opens the page (expected?)

image

### DIFF
--- a/packages/frontend-2/components/settings/Dialog.vue
+++ b/packages/frontend-2/components/settings/Dialog.vue
@@ -84,10 +84,13 @@
                   "
                   :disabled="!isAdmin && (workspaceMenuItem.disabled || needsSsoSession(workspaceItem, itemKey as string))"
                   extra-padding
-                  @click="() => {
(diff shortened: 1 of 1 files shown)


Mikehrn (approved):

packages/frontend-2/components/settings/Dialog.vue
```diff
@@ -82,15 +82,14 @@
                       ? 'Log in with your SSO provider to access this page'
                       : workspaceMenuItem.tooltipText
                   "
-                  :disabled="
-                    workspaceMenuItem.disabled || needsSsoSession(workspaceItem, itemKey as string)
-                  "
+                  :disabled="!isAdmin && (workspaceMenuItem.disabled || needsSsoSession(workspaceItem, itemKey as string))"
```
andrewwallacespeckle, line 85: Don't show disabled state when use is server:admin

PR #3621, opened 2024-12-03
fix(sso): disable slug edit if sso enabled

cdriesler:
## Description & motivation

- SSO is currently configured with redirects specific to a workspace slug. If the slug changes, SSO breaks. If SSO breaks and your session expires, you're locked out.

## Changes:

- Disable slug edit on FE & BE if SSO is configured

image


andrewwallacespeckle (approved):
Frontend looks good

PR #3627, opened 2024-12-04
fix(automate): cascade delete automations with projects

cdriesler:
## Description & motivation

- Projects with automations couldn't be deleted because of fk constraints
- It was not enough to mark `automationRevisionId` as a cascade delete because of its use in a composite primary key

## Changes:

- Migrate `automation_revision_functions` to have its own id
- Populate it randomly so we don't have to think about it / change repo access
- Add automation count in delete dialog, as a treat


packages/server/modules/automate/migrations/20241203212110_cascade_delete_automations.ts
```diff
@@ -0,0 +1,27 @@
+import { Knex } from 'knex'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable('automation_revision_functions', (table) => {
```
fabis94, line 4: make sure u try these out locally (both up & down), dropping and re-creating a foreign key within the same alterTable() scope could be problematic

fabis94 (approved):

packages/frontend-2/components/project/page/settings/general/block/Delete/Delete.vue
```diff
@@ -39,6 +39,9 @@ graphql(`
     workspace {
       slug
     }
+    automations {
```
fabis94, line 42: set limit:0, if possible, we sometimes use that to shortcut query execution on the backend

packages/server/modules/automate/migrations/20241203212110_cascade_delete_automations.ts
cdriesler: can confirm it appears to behave as expected 👍

fabis94 (approved):

PR #3628, opened 2024-12-04
Update wizard copy

benjaminvo:
- Update copy on the wizard steps
- Also append the slug URL to the helper text (including in settings)

CleanShot 2024-12-04 at 14 39 59@2x

### DIFF
--- a/packages/frontend-2/components/settings/workspaces/General.vue
+++ b/packages/frontend-2/components/settings/workspaces/General.vue
@@ -307,7 +307,7 @@ onResult((res) => {
 const baseUrl = config.public.baseUrl
 
 const slugHelp = computed(() => {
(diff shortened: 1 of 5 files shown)


packages/frontend-2/components/settings/workspaces/General.vue
```diff
@@ -307,7 +307,7 @@ onResult((res) => {
 const baseUrl = config.public.baseUrl
 
 const slugHelp = computed(() => {
-  return `Used after ${baseUrl}/workspaces/`
+  return `${baseUrl}/workspaces/${slug.value}`
```
benjaminvo, line 310: I decided to not include "Preview:" here because of the space.

Mikehrn (approved):

PR #3633, opened 2024-12-04
chore(automate): track redirects from beta site

cdriesler:
## Description & motivation

- We recently started redirecting all traffic from `automate.speckle.dev` to `app.speckle.systems/functions?automateBetaRedirect=true`
- This page is not ideal, it gives no further context, and it's not clear what the best redirect experience should be. But I want to get a clearer picture of how many people are actually affected before investing any time.

## Changes:

- Tracks a mixpanel event when people are redirected from `automate.speckle.dev`

### DIFF
--- a/packages/frontend-2/components/automate/functions/page/Header.vue
+++ b/packages/frontend-2/components/automate/functions/page/Header.vue
@@ -133,8 +133,10 @@ if (import.meta.client) {
     (isRedirect) => {
       if (!isRedirect?.length) return
       mixpanel.track('Automate Beta Visit Redirected')
-      void router.replace({ query: {} })
-    }
+      const { automateBetaRedirect, ...query } = route.query
+      void router.replace({ query })
+    },
+    { immediate: true }
   )
 }
 </script>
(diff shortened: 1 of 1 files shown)


packages/frontend-2/components/automate/functions/page/Header.vue
```diff
@@ -126,5 +128,12 @@ if (import.meta.client) {
     },
     { immediate: true }
   )
+  watch(
+    () => route.query['automateBetaRedirect'] as Nullable<string>,
+    (isRedirect) => {
+      if (!isRedirect?.length) return
+      mixpanel.track('Automate Beta Visit Redirected')
```
fabis94, line 135: you should always wipe these tracking query params after u process them otherwise you risk processing them multiple times

fabis94 (approved):
would be better to just wipe that single query param, you don't know what other kind of stuff is in the querystring and whether its useful for some other part of the app like utm tracking...

that said, its probably fine

fabis94 (approved):

PR #3635, opened 2024-12-05
feat(fe2): scope comment mentions to parent project collaborators

fabis94:
+ various related fixes


packages/server/assets/core/typedefs/user.graphql
```diff
@@ -42,7 +42,14 @@ extend type Query {
     cursor: String
     archived: Boolean = false
     emailOnly: Boolean = false
-  ): UserSearchResultCollection!
+  ): UserSearchResultCollection! @deprecated(reason: "Use users() instead.")
+
+  """
+  Look up server users
+  """
+  users(input: UsersRetrievalInput!): UserSearchResultCollection!
```
fabis94, line 50: The param list for `userSearch` was already too long and I expect it to grow in the future. Plus, why would we want to allow (non-admin)users to search for archived users?

packages/server/modules/notifications/services/handlers/mentionedInComment.ts
```diff
@@ -151,7 +151,7 @@ function buildEmailTemplateParams(
     objectId,
     commitId
   })
-  const url = new URL(commentRoute, getBaseUrl()).toString()
+  const url = new URL(commentRoute, getFrontendOrigin()).toString()
```
fabis94, line 154: mention email cta links were busted locally (only where FE and BE origin differs afaik)

packages/server/modules/core/repositories/users.ts
```diff
@@ -449,39 +450,84 @@ export const getUserRoleFactory =
...
-          prefixedLimitedUserFields.includes(col)
+        ...Object.values(
+          omit(Users.col, [
+            Users.col.email,
+            Users.col.verified,
+            Users.col.passwordDigest
+          ])
         ),
-        knex.raw(`(array_agg("user_emails"."verified"))[1] as verified`)
+        knex.raw(`(array_agg(??))[1] as "verified"`, [UserEmails.col.verified]),
```
fabis94, line 480: @alemagio it was this way before already, but i'm wondering if this logic is valid - surely we should look for the first verified email, instead of just the first email that exists? otherwise we risk marking the user as unverified, when one of their emails is verified

Mikehrn (dismissed):
FE looks good :)

Mikehrn (approved):

PR #3644, opened 2024-12-05
Trigger mention dialog after a single character

benjaminvo:
Follow up to #3635.

Now that you can only mention project collaborators I think it's safe to show the mention dropdown after typing a single character. Better UX.

### DIFF
--- a/packages/server/modules/core/tests/integration/users.graph.spec.ts
+++ b/packages/server/modules/core/tests/integration/users.graph.spec.ts
@@ -201,7 +201,7 @@ describe('Users @graphql', () => {
 
     it('doesnt work with less than 1 character', async () => {
       const res = await search({
-        query: 'fi'
+        query: ''
       })
       expect(res).to.haveGraphQLErrors('Search query must be at least 1 character')
     })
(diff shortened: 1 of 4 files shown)


fabis94 (approved):

packages/server/modules/core/tests/integration/users.graph.spec.ts
```diff
@@ -199,11 +199,11 @@ describe('Users @graphql', () => {
       ])
     })
 
-    it('doesnt work with less than 3 characters', async () => {
+    it('doesnt work with less than 1 character', async () => {
       const res = await search({
         query: 'fi'
```
fabis94, line 204: this isnt right, the query should be invalid, so less than 1 character

fabis94 (changes requested):

packages/server/modules/core/tests/integration/users.graph.spec.ts
benjaminvo: Ah yes, adjusted

fabis94 (approved):

PR #3649, opened 2024-12-05
Feat: Send webhook for upgrade event

Mikehrn:


packages/frontend-2/lib/billing/composables/actions.ts
```diff
@@ -174,6 +203,18 @@ export const useBillingActions = () => {
           type: ToastNotificationType.Success,
           title: 'Your workspace plan was successfully updated'
         })
+
+        if (import.meta.server) {
+          await sendWebhook(defaultZapierWebhookUrl, {
+            workspaceId: workspace.id,
```
benjaminvo, line 209: Would be nice to also include Workspace name

Mikehrn: Good point, just completely forgot that one

andrewwallacespeckle (approved):

PR #3650, opened 2024-12-05
fix(fe2): Upgrading from annual to monthly billing tooltip

andrewwallacespeckle:
@benjaminvo This is one of those scenarios where this can't be tested locally, when you've used the CLI for billing. We are safe to push to on to app/latest though and check there.


benjaminvo (dismissed):

benjaminvo (approved):
Nice! Hope it works :)

PR #3654, opened 2024-12-09
Add Gendo Mixpanel event

benjaminvo:
Adds an MP event when a Gendo render is triggered. I also log the remaining renders left so we can track how many hit the limit or just try it a few times.


didimitrie (approved):
TY for doing my homework

PR #3659, opened 2024-12-09
Alessandro/web 2304 expose graphql field readonly for workspace type

alemagio:
## Checklist:


packages/server/modules/gatekeeper/domain/constants.ts
```diff
@@ -0,0 +1,7 @@
+import { z } from 'zod'
```
gjedlicska, line 1: You need to explain this to me a bit, why do we need this?

packages/server/modules/gatekeeper/services/readOnly.ts
```diff
@@ -0,0 +1,28 @@
...
+  workspacePlanStatusCanceled,
+  workspacePlanStatusExpired
+])
+
+export const isWorkspaceReadOnlyFactory =
+  ({ getWorkspacePlan }: { getWorkspacePlan: GetWorkspacePlan }) =>
+  async ({ workspaceId }: { workspaceId: Workspace['id'] }) => {
+    const workspacePlan = await getWorkspacePlan({ workspaceId })
+    // Should never happen
+    if (!workspacePlan) return true
```
gjedlicska, line 21: its prob safer to return false here

packages/server/modules/gatekeeper/services/readOnly.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    const { success: workspaceReadOnly } = readOnlyWorkspacePlanStatuses.safeParse(
```
gjedlicska, line 23: I like switch statements here better, it can give immediate typescript feedback if there are changes

gjedlicska (changes requested):

packages/server/modules/gatekeeper/services/readOnly.ts
```diff
@@ -0,0 +1,24 @@
...
+  async ({ workspaceId }: { workspaceId: Workspace['id'] }) => {
+    const workspacePlan = await getWorkspacePlan({ workspaceId })
+    // Should never happen
+    if (!workspacePlan) return true
+
+    switch (workspacePlan.status) {
+      case 'cancelationScheduled':
+      case 'valid':
+      case 'trial':
+      case 'paymentFailed':
```
alemagio, line 16: As stated [here](https://docs.stripe.com/billing/subscriptions/overview#failed-payments), stripe marks the subscription as `past_due` after one payment incomplete, which means we mark the workspace_plan as `paymentFailed`.
We might not want to block the workspace at this point but wait for stripe to follow its rules and retry.
After a certain amount of retries and failures the subscription will be marked as cancelled by stripe itself

packages/server/modules/gatekeeper/services/readOnly.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export const isWorkspaceReadOnlyFactory =
+  ({ getWorkspacePlan }: { getWorkspacePlan: GetWorkspacePlan }) =>
+  async ({ workspaceId }: { workspaceId: Workspace['id'] }) => {
+    const workspacePlan = await getWorkspacePlan({ workspaceId })
+    // Should never happen
+    if (!workspacePlan) return true
+
+    switch (workspacePlan.status) {
+      case 'cancelationScheduled':
```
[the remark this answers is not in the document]
alemagio: it is a typo, but it is already deep into the codebase, I would rather create a dedicated PR to fix it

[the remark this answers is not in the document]
Mikehrn: Yeah we should do one branch with BE + FE, because both use this enum so we will need to change them at the same time

PR #3676, opened 2024-12-11
Feat: Change server invite to new design (WIP)

Mikehrn:


packages/frontend-2/components/invite/dialog/Server.vue
```diff
@@ -0,0 +1,175 @@
...
+}
+
+const removeInviteItem = (index: number) => {
+  removeInvite(index)
+}
+
+const onSubmit = handleSubmit(() => {
+  const invites = fields.value.filter((invite) => invite.value.email)
+
+  invites.forEach(async (invite) => {
```
andrewwallacespeckle, line 138: Can you use async with a forEach? I think there is an issue with promises here

Mikehrn: I'm not awaiting anything here, they can all sent out an invite async, the modal closes in any case. This is because they can all have different values (role, project etc) so they need to be seperate invites

andrewwallacespeckle (approved):

PR #3686, opened 2024-12-13
Fix: Remove automations from General settings query

Mikehrn:
Its causing issues when the automate FF is off


benjaminvo (approved):

benjaminvo:
Approved, we can merge and follow up with a proper fix.

PR #3703, opened 2024-12-17
Set up email notifications for trial expirations

gjedlicska:


packages/server/modules/gatekeeper/index.ts
```diff
@@ -73,10 +82,59 @@ const scheduleWorkspaceTrialEmails = ({
...
-    // await manageSubscriptionDownscale()
+    const getWorkspacesByPlanAge = getWorkspacesByPlanAgeFactory({ db })
+    const trialValidForDays = 31
+    const trialWorkspacesExpireIn14Days = await getWorkspacesByPlanAge({
+      daysTillExpiry: 14,
+      planValidFor: trialValidForDays,
+      plan: 'starter',
+      status: 'trial'
+    })
+    if (trialWorkspacesExpireIn14Days.length) {
```
benjaminvo, line 105: @alemagio I think we can remove the 14 days email for now. I initially wanted it but now I think it's a little too much to send it this long time before the trial ends. 3 days before and on the same day is enough for now.

Maybe we will send another type of email 14 days before just to check in with the user, but that can be a separate task. Additionally we'll also be adding more automated emails _after_ the trial has expired to try and reactive the workspace. But also a separate task.

packages/server/modules/gatekeeper/services/trialEmails.ts
```diff
@@ -0,0 +1,108 @@
...
+  return { bodyStart }
+}
+
+const buildEmailTemplateParams = (
+  args: TrialExpiresArgsWithAdmin
+): EmailTemplateParams => {
+  const url = new URL(`workspaces/${args.workspace.slug}`, getServerOrigin()).toString()
+  return {
+    mjml: buildMjmlBody(args),
+    text: buildTextBody(),
```
benjaminvo, line 63: What's up with this? It looks like `buildTextBody()` above doesn't return anything.

alemagio: it is fine
bodyStart and bodyEnd in emails are optional

benjaminvo:
@alemagio Anything missing on your end to get this in? I'm not sure how to test that these work as intended.

alemagio:
> @alemagio Anything missing on your end to get this in? I'm not sure how to test that these work as intended.

I'm testing my changes

alemagio (approved):

PR #3713, opened 2024-12-18
Feat: Change blog to tutorials

Mikehrn:
Screenshot 2024-12-18 at 10 59 58

### DIFF
--- a/packages/frontend-2/components/dashboard/tutorials/Card.vue
+++ b/packages/frontend-2/components/dashboard/tutorials/Card.vue
@@ -1,5 +1,5 @@
 <template>
-  <NuxtLink :to="webflowItem.url" target="_blank">
+  <NuxtLink :to="webflowItem.url" target="_blank" @click="trackClick">
     <div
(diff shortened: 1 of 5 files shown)


packages/frontend-2/components/dashboard/tutorials/Card.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <NuxtLink :to="webflowItem.url" target="_blank">
+  <NuxtLink :to="webflowItem.url" target="_blank" @click="trackClick">
```
andrewwallacespeckle, line 2: Were you able to test this to make sure the `to` doesn't interrupt the tracking call?

andrewwallacespeckle (approved):

packages/frontend-2/components/dashboard/tutorials/Card.vue
Mikehrn: Yeah because it opens in a new tab it doesn't intervere, original tab stays open and emits the event :)

PR #3716, opened 2024-12-18
feat(fe2): New workspace page layout

andrewwallacespeckle:
image

[FIGMA FILE](https://www.figma.com/design/CFVYfzYSyUdTVnrlFUEcUZ/Onboarding?node-id=0-1&p=f&m=dev)


packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -1,5 +1,13 @@
 <template>
   <div>
+    <Portal to="right-sidebar">
+      <WorkspaceSidebar
```
Mikehrn, line 4: Needs a `w-full` somewhere 

Screenshot 2024-12-18 at 16 59 08

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -1,5 +1,13 @@
 <template>
```
Mikehrn, line 1: Because the content area is very narrow on some screens the empty states get very squeezed, probably increasing the breakpoint for grid-cols should fix this
Screenshot 2024-12-18 at 17 00 33

packages/frontend-2/components/workspace/ProjectList.vue
```diff
@@ -1,5 +1,13 @@
...
   <div>
```
Mikehrn, line 2: I dont know if it's my local and/or related, but search seems broken?

Screenshot 2024-12-18 at 17 03 28

packages/frontend-2/components/workspace/ProjectList.vue
Mikehrn, line 1: When logged in as member the actions in the sidebar still show, however members dont have permissions to access these actions

Screenshot 2024-12-18 at 17 05 50

Mikehrn, line 1: If you are guests actions show you are not allowed to take, such as editing security settings, moving/adding projects. Members are hidden but the toggle to collapse is still shown. Maybe remove the toggle when there is no content

Screenshot 2024-12-18 at 17 07 57

Mikehrn (changes requested):

packages/frontend-2/components/workspace/sidebar/Members.vue
```diff
@@ -0,0 +1,66 @@
...
+      v-if="!isWorkspaceGuest"
+      class="flex lg:flex-col items-center lg:items-start justify-between lg:justify-start gap-2 pb-0 lg:pb-4 mt-1"
+    >
+      <div class="flex items-center gap-1">
+        <UserAvatarGroup
+          :overlap="false"
+          :users="team.map((teamMember) => teamMember.user)"
+          :max-avatars="3"
+        />
+        <button
```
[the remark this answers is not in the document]
andrewwallacespeckle: If you mean `CommonBadge`, we could but it doesn't have a dashed variant, it also doesn't have a click event on the whole thing, so i'd have to wrap it in a button anyway. It would also be purple so i'd have to use the color-classes to style it. Seems like a lot of overwrites for no real benefit that I can see.

packages/frontend-2/components/workspace/ProjectList.vue
andrewwallacespeckle: I was going to do it in another branch, but there is a new design for the empty state. I've just done it in this branch now:
image

packages/frontend-2/components/workspace/ProjectList.vue
andrewwallacespeckle: image
Seems to work ok here, can you try with latest and see if it's working again?

packages/ui-components/src/components/layout/sidebar/menu/group/Group.vue
```diff
@@ -59,17 +66,20 @@
...
 import CommonBadge from '~~/src/components/common/Badge.vue'
 
 defineProps<{
   tag?: string
   title?: string
   collapsible?: boolean
   collapsed?: boolean
-  plusText?: string
-  plusClick?: () => void
+  icon?: 'add' | 'edit'
```
Mikehrn, line 79: Also not a blocker; but we can probably just make it accept any icon, like how buttons for example do

Mikehrn (dismissed):

packages/ui-components/src/components/layout/sidebar/menu/group/Group.vue
andrewwallacespeckle: Yeah, this can definitely be improved, I'll do it in another branch though

Mikehrn (dismissed):

Mikehrn (approved):

Mikehrn (approved):

PR #3717, opened 2024-12-18
Offline Object Loader

AlexandruPopovici:
## Description & motivation
This PR enables the loading of speckle objects from JSON.

## Changes:
- The `object-loader` library now defines a `createFromJSON` static method that returns an anonymous extension of the ObjectLoader class that handles parsing JSON strings. This functionality currently exists in this form so we can have it 100% isolated from the existing object-loader implementation.
- The `viewer` library now defines a `SpeckleOfflineLoader` loader implementation that makes use of the above and handles offline JSON loading transparently.

## Checklist:


didimitrie (approved):
i trust alex. i'm actually sorry i don't have time to doa more proper reivew, as this is interesting!

AlexandruPopovici:
The idea with this PR is to have something published for Oguzhan and Dogukan to run the offline loader in power-bi. Because a total revamp of the object-loader library is scheduled for early next year, we intentionally kept the offline object loader implementation minimal and strictly functional since it will get scrapped in the near future anyway. The focus was on making it 100% isolated from the rest of the object-loader implementation so there is no risk of regression.

PR #3720, opened 2024-12-18
feat(users): bulk lookup users by email

cdriesler:
## Description & motivation

- The invites flow reaches a point where it knows:
  - A collection of emails
  - Where it would like to invite them to
- The current lookup allows searching one query => many user matches, but not many queries => any matches

## Changes:

- Adds `usersByEmail` query modeled after `users` query
- Extracts common logic to `lookupUsersBaseQuery`
- Adds tests for both lookups


cdriesler:
The PR is parity with existing data egress if you know/guess a user's email. (You must be signed in/have a token.) The ability to do this in bulk does speed things up, in theory. I've lowered the limit to 20 instead of 100 (cc @Mikehrn it sounded like this would be plenty for invite use cases?)

If we want to add additional rate limiting I think that can happen in a follow-up so we unblock Mike

Mikehrn:
> The PR is parity with existing data egress if you know/guess a user's email. (You must be signed in/have a token.) The ability to do this in bulk does speed things up, in theory. I've lowered the limit to 20 instead of 100 (cc @Mikehrn it sounded like this would be plenty for invite use cases?)
> 
> If we want to add additional rate limiting I think that can happen in a follow-up so we unblock Mike

Nice :) Yeah 20 is totally fine, I will limit it 20 as well in the FE, multiple invites at one are are not super common, and if people invite multiple people it's usually 2-3 max

cdriesler:
> Can you create a ticket to review rate limiting for this?

Absolutely: https://linear.app/speckle/issue/WEB-2425/rate-limiting-for-bulk-user-lookup

PR #3721, opened 2024-12-18
feat(workspaces): add workspaceRole to LimitedUser type

cdriesler:
## Description & motivation

- The invite flow needs to check if a given `LimitedUser` from a user search is already a member of a known workspace

## Changes:

- Adds `workspaceRole` to `LimitedUser` type, modeled after `workspaceDomainPolicyCompliant`


cdriesler:
Ready for re-review. We now assert that users who query `LimitedUser.workspaceRole(workspaceId)` are at least members of the given workspace.

With this change, there shouldn't be any security implications. This is parity with current data access, just visible another way. (Workspace members can see peoples' roles)

gjedlicska (approved):

PR #3723, opened 2024-12-19
Feat: New workspace invite modal

Mikehrn:
This implements the new workspace invite modal. 
- In the first version the second modal in the flow, to select a server role, is left out because the BE for it is not ready yet. It default to Server.User, if the user is already a server user under another role it will not change it.
- It's possible to invite somebody that is already in the workspace, the only consequence of this is that the invite will not be send. Other invites will still be sent,

Screenshot 2024-12-19 at 14 46 30

Screenshot 2024-12-19 at 14 49 00


packages/frontend-2/components/invite/dialog/Workspace.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div>
+    <InviteDialogSharedSelectUsers
```
Mikehrn, line 3: There will be multiple modals here, it's just left out of this branch, which is an abstract of the branch which has all the modals

packages/frontend-2/components/invite/dialog/Workspace.vue
```diff
@@ -0,0 +1,147 @@
...
+
+const props = defineProps<{
+  workspace: InviteDialogWorkspace_WorkspaceFragment
+}>()
+const isOpen = defineModel<boolean>('open', { required: true })
+
+const mixpanel = useMixpanel()
+const inviteToWorkspace = useInviteUserToWorkspace()
+
+const isSelectUsersOpen = ref(false)
```
Mikehrn, line 68: This is not super relevant, in the branch there is only one submodal, but will make merging back later easier

andrewwallacespeckle:
image
It is possible to unselect the role select, and submit with no role. Expected?

Mikehrn:
> image
> It is possible to unselect the role select, and submit with no role. Expected?

Nope, I added the prop shouldnt be possible anymore :)

packages/frontend-2/components/invite/dialog/shared/SelectUsers.vue
```diff
@@ -0,0 +1,165 @@
...
+                mount-menu-on-body
+                :allow-unset="false"
+                :show-label="index === 0"
+                :disabled-items="getDisabledWorkspaceItems(item.value.email)"
+                disabled-item-tooltip="This email does not match the set domain policy, and can only be invited as a guest"
+              />
+            </div>
+          </div>
+          <div class="relative w-4">
+            <CommonTextLink
```
andrewwallacespeckle, line 55: Not too important, but for icon clicks i've been using FormButton with `hide-text` `subtle` and `icon-left`. This means there is a hover effect consistent with our other buttons

andrewwallacespeckle:
> One potential improvement - (if this is a lot of work to implement, it shouldn't block golive)
> 
> If I add a lot of users via this dialog, and 1 fails, all invites fail. I have to go in and re-add all users.
> 
> I had it fail because a user was already a member of the workspace, but it didn't say who

This was a false-positive. Confirmed working

PR #3726, opened 2024-12-20
chore: Track embed views

benjaminvo:
Track embed loads – we never implemented this Mixpanel event in FE2.


Mikehrn (dismissed):

packages/frontend-2/components/viewer/PreSetupWrapper.vue
```diff
@@ -198,4 +199,14 @@ const lastUpdate = computed(() => {
 })
 
 useHead({ title })
+
+const mp = useMixpanel()
+onMounted(() => {
+  const referrer = document.referrer
+  const shouldTrackEvent = !referrer?.includes('speckle.systems') || import.meta.dev
```
andrewwallacespeckle, line 206: Is this name accurate? Isn't the condition more `shouldNotTrackEvent`?

andrewwallacespeckle (changes requested):

packages/frontend-2/components/viewer/PreSetupWrapper.vue
andrewwallacespeckle: Or maybe the error is the missing `!` in import.meta.dev?

andrewwallacespeckle (approved):

PR #3740, opened 2024-12-20
refactor(fe2): various changes to workspace avatar

andrewwallacespeckle:
image
image

- Align the styling of the workspace icon – square instead of round and with they gray BG and letter
- Call it "icon" instead of "logo"
- Remove the highlight text below
- The tooltip "Only admins can edit this field" shows even if you're the admin


Mikehrn (dismissed):
Nice one 😄

packages/frontend-2/components/workspace/Avatar.vue
```diff
@@ -1,9 +1,12 @@
 <template>
-  <div
+  <component
+    :is="logo ? 'div' : 'button'"
+    v-tippy="logo ? undefined : 'Add a workspace icon'"
     :class="[
       'flex shrink-0 overflow-hidden rounded-md border border-outline-2 bg-foundation-2',
       sizeClasses
     ]"
+    @click="logo ? undefined : openSettingsDialog(SettingMenuKeys.Workspace.Members)"
```
benjaminvo, line 9: Looks like this will open the Members tab not the General?

benjaminvo (approved):

PR #3742, opened 2024-12-22
refactor(blobstorage): tests converted to typescript

iainsproat:
## Description & motivation

Typescripting more of the things.

## Changes:

- converts remaining modules/blobstorage tests to typescript.
- uses a random user, as a small start to ensuring tests are not polluting and can be run in parallel (they can't be, currently).
- requires that the owning user be provided as a parameter to the `createStreamForTest` function. This reduces surprising side-effects (in the previous case, the global user being used as the owner).
- adds a comment to a helper method with, possibly surprising, side-effects.

## Checklist:

### DIFF
--- a/packages/server/test/speckle-helpers/streamHelper.ts
+++ b/packages/server/test/speckle-helpers/streamHelper.ts
@@ -134,7 +134,7 @@ export async function createTestStreams(
 /**
  * Create basic stream for testing and update streamObj in-place, via reference, to have a real ID
  */
-export async function createTestStream( //FIXME this function has side-effects, and amends the object in-place
+export async function createTestStream(
   streamObj: Partial<BasicTestStream>,
   owner: BasicTestUser
 ) {


fabis94 (approved):

packages/server/test/speckle-helpers/streamHelper.ts
```diff
@@ -132,10 +132,10 @@ export async function createTestStreams(
 }
 
 /**
- * Create basic stream for testing and update streamObj to have a real ID
+ * Create basic stream for testing and update streamObj in-place, via reference, to have a real ID
  */
-export async function createTestStream(
-  streamObj: BasicTestStream,
+export async function createTestStream( //FIXME this function has side-effects, and amends the object in-place
```
fabis94, line 137: @iainsproat The side-effects are definitely intentional

packages/server/modules/blobstorage/tests/blobstorage.integration.spec.ts
```diff
@@ -86,33 +93,29 @@ const createToken = createTokenFactory({
...
     name: 'Anutha Blob Test Workspace #1',
     ownerId: '',
     id: '',
     slug: ''
   }
 
-  const createStreamForTest = async () => {
-    const stream = {
+  const createStreamForTest = async (streamOwner: BasicTestUser) => {
+    const stream: Partial<BasicTestStream> = {
```
fabis94, line 107: Isn't it always the same user?

fabis94 (changes requested):

packages/server/test/speckle-helpers/streamHelper.ts
fabis94: well, this aspect of the function is a part of its core responsibilities, its pretty important to have it cause manually having to import id-generating functions, having to remember the correct lengths & formats of ids is a PITA that this aspect of the helper resolves. in fact, this was one of the main reasons for me even creating these. it would be a serious impact to DX if we were to remove it.

its not immediately obvious to me how these helpers would cause issues in parallel tests, but if any pop up we can deal with them then

fabis94 (approved):

PR #3762, opened 2025-01-03
Fix: Updates to trial expired state in billing settings

Mikehrn:
Trial expired state was not being reflected correctly on the billing state, some parts are similar to trial and some exceptions. This also makes adjustments for the cancelled state (active subscription that is manually cancelled), it's not 100% as I need to verify with Gergö wether downgrading is possible or not. It's not a relevant scenario yet but will make it easier to adjust later.
 
Screenshot 2025-01-03 at 13 41 49


fabis94 (approved):
some of the conditionals repeat and could be extracted to their own computeds

benjaminvo (approved):

PR #3763, opened 2025-01-03
Fix: Revert notification changes

Mikehrn:
- Revert back to old notification system of only triggering one notification at the time
- Add option to hide toasts in compostables, and handle them in the parent in the case of server invites

### DIFF
--- a/packages/frontend-2/components/invite/dialog/Server.vue
+++ b/packages/frontend-2/components/invite/dialog/Server.vue
@@ -135,50 +135,52 @@ const removeInviteItem = (index: number) => {
   removeInvite(index)
 }
 
-const onSubmit = handleSubmit(() => {
+const onSubmit = handleSubmit(async () => {
   const invites = fields.value.filter((invite) => invite.value.email)
 
-  invites.forEach(async (invite) => {
-    invite.value.project
-      ? await inviteUserToProject(
-          invite.value.project.id,
-          [
-            {
-              email: invite.value.email,
-              serverRole: invite.value.serverRole
-            }
-          ],
-          { hideToasts: true }
-        )
-      : await inviteUserToServer(
-          [
-            {
-              email: invite.value.email,
-              serverRole: invite.value.serverRole
-            }
-          ],
-          { hideToasts: true }
-        )
-  })
+  try {
+    await Promise.all(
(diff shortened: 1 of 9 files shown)


packages/frontend-2/components/invite/dialog/Server.vue
```diff
@@ -133,35 +135,52 @@ const removeInviteItem = (index: number) => {
...
+      type: 'server invite',
+      name: 'send',
+      multiple: fields.value.length !== 1,
+      count: fields.value.length,
+      hasProject: !!fields.value.some((invite) => invite.value.project),
+      to: 'email'
+    })
+
+    isOpen.value = false
+  } catch {
```
Mikehrn, line 175: The change of any of them failing is slim, main cause probably network errors, as validation is already done, but this adds at least some basic error handling in the case this happens

packages/frontend-2/components/invite/dialog/Server.vue
```diff
@@ -133,35 +135,52 @@ const removeInviteItem = (index: number) => {
...
+    triggerNotification({
+      type: ToastNotificationType.Success,
```
fabis94, line 177: error not success

packages/frontend-2/lib/common/composables/toast.ts
```diff
@@ -1,93 +1,60 @@
...
 import { useSynchronizedCookie } from '~/lib/common/composables/reactiveCookie'
-import { nanoid } from 'nanoid'
 
 /**
  * Persisting toast state between reqs and between CSR & SSR loads so that we can trigger
  * toasts anywhere and anytime
  */
 const useGlobalToastState = () =>
-  useSynchronizedCookie<Optional<ToastNotification[]>>('global-toast-state')
+  useSynchronizedCookie<Optional<ToastNotification>>('global-toast-state')
```
fabis94, line 12: i'm wondering what will happen when the page loads in for someone who already has this cookie set with an array of toast notifications, hopefully the code doesn't break (expects a single object, finds an array).

its probably fine, but we should probably validate this assumption locally

fabis94 (changes requested):

fabis94 (approved):
lgtm

PR #3772, opened 2025-01-07
feat(server): ability to set global knex .select() query timeout

fabis94:
https://knexjs.org/guide/query-builder.html#timeout

^this, but globally set on all .select() queries

This is an attempt to deal with the issue where DB connections get taken up by long running queries. Timeout is in ms and can be adjusted through the helm chart

### DIFF
--- a/packages/server/bootstrap.js
+++ b/packages/server/bootstrap.js
@@ -30,10 +30,6 @@ if (isApolloMonitoringEnabled() && !getApolloServerVersion()) {
   process.env.APOLLO_SERVER_USER_VERSION = getServerVersion()
 }
 
-//knex is a singleton controlled by module so can't wait til app init
-const { initOpenTelemetry } = require('./otel')
-initOpenTelemetry()
-
 // If running in test env, load .env.test first
 // (appRoot necessary, cause env files aren't loaded through require() calls)
 if (isTestEnv()) {
@@ -48,6 +44,12 @@ if (isTestEnv()) {
 }
 
 dotenv.config({ path: `${packageRoot}/.env` })
+
+//knex is a singleton controlled by module so can't wait til app init
+const { initOpenTelemetry } = require('./otel')
+initOpenTelemetry()
+const { patchKnex } = require('./modules/core/patches/knex')
+patchKnex()
 
 module.exports = {
   appRoot,
--- a/packages/server/modules/core/patches/knex.ts
+++ b/packages/server/modules/core/patches/knex.ts
@@ -14,7 +14,10 @@ export const patchKnex = () => {
   // Add .timeout() to all .select() calls
   if (DEFAULT_QUERY_TIMEOUT_MS) {
     const originalSelect = knexPgQueryBuilder.prototype.select
-    knexPgQueryBuilder.prototype.select = function (...args: any) {
+    const newSelect: typeof originalSelect = function (
+      this: typeof knexPgQueryBuilder.prototype,
+      ...args: any
+    ) {
       let ret = originalSelect.apply(this, args)
       if (!this._timeout) {
         ret = ret.timeout(DEFAULT_QUERY_TIMEOUT_MS, { cancel: true })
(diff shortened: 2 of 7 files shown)


packages/server/bootstrap.js
```diff
@@ -49,6 +45,12 @@ if (isTestEnv()) {
 
 dotenv.config({ path: `${packageRoot}/.env` })
 
+//knex is a singleton controlled by module so can't wait til app init
+const { initOpenTelemetry } = require('./otel')
```
fabis94, line 49: i'm assuming there wasnt a specific reason to put this stuff before dotenv init

gjedlicska:
I think before we merge this, it would be nice to know, what are the queries, that are long running. 

We see this pattern only on the objects services, which has not that many repository functions. I think we should build some metrics to nail down the specific query and figure out why its causing the stuck pattern.

My bet is still on the streaming from the DB stuff being buggy, but that is only a hunch...

packages/server/bootstrap.js
adamhathcock: There is/was: otel didn't like other things being initialized before it.

adamhathcock (commented):
Maybe that fixes it 😬

fabis94:
@gjedlicska I think we need both - a sane default timeout and a UI (which is another thing I started sketching out).

I can't see a reason why we would not want a default timeout for selects, from my experience thats pretty standard. I was in fact surprised that we haven't done this yet. Remember that the timeout can be very high, like Iain mentioned 30 minutes. But we have to have something, cause these DB connection issues are getting ridiculous

packages/server/bootstrap.js
fabis94: @adamhathcock even dotenv? sounds very odd, what exactly did it not like about it? note that even in its existing form there is stuff happening before it

fabis94:
@fabis94 The LISTEN calls are not .select() queries, in fact they sidestep knex entirely and are invoked on a pg.Connection directly.

Regarding adding this to other services - We can, but if i understand correctly the DB connection issues usually happen to `server` pods, no? Also, might be worth testing this out first on `server`, and then if there's no issues, we can also add it to the other ones

packages/server/bootstrap.js
adamhathcock: maybe not that but I just put it above most/everything to be sure

gjedlicska:
There has to be a real technical reason, why knex does not have this feature by default. There are so many projects using knex, that calling the other, non knex, behavior something a "standard", is a bit bold for me. 

Also, we're forcing our own "non-standard" way on knex, just to solve a specific problem, that we do not even know what the actual problem is. Pls lets not do this before we actually know what problems we want to solve with this. Its like shooting at a pidgeon with a cannon made out of wood.

Some ppl have very large models and on crappy internet connections receives can take a while. In the Autodesk Revit world sync to central (publish or load in our terminology) is common place to take hours on massive projects. We do not have enough data to set a global timeout for these workloads, and most probably, that is where the actual problem lies. All other gql api serving queries are stable and behave normally and free up connections as they supposed to. That is the pattern we've been seeing for years now.

So, lets pinpoint the query, that is responsible for the stuck connection, provide data, to set a timeout for that specifically is the solution that we should go for.

fabis94:
> There has to be a real technical reason, why knex does not have this feature by default. There are so many projects using knex, that calling the other, non knex, behavior something a "standard", is a bit bold for me.

knex, [the library that only removed bluebird promises from their lib in 2020](https://github.com/knex/knex/pull/3650) when native promises had already existed for half a decade and that still hasn't merged my many year old PR for enabling default timestamp precision, is not really trustworthy when it comes to their defaults or expectations about how things should work. Its an old and outdated DB lib, one that I would definitely not use in any new projects.

Just having a global query timeout **is** standard across apps and programming languages, even if its a large timeout. Considering that we have a pitiful amount of available DB connections and that we've been having DB connections for as far back as I can remember, I would think that cutting down on rogue queries whenever we can would be an easy win, but sure, whatever you say boss

fabis94:
@iainsproat Yea, I can understand the point about metrics, just relying on seq errors may not be enough 👍

PR #3779, opened 2025-01-08
fix(auth): remove direct form value mutation in SSO login

andrewwallacespeckle:
Remove direct form value mutation through v-model in SSO login component.
The component already handles form state through internal vee-validate integration.


fabis94 (approved):
looks good, the CI failure can be ignored too

PR #3790, opened 2025-01-09
fix(fe2): SSR memory leak caused by vue apollo

fabis94:
you can find the full source code of the fork as well as my specific changes through this PR: https://github.com/vuejs/apollo/pull/1582/files


packages/ui-components/package.json
```diff
@@ -45,6 +45,7 @@
     }
   },
   "peerDependencies": {
+    "vee-validate": "^4.7.0",
```
fabis94, line 48: made it a peerDep so that it can be independently controlled

Mikehrn (dismissed):

package.json
```diff
@@ -59,6 +59,7 @@
     "husky": "^7.0.4",
     "lint-staged": "^12.3.7",
     "lockfile": "^1.0.4",
+    "npkill": "^0.12.2",
```
Mikehrn, line 62: Only curious why we need this dependency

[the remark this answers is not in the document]
fabis94: i dislike using npx in a yarn project, because it uses an entirely different package manager to install the dep in a yarn project. you don't need a script in package.json, you can just do `yarn npkill`

fabis94: common dev tool for resetting node_modules

packages/frontend-2/package.json
```diff
@@ -53,7 +53,7 @@
     "@tiptap/suggestion": "2.10.3",
     "@tiptap/vue-3": "2.10.3",
     "@tryghost/content-api": "^1.11.21",
-    "@vue/apollo-composable": "4.0.2",
+    "@vue/apollo-composable": "npm:@speckle/apollo-composable@4.2.1-patch.1",
```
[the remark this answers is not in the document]
fabis94: @iainsproat i added a link to the description

PR #3802, opened 2025-01-13
feat(fe2): allow logging CSR non-errors to seq

fabis94:
logger calls w/ any level can be sent to Seq if they've got `important: true` set

### DIFF
--- a/packages/frontend-2/plugins/001-logger.ts
+++ b/packages/frontend-2/plugins/001-logger.ts
@@ -2,9 +2,9 @@ import { omit } from 'lodash-es'
 import type { SetRequired } from 'type-fest'
 import { useReadUserId } from '~/lib/auth/composables/activeUser'
 import {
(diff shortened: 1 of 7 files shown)


packages/frontend-2/plugins/001-logger.ts
```diff
@@ -293,6 +314,15 @@ export default defineNuxtPlugin(async (nuxtApp) => {
     })
   }
 
+  // Log important tier app mounted msg for seq
+  if (!import.meta.server) {
+    nuxtApp.hook('app:mounted', () => {
+      logger.info('App mounted in the client', {
```
fabis94, line 320: example - now in Seq we can correlate if SSR renders also lead to full hydration and load in the browser, as opposed to SSR render being triggered by curl or something and no clientside JS actually being loaded

Mikehrn (approved):

andrewwallacespeckle (approved):

PR #3803, opened 2025-01-13
fix(fe2): admin project list comment threads authorization

andrewwallacespeckle:
[DISCORD THREAD 
](https://discord.com/channels/726756379083145269/1327282803909591093)

image
image
image

I noticed this was also happening on latest. Looking at the network tab, the GQL error `STREAM_INVALID_ACCESS_ERROR` happens when trying to fetch `commentThreads` in the admin project list.

The issue seems to be in `authorizeProjectCommentsAccessFactory` where the admin authorisation check happens after other permission checks, overriding the admin access. 

I've made a fix and opened a PR, but since it's backend code I'd appreciate if @iamcgi , @fabiansgeikinsspeckle or anyone with knowledge on this could sense check this for me.

### DIFF
--- a/packages/server/modules/comments/services/management.ts
+++ b/packages/server/modules/comments/services/management.ts
@@ -67,13 +67,16 @@ export const authorizeProjectCommentsAccessFactory =
       throw new StreamInvalidAccessError('Stream not found')
     }
 
+    // Check admin access first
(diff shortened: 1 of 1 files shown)


gjedlicska:
with the more appropriate fix in #3806 we do not need this any more

PR #3804, opened 2025-01-13
fix(fe2): resolve PortalTarget SSR attribute inheritance warning

andrewwallacespeckle:
I noticed a hydration warning in console, so wrapped PortalTarget with ClientOnly to prevent SSR-related issues.

image


fabis94 (approved):
hoping we can get rid of the portal altogether at some point

PR #3813, opened 2025-01-14
Feat: Setting as pages

Mikehrn:
WIP: DONT MERGE


packages/frontend-2/components/settings/user/email/ListItem.vue
```diff
@@ -39,22 +33,28 @@
...
-          @click="toggleSetPrimaryDialog"
-        >
-          Set as primary
-        </FormButton>
-        <FormButton
-          :disabled="emailData.primary"
-          color="outline"
-          size="sm"
-          @click="toggleDeleteDialog"
+        <div v-tippy="primaryTooltip">
```
andrewwallacespeckle, line 36: image
This tooltip is the wrong text

Mikehrn: Nice catch thanks, fixed it :)

packages/frontend-2/lib/common/helpers/route.ts
```diff
@@ -21,6 +21,71 @@ export const defaultZapierWebhookUrl =
   'https://hooks.zapier.com/hooks/catch/12120532/2m4okri/'
 export const guideBillingUrl = 'https://speckle.guide/workspaces/billing.html'
 
+export const settingsRoutes = {
+  user: {
+    profile: {
```
Mikehrn, line 26: Both route and name are needed, because they are used in different scenarios (for example checking if a route by name needs SSO access)

packages/frontend-2/lib/workspaces/graphql/queries.ts
```diff
@@ -123,3 +123,11 @@ export const workspaceWizardRegionQuery = graphql(`
     }
   }
 `)
+
+export const workspaceGetIdBySlugQuery = graphql(`
```
Mikehrn, line 127: Some queries are not possible with workspaceBySlug, as nested properties in some cases need an ID and don't work with slug, so this call is done first to get the id

packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
```diff
@@ -8,20 +8,20 @@
         class="mb-6"
       />
       <LayoutTabsHorizontal v-model:active-item="activeTab" :items="tabItems">
-        <template #default="{ activeItem }">
+        <template v-if="workspace" #default="{ activeItem }">
```
Comment (fabis94) [inline packages/frontend-2/pages/settings/workspaces/[slug]/members.vue:11]: I dislike the fact that you need to load to completion the workspace GQL query before you can even start mounting the subsequent components and their queries. This is why I think I'd rather rely on the apollo.cache hack

packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
```diff
@@ -116,9 +110,9 @@
     </div>
 
     <SettingsWorkspacesSecurityDomainRemoveDialog
-      v-if="removeDialogDomain"
+      v-if="removeDialogDomain && workspace"
```
Comment (fabis94) [inline packages/frontend-2/pages/settings/workspaces/[slug]/security.vue:113]: ditto, every time we do v-if="workspace" on a component we're making the components load in serially, so if they have their own queries they're executed serially instead of in parallel

packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
```diff
@@ -258,23 +260,24 @@ const isDomainProtectionEnabled = computed({
...
     const mutationResult = await apollo.mutate({
       mutation: SettingsUpdateWorkspaceSecurityDocument,
       variables: {
         input: {
-          id: props.workspaceId,
+          id: workspace.value?.id,
           discoverabilityEnabled: newVal
         }
       },
       optimisticResponse: {
```
Comment (fabis94) [inline packages/frontend-2/pages/settings/workspaces/[slug]/security.vue:274]: can we at least add a FIXME here to clean this up? the optimisticResponse still doesn't make any sense to me and cache.modify should be replaced with modifyObjectField

fabis94 (changes requested):
looks very good for the most part, great to see a lot of this cleaned up

packages/frontend-2/pages/settings/server/members.vue
```diff
@@ -27,6 +27,15 @@ import {
 } from '~~/lib/server-management/graphql/queries'
 import { useQuery } from '@vue/apollo-composable'
 
+definePageMeta({
+  middleware: ['auth', 'settings', 'admin'],
```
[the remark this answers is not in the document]
fabis94: https://nuxt.com/docs/guide/directory-structure/pages#nested-routes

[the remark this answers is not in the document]
fabis94: Same goes for 'admin'  - assuming all admin pages go under settings/server/*, you can just add that middleware to server.vue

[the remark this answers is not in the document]
Mikehrn: Afaik this will also create a route for /settings, /user & /server, do you know if there a way to avoid it? I can redirect those, but I don't know if that's worth the trade-off of just defining the middleware

packages/frontend-2/lib/workspaces/graphql/queries.ts
Mikehrn: Do you remember which hack this was, or maybe @andrewwallacespeckle knows? I vaguely remember it, but dont remember where we are doing this

packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
Comment (Mikehrn) [reply packages/frontend-2/pages/settings/workspaces/[slug]/security.vue:274]: Will do 👍

packages/frontend-2/pages/settings/server/members.vue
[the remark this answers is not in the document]
fabis94: @Mikehrn we already do this for /authn/* pages, see authn.vue for how we redirect away people who just go to /authn not /authn/login or whatever.

It seems easier to do that once than to have to remember to put the correct middlewares on every single subpage (something that we have already forgotten to do in the past for certain pafges)

packages/frontend-2/lib/workspaces/graphql/queries.ts
fabis94: I don't know, I think Andrew will know more

packages/frontend-2/pages/settings/server/members.vue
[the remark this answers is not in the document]
fabis94: @Mikehrn Actually there's another option: https://nuxt.com/docs/guide/concepts/rendering#route-rules

We could define a /settings/* route rule to have the appMiddleware keys that we need. This could even be the simplest route.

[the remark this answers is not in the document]
Mikehrn: @fabis94 Nice thanks!

packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
Comment (Mikehrn) [reply packages/frontend-2/pages/settings/workspaces/[slug]/security.vue]: @fabis94 I changed this to use normal mutations, manual updating the cache and optismisticResponse is now not needed anymore

packages/frontend-2/lib/settings/composables/menu.ts
```diff
@@ -38,36 +23,48 @@ export const useSettingsMenu = () => {
...
   const isMultiRegionEnabled = useIsMultiregionEnabled()
 
-  const workspaceMenuItems = shallowRef<SettingsMenuItems>({
-    [SettingMenuKeys.Workspace.General]: {
+  const workspaceMenuItems = shallowRef<SettingsMenuItem[]>([
+    {
       title: 'General',
-      component: SettingsWorkspaceGeneral,
+      getRoute: (slug?: string) =>
+        slug ? settingsRoutes.workspace.general.route(slug) : '',
```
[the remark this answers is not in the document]
Mikehrn: Updated this to use a separate type for the two

packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
Comment (Mikehrn) [reply packages/frontend-2/pages/settings/workspaces/[slug]/members.vue]: With some refactoring I was able to just rely on just the slug, so we don't need the hack to get the ID

packages/frontend-2/lib/workspaces/graphql/queries.ts
Mikehrn: Worked around this by refactoring some of the child components to use workspaceBySlug

packages/frontend-2/nuxt.config.ts
```diff
@@ -186,23 +186,30 @@ export default defineNuxtConfig({
...
+      }
+    },
+    '/settings/**': {
+      appMiddleware: ['auth', 'settings']
+    },
+    '/settings/server/*': {
+      appMiddleware: ['auth', 'settings', 'admin']
+    },
+    '/settings/workspaces/*': {
+      appMiddleware: ['auth', 'settings', 'admin', 'requires-workspaces-enabled']
```
fabis94, line 212: shouldn't these also have the workspace access check middleware?

packages/frontend-2/nuxt.config.ts
```diff
@@ -186,23 +186,30 @@ export default defineNuxtConfig({
...
+      redirect: {
+        to: '/settings/server/members',
+        statusCode: 301
+      }
+    },
+    '/settings/**': {
+      appMiddleware: ['auth', 'settings']
+    },
+    '/settings/server/*': {
+      appMiddleware: ['auth', 'settings', 'admin']
```
fabis94, line 209: hoping these don't get added & invoked twice (cause of the /settings/** rule above)

fabis94 (approved):
looks good

fabis94 (approved):

PR #3821, opened 2025-01-14
Fix: Only list project in workspace in automation project select

Mikehrn:


cdriesler (approved):
ty king

edit: missed that it was a draft will take a closer look later then

Mikehrn:
> ty king
> 
> edit: missed that it was a draft will take a closer look later then

It's only draft because I couldnt test it, if it's working as expected it's fine to change it to a normal PR :)

cdriesler (approved):

PR #3843, opened 2025-01-16
feat(regions): move project branches and commits

cdriesler:
## Checklist:


packages/server/modules/core/services/projects.ts
```diff
@@ -61,7 +61,7 @@ export const createNewProjectFactory =
             const replicatedProject = await getProject({ projectId })
             if (!replicatedProject) throw new StreamNotFoundError()
           },
-          { maxAttempts: 10 }
+          { maxAttempts: 100 }
```
alemagio, line 64: don't think this is necessary

packages/server/modules/multiregion/utils/dbSelector.ts
```diff
@@ -223,7 +223,16 @@ const setUpUserReplication = async ({
...
       )
-    if (!err.message.includes('already exists')) throw err
+    }
+
+    const errorMessage = err.message
+
+    if (
+      !['already exists', 'violates unique constraint'].some((message) =>
+        errorMessage.includes(message)
+      )
```
alemagio, line 233: just double check there is not a better way to make this check.
Like using an error code

packages/server/modules/workspaces/repositories/projectRegions.ts
```diff
@@ -0,0 +1,218 @@
...
+      throw new WorkspaceNotFoundError()
+    }
+
+    await tables
+      .workspaces(deps.targetDb)
+      .insert(workspace)
+      .onConflict(Workspaces.withoutTablePrefix.col.id)
+      .merge(Workspaces.withoutTablePrefix.cols as (keyof Workspace)[])
+
+    return workspaceId
```
alemagio, line 61: we should check how workspaces are propagated to the region db and use that to publish the change:
- always update the main db
- update (fake) the main db and use the publish system to update the region

alemagio (approved):

packages/server/modules/core/services/projects.ts
cdriesler: fixed 👍

packages/server/modules/multiregion/utils/dbSelector.ts
cdriesler: this does appear our best option in this case for now

packages/server/modules/workspaces/repositories/projectRegions.ts
cdriesler: It appears that we quite literally only care that the workspace id is defined in the regional database for foreign key constraints. Anything user-visible is pulled from the main db record. Simplified this to a copy then ignore on conflict.

alemagio (approved):

PR #3844, opened 2025-01-17
feat(fe2): Update permissions needed to use Gendo

andrewwallacespeckle:
Check user has permissions before generating AI renders. 

Update feedback mechanism


Mikehrn (dismissed):

packages/frontend-2/components/viewer/gendo/Dialog.vue
```diff
@@ -9,39 +9,101 @@
...
   renderPrompt?: string
 }>()
 
 const isOpen = defineModel<boolean>('open', { required: true })
+
+const isFeedbackDialogOpen = ref(false)
+const hasFeedback = ref(false)
+const initialFeedback = ref<'positive' | 'negative' | null>(null)
+
+const handleThumbsUp = () => {
```
Mikehrn, line 94: You could also have one function for both (and make render_url snakeCase, unless we already use it, and end the the event with a past-tense verb)

```
type FeedbackValue = 'positive' | 'negative';

const initialFeedback = ref<FeedbackValue>()

const handleFeedback = (value: FeedbackValue) => {
  initialFeedback.value = value
  mixpanel.track('Gendo Render Feedback Given', {
    feedback: value,
    prompt: props.renderPrompt,
    renderUrl: props.renderUrl
  })
  
  hasFeedback.value = true
}
```

Mikehrn (changes requested):

Mikehrn (approved):

PR #3845, opened 2025-01-17
chore(server): refactor activityStream invocations - batch #1 - user

fabis94:
## Checklist:


packages/server/modules/core/tests/usersAdmin.spec.ts
```diff
@@ -178,7 +171,7 @@ describe('User admin @user-services', () => {
         await changeUserRole({ userId: myTestActor.id, role })
         assert.fail('This should have failed')
       } catch (err) {
-        expect(err.message).to.equal(`Invalid role specified: ${role}`)
+        expect(ensureError(err).message).to.equal(`Invalid role specified: ${role}`)
```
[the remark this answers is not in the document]
fabis94: its a pretty common thing to do in TS - in JS anything can be thrown, so `err` is gonna be `unknown`, not something that you can access .message on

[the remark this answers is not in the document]
fabis94: 1. In practice, err is never a non-error. I don't think I've seen it happen even once at my whole time at speckle.
2. If its not an error, ensureError will already swap the error out for a different error type & message which will make the message assertion fail

packages/server/modules/activitystream/services/userActivity.ts
```diff
@@ -22,3 +40,38 @@ export const addUserUpdatedActivityFactory =
...
+      resourceType: 'user',
+      resourceId: targetUserId,
+      actionType: ActionTypes.User.Delete,
+      userId: invokerUserId,
+      info: {},
+      message: 'User deleted'
+    })
+  }
+
+export const reportUserActivityFactory =
```
gjedlicska, line 60: This func is a valid listener initalization func, but i think it is not a service. This should be part of the application itself, next to the module init file somewhere.

I have not yet landed on a good pattern for this, but probably something like `eventListeners.ts` should be a good start

gjedlicska (changes requested):
just a minor function placement remark, otherwise looks good

packages/server/modules/activitystream/services/userActivity.ts
fabis94: altho i've done it myself, i don't like putting files next to the module init, it feels like a shortcut to avoid thinking about a proper place to actually put them, and then you end up with a bunch of unrelated, disorganized modules in the same location

i also don't want this to be a single file, cause you could have a ton of listeners of different types/concerns, especially in activitystream

I ended up moving this to /events/userListeners.ts

gjedlicska: that makes sense, as we are moving into utilizing events more and more

gjedlicska (approved):

PR #3850, opened 2025-01-20
fix: tailwind config loading randomly breaking cause of import.meta in cjs

fabis94:


Mikehrn (approved):
I don't have the issue atm, but can at least confirm it still works as expected

PR #3853, opened 2025-01-20
chore(emails): add code column to email_verifications table

alemagio:
## Checklist:


packages/server/modules/emails/migrations/20250120145453_add_code_column_to_email_verifications.ts
```diff
@@ -0,0 +1,15 @@
+import { Knex } from 'knex'
+
+const TABLE_NAME = 'email_verifications'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(TABLE_NAME, (table) => {
+    table.string('code')
```
gjedlicska, line 7: Shouldn't this be unique?
Also this prob needs an index, if we're looking up the verification based on the code

gjedlicska (changes requested):

packages/server/modules/emails/migrations/20250120145453_add_code_column_to_email_verifications.ts
alemagio: I always store a hashed verion of the code 
so I don't see value in storing it as unique
Also, I always use the email to find the record.

There should be a constraint for the email to be unique, but I have another PR for the constraints

gjedlicska (approved):

PR #3854, opened 2025-01-20
Feat: Allow invites with subdomains

Mikehrn:


Mikehrn:
It's not solving the right problem

gjedlicska:
maybe you are

gjedlicska:
ok, turns out we're taking a different route, with allowing explicitly adding subdomains

PR #3863, opened 2025-01-22
feat(server): greatly optimizing Branch.commits & Query.streams, when limit=0

fabis94:
when limit === 0, there should be a lot less SQL queries


packages/server/modules/core/graph/plugins/logging.ts
```diff
@@ -107,37 +107,44 @@ export const loggingPluginFactory: (deps: {
           apollo_query_duration_ms: Date.now() - apolloRequestStart
         })
 
-        for (const err of ctx.errors) {
```
fabis94, line 110: this was the reason why there were many "req failed" entries logged for a single request

gjedlicska (approved):

PR #3874, opened 2025-01-23
chore(server): refactor activityStream invocations - batch #3 - branches

fabis94:
## Checklist:


packages/server/modules/core/services/branch/management.ts
```diff
@@ -113,12 +135,31 @@ export const updateBranchAndNotifyFactory =
...
+        deps.eventEmit({
+          eventName: ModelEvents.Updated,
+          payload: {
+            update: input,
+            userId,
+            oldModel: existingBranch,
+            newModel: newBranch
+          }
+        }),
+        // TODO: Move to event bus listeners
```
gjedlicska, line 148: r u planning to move these to an event listener in a separate PR?

gjedlicska: ideally, we could remove the publishSub func from the dependencies

gjedlicska (commented):
shouldn't we move the pubsubs out?

packages/server/modules/core/services/branch/management.ts
fabis94: @gjedlicska i'd rather have this be a separate task (or something to pick up within an unrelated task), cause its not really related to refactoring activityStream invocations. there's still a bunch I need to get through, and this is already better than what it was - activityStream invocations also triggering subscriptions

fabis94: if i wanna refactor this i need to now also introduce entirely new listeners only for subscriptions

fabis94: at that point, you could also clean up other events that happen during CRUD actions like "markBranchStreamAsUpdated", which strays even further from the original idea behind this task

gjedlicska (dismissed):

PR #3879, opened 2025-01-23
fix(server/rate limiting): Rate limit by IP or token ID

iainsproat:
## Description & motivation

- we need to protect the auth context middleware
- however, we will now never have the userId in the request context to use as a rate limit source, so will always limit by IP address.
- Rate limiter now behaves as it did before with unauthenticated users.

## Changes:

- rate limiting middleware is applied prior to the auth context middleware

## Checklist:


gjedlicska:
If it makes sense, we can rate limit based on the first 10 chars of the bearer token, that is a better approximation of a user

fabis94 (approved):

PR #3885, opened 2025-01-24
Fix(dui3): etabs set selection manually

oguzhankoral:
Etabs doesnt have selection events that we cannot setSelection dynamically.


oguzhankoral:
@bjoernsteinhagen fixed it differently, we do not need it anymore

PR #3895, opened 2025-01-24
Fix: Align forgot password screens

Mikehrn:
I made a few changes to these screens to align them with the other sign up screen, minor copy changes and added the password validation checks (personal papercut..)

Screenshot 2025-01-24 at 21 38 51
Screenshot 2025-01-24 at 21 35 37


benjaminvo (approved):
Thanks for changing this

PR #3915, opened 2025-01-30
Fix(dui3): handle update connector errors

oguzhankoral:
Since we do not have released version on ETABS and the exception was not handled, it caused the issue https://linear.app/speckle/issue/CNX-1144/installer-version-shows-documentless-state


bjoernsteinhagen:
Fixed ETABS' documentless state:
image

didimitrie (approved):

PR #3918, opened 2025-01-30
fix(server/errors): derive errors from BaseError

iainsproat:
## Description & motivation

`throw new Error` should only be present in exceptional circumstances. We should instead extend from `BaseError`, as this allows us to ensure an HTTP StatusCode is present and can be used to properly handle errors returned to the user and logged.

Currently `Error` will be returned to the user as a HTTP status code 500, and logged as error level. For many errors, this is incorrect. This PR aims to reduce or eliminate these occurrences.

## Changes:

- `throw new Error` replaced with an appropriate Error class extended from BaseError.

## Checklist:


fabis94 (approved):
I think as long as no existing BaseError types get swapped it should be fine

PR #3931, opened 2025-02-04
Fix(dui3): eport shows success despite more warning items than

oguzhankoral:
WARNING STATE
image

FAIL STATE
image

SUCCESS STATE
image


didimitrie (approved):
happiness level 5/10, but we know this

PR #3932, opened 2025-02-04
feat(fe2): New user onboarding flow

andrewwallacespeckle:
New onboarding flow. 

https://www.figma.com/design/CFVYfzYSyUdTVnrlFUEcUZ/Onboarding?node-id=1-3&p=f&t=m1kjiSLdkKaABHBv-0

2 feature flags to force onboarding and verification. 
FF_FORCE_ONBOARDING
FF_FORCE_EMAIL_VERIFICATION


packages/frontend-2/components/auth/LoginWithEmailBlock.vue
```diff
@@ -12,6 +12,7 @@
         show-label
         :disabled="!!(loading || shouldForceInviteEmail)"
         auto-focus
+        autocomplete="email"
```
andrewwallacespeckle, line 15: these autocomplete changes are to fix some console warnings

packages/frontend-2/components/header/WithEmptyPage.vue
```diff
@@ -0,0 +1,35 @@
+<template>
```
Mikehrn, line 1: This is in the `header` directory, but maybe beter suited as layout or page?

packages/frontend-2/components/onboarding/JoinTeammates.vue
```diff
@@ -0,0 +1,105 @@
...
+  <div class="flex flex-col items-center gap-2 w-full">
+    <CommonCard
+      v-for="workspace in workspaces"
+      :key="workspace.id"
+      class="w-full bg-foundation"
+    >
+      <div class="flex gap-4">
+        <div>
+          <WorkspaceAvatar
+            v-if="workspace"
```
Mikehrn, line 11: Check if this doesn't cause any hydration issue, I fixed something similar in another PR where instead I had to update the children to use MayBeNullOrUndefined instead

packages/frontend-2/components/onboarding/questions/Form.vue
```diff
@@ -0,0 +1,56 @@
...
+      <FormButton size="lg" :disabled="!meta.valid || isSubmitting" submit full-width>
+        Continue
+      </FormButton>
+      <div class="opacity-70 hover:opacity-100 max-w-max mx-auto px-1">
+        <FormButton
+          v-if="!isOnboardingForced"
+          size="sm"
+          text
+          link
+          color="subtle"
```
Mikehrn, line 16: Is subtle still relevant as a prop when it's text/link?

packages/frontend-2/components/settings/user/email/DeleteDialog.vue
```diff
@@ -52,24 +47,10 @@ const dialogButtons = computed((): LayoutDialogButton[] => [
...
-
-    mixpanel.track('Email Deleted')
-  } else {
-    const errorMessage = getFirstErrorMessage(result?.errors)
-    triggerNotification({
-      type: ToastNotificationType.Danger,
-      title: errorMessage
-    })
+  if (!props.email) return
+  const success = await deleteUserEmail(props.email)
```
Mikehrn, line 51: Won't we still need the notifications and mixpanel event here in case of deletion?

packages/frontend-2/lib/auth/composables/activeUser.ts
```diff
@@ -68,6 +72,17 @@ export function useActiveUser() {
   const isGuest = computed(() => activeUser.value?.role === Roles.Server.Guest)
   const isAdmin = computed(() => activeUser.value?.role === Roles.Server.Admin)
 
+  // const emails = computed(() => activeUser.value?.emails)
```
Mikehrn, line 75: Left over?

packages/frontend-2/lib/auth/composables/auth.ts
```diff
@@ -480,6 +483,12 @@ const useAuthAppIdAndChallenge = () => {
   return { appId, challenge }
 }
 
+/**
+ * Indicates whether the user just completed registration
+ */
+export const useIsJustRegistered = () =>
```
Mikehrn, line 489: PP: Maybe make it more specific to this current session
```
export const useRegisteredThisSession = () =>
  useState<boolean>('registered-this-session', () => false)
  ```

File: packages/frontend-2/lib/auth/composables/onboarding.ts [hunk 8]
```diff
@@ -1,7 +1,7 @@
 import { useApolloClient } from '@vue/apollo-composable'
 import { useMixpanel } from '~~/lib/core/composables/mp'
 import { UnsupportedEnvironmentError } from '~~/lib/core/errors/base'
-import type { OnboardingState } from '~~/lib/auth/helpers/onboarding'
+import type { OnboardingState } from '../helpers/onboarding'
```
Mikehrn, line 4: Any specific reason for changing this from `~~/lib/auth/` to `../`?

packages/frontend-2/lib/auth/composables/onboarding.ts
```diff
@@ -15,26 +15,20 @@ import { ToastNotificationType, useGlobalToast } from '~~/lib/common/composables
...
   const goHome = useNavigateToHome()
 
-  /**
-   * Sends to mp the segmentation info (industry, role)
-   * @param state
-   */
-  const setMixpanelSegments = (state: OnboardingState) => {
-    mixpanel.people.set_once(ONBOARDING_PROP_INDUSTRY, state.industry || null)
-    mixpanel.people.set_once(ONBOARDING_PROP_ROLE, state.role || null)
+  const setMixpanelSegments = (segments: Partial<OnboardingState>) => {
```
Mikehrn, line 30: Will there be a way to map previous user properties to the new one? Will they use new names?

packages/frontend-2/lib/user/composables/emails.ts
```diff
@@ -0,0 +1,179 @@
...
+    const errorMessage = getFirstErrorMessage(result?.errors)
+    triggerNotification({
+      type: ToastNotificationType.Danger,
+      title: 'Error deleting email',
+      description: errorMessage
+    })
+    return false
+  }
+
+  const verifyUserEmail = async (email: UserEmail, code: string) => {
```
Mikehrn, line 123: Maybe we can add some additional events here, so we can track the funnel of adding emails?
Email Added -> Email Verification Started -> Email Verified / Email Verified Cancelled, something like this maybe

packages/frontend-2/middleware/004-emailVerification.global.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!data?.activeUser?.id) return
+
+  const isAuthPage = to.path.startsWith('/authn/')
+  const isVerifyEmailPage = to.path === verifyEmailRoute
+
+  // Skip if not forced and not on verify email page
+  if (!isEmailVerificationForced.value && !isVerifyEmailPage) return
+  if (isAuthPage) return
+
+  const hasUnverifiedEmails = data.activeUser.emails.some((email) => !email.verified)
```
Mikehrn, line 28: What will happen if you have a secondary unverified email? Will you also be forced into this flow until you verify it?

Mikehrn (changes requested):
Overall looks good :) Couple small things, will also spend some time testing it today

packages/frontend-2/components/onboarding/JoinTeammates.vue
andrewwallacespeckle: I confirmed there was no hydration mismatch. In fact, I can remove the v-if completely because workspace will always be defined by the v-for

packages/frontend-2/components/onboarding/questions/Form.vue
andrewwallacespeckle: Yeah, it still adjusts the text colour. 
Screenshot 2025-02-07 at 13 54 09
image

packages/frontend-2/components/settings/user/email/DeleteDialog.vue
andrewwallacespeckle: This is now in the useUserEmails composable
image

packages/frontend-2/middleware/004-emailVerification.global.ts
andrewwallacespeckle: Yeah, i checked and this is how we want it. But, you will get the cancel button for a secondary email, so you can cancel adding it and restore app access. 

Really, the cancel button is deleting the email from the user.

packages/frontend-2/lib/auth/composables/onboarding.ts
andrewwallacespeckle: There were new names, supplied by Matteo, so I'm not sure how we could map them to the old ones, or if there is any value to it.

benjaminvo (changes requested):
I tested and found some things. Added to Linear: https://linear.app/speckle/project/user-creation-and-activation-34915d92dd9a/issues

packages/frontend-2/pages/onboarding.vue
```diff
@@ -1,70 +1,107 @@
...
-import { homeRoute, modelRoute, projectRoute } from '~~/lib/common/helpers/route'
+import { useProcessOnboarding } from '~~/lib/auth/composables/onboarding'
+import { useAuthManager } from '~/lib/auth/composables/auth'
+import { useQuery } from '@vue/apollo-composable'
+import { graphql } from '~/lib/common/generated/gql'
+import { useActiveUser } from '~~/lib/auth/composables/activeUser'
+import { CommonLoadingIcon } from '@speckle/ui-components'
+
+const discoverableWorkspacesQuery = graphql(`
+  query DiscoverableWorkspaces {
```
Mikehrn, line 56: Our current pattern is to define the query in the component following the naming of the component, so PagesOnboardingDiscoverableWorkspaces_ActiveUser` or something like that, and have the query in the separate file that groups them

Mikehrn (approved):

benjaminvo (approved):

PR #3938, opened 2025-02-05
fix(server/database streams): remove event listener on batch completion

iainsproat:
## Description & motivation

In response to observed log line:

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [ServerResponse]. Use emitter.setMaxListeners() to increase limit
```

This might explain the remaining database connection leaks. We're batching the database streams and each batch adds a listener to the response. Large models with a large number of batches could exceed the number of listeners.
This max value will limit the number of listeners, meaning that subsequent batches cannot add listeners and will not be closed if the client connection is subsequently closed.

## Changes:

- refactors anonymous event handler to a named event handler
- removes the named event handler from the response event listeners when the database stream is closed

## Checklist:

### DIFF
--- a/packages/preview-service/src/server/routes/api.ts
+++ b/packages/preview-service/src/server/routes/api.ts
@@ -36,12 +36,18 @@ const apiRouterFactory = () => {
         streamId: req.params.streamId,
         objectIds: getObjectsRequestBody.objects
       })
+
       // https://knexjs.org/faq/recipes.html#manually-closing-streams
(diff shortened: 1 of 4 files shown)


gjedlicska (approved):

gjedlicska:
so far, that warning was not causing issues in my testings with the EventBus, but the change is legit, so lets see.

PR #3963, opened 2025-02-11
chore(dui3): merge main into dui3

oguzhankoral:
i was expecting more graphql changes but i believe we never used `defaultLogoIndex` in dui3


didimitrie (approved):
Fingers s

PR #3977, opened 2025-02-13
feat(regions): move file uploads and blobs

cdriesler:
## Description & motivation

- Requires #3957


packages/server/modules/workspaces/repositories/projectRegions.ts
```diff
@@ -577,3 +587,66 @@ export const copyProjectWebhooksFactory =
...
+        await deps.targetObjectStorage.client.send(
+          new PutObjectCommand({
+            Bucket: deps.targetObjectStorage.bucket,
+            Key: objectKey,
+            Body: sourceBlob.Body,
+            ContentLength: sourceBlob.ContentLength
+          })
+        )
+      }
+    }
```
alemagio, line 638: verify if we can use the pipe method

alemagio (approved):

packages/server/modules/workspaces/repositories/projectRegions.ts
cdriesler: sources appear to say that the aws sdk handles this for us, but TIL !

PR #3983, opened 2025-02-13
fix(server): consistently handle errors in middleware - Part 1

iainsproat:
## Description & motivation

Express middleware we use may handle errors incorrectly. For example, external middleware may create errors which are not compatible with `BaseError`. Or we may have been inconsistent in catching errors.

This can cause issues when recording error metrics & logs, or can cause Express to suspend the request without logging or metrics.

In this PR we aim to provide a consistent means of handling errors from middleware and routes, ensuring they are correctly recorded in metrics & logs. We also aim to prevent unexpected behaviour from Express by handling errors in a manner it expects.

## Changes:

- Introduces a wrapper function / boilerplate to be wrapped around every Express request handler as so:

    ```
    handleErrors({
        handler: (req, res, next) => { /* actual request handle here */ },
        verbPhraseForErrorMessage: 'doing some request handling',
        defaultErrorType: MyRequestHandlerError
      })
    ```

## Checklist:

### DIFF
--- a/packages/server/modules/shared/middleware/index.ts
+++ b/packages/server/modules/shared/middleware/index.ts
@@ -292,7 +292,7 @@ export const requestBodyParsingMiddlewareFactory =
     //default
     expressJsonBodyParser({ limit: maxRequestBodySize })(req, res, next)
 
-    // expressRawBodyParser calls `next` internally, so we cannot call it again here
+    // expressJsonBodyParser calls `next` internally, so we cannot call it again here
     return
   }
 


packages/server/app.ts
```diff
@@ -433,46 +443,113 @@ export async function init() {
...
+    })
+  )
   if (asyncRequestContextEnabled()) {
     startupLogger.info('Async request context tracking enabled 👀')
   }
 
-  if (process.env.COMPRESSION) {
-    app.use(compression())
+  app.use(
+    handleErrors({
```
fabis94, line 470: instead of having to do all this for every third party middleware, why not just add an extra middleware at the very end of the chain, before defaultErrorHandler & errorMetricsMiddleware, that formats the error in a way that the subsequent error tracking/handling middlewares can handle it?

fabis94: i see that there's custom options for each usage, but it's just a bunch of extra complexity that I'm not sure is worth it. i'd rather have them all format to the same error subtype and be able to forget about this, rather than have to remember to add this wrapper for all future third party middlewares

PR #3988, opened 2025-02-14
Fix: Allow opacity modifiers in TW

Mikehrn:
This will allow us to use the TW opacity modifies like `bg-primary/50`


andrewwallacespeckle (approved):

gjedlicska:
not nitpicking, just askin, shouldn't we move to `oklch` instead of `rgba`?

ppl seam to preffer that nowadays

Mikehrn:
> not nitpicking, just askin, shouldn't we move to `oklch` instead of `rgba`?
> 
> ppl seam to preffer that nowadays

There are a few caveats with using oklch, it requires some further design/FE alignment to get the full power of it, and Figma doesn't support it only with a plugin. But I didn't want to spend too much time on this, I just need to be able to use tailwind opacity modifiers :)

PR #3999, opened 2025-02-17
Alessandro/web 2689 god mode allows project creation in other ppls workspaces

alemagio:
## Description & motivation

- we want god mode to only be able to read data, not write
- for now I did not want to update all the places where the `authorizeResolver` is used but we should
- this is probably not the best way to do it but we should the new `authorizationPipeline`

## Checklist:


gjedlicska:
I think we're going to re-think the whole authorization pipeline in new pricing and packaging, so for now, this is prob enough

gjedlicska (approved):

PR #4000, opened 2025-02-17
feat: Update user tags in Mailchimp when user completes onboarding questions

andrewwallacespeckle:
Adds Mailchimp tag updates when users complete onboarding, capturing their role, plans, and source of discovery. Implementation extends the existing finishOnboarding mutation with error handling to ensure core onboarding flow isn't affected.


packages/frontend-2/lib/auth/composables/onboarding.ts
```diff
@@ -1,7 +1,12 @@
 import { useApolloClient } from '@vue/apollo-composable'
 import { useMixpanel } from '~~/lib/core/composables/mp'
 import { UnsupportedEnvironmentError } from '~~/lib/core/errors/base'
-import type { OnboardingState } from '~~/lib/auth/helpers/onboarding'
+import type {
```
Mikehrn, line 4: Maybe these types should be in the defined here instead, so both BE and FE can validate it's actually part of the enum rather than just a random string. It will probably require you to move these consts to shared

```
input OnboardingCompletionInput {
  role: String
  plans: [String!]
  source: String
}
```

gjedlicska:
With onboarding we do not need to consider the users's marketing email preferences right?

If not, lgtm

packages/server/modules/core/graph/resolvers/users.ts
```diff
@@ -266,9 +267,9 @@ export = {
           const { listId } = getMailchimpOnboardingIds()
 
           await updateMailchimpMemberTags(user, listId, {
-            role: args.input?.role || undefined,
-            plans: args.input?.plans || undefined,
-            source: args.input?.source || undefined
+            role: args.input?.role as OnboardingRole | undefined,
```
gjedlicska, line 270: lets double check these gql / shared types, we should not be casting

packages/server/modules/auth/services/mailchimp.ts
```diff
@@ -72,21 +72,22 @@ async function updateMailchimpMemberTags(
   initializeMailchimp()
   const subscriberHash = md5(user.email.toLowerCase())
 
-  // First ensure user is in the audience
-  await addToMailchimpAudience(user, listId)
+  // Check if user is already in audience (meaning they consented to marketing emails)
+  try {
+    await mailchimp.lists.getListMember(listId, subscriberHash)
+  } catch (e) {
+    return
```
gjedlicska, line 79: this should be an actual error no?
we do expect the user to be added to the onboarding list

gjedlicska (changes requested):

gjedlicska:
we can take a look at the types issue on monday @andrewwallacespeckle

andrewwallacespeckle:
@gjedlicska I have made those requested changes. I've also tested with both consent and no consent, can confirm that it is working as expected.

gjedlicska (approved):

PR #4008, opened 2025-02-17
feat(workspaces): remove regular workspace metrics tracking

gjedlicska:
This removes the regular emit of workspace.metrics event for all workspaces. It has done its job, getting all existing ws states into mixpanel, so its not needed any more.


cdriesler (approved):
adios

PR #4011, opened 2025-02-18
Feat: Add projects subscription to dashboard

Mikehrn:
Moved the UserProjects subscription to a composable so it can be used both on the projects and the dashboard page, didn't really modify the original code just shifted it around. Moved dashboard to a component, that we can remove once the index will be changed to the new, non-dashboard approach (whatever that may be)


packages/frontend-2/lib/user/graphql/subscriptions.ts
fabis94, line 1: empty file

fabis94 (changes requested):
1 minor issue

packages/frontend-2/lib/user/graphql/subscriptions.ts
Mikehrn: Whoops, fixed :)

fabis94 (approved):

PR #4014, opened 2025-02-18
chore(server): refactor activityStream invocations - batch #7 - streams

fabis94:
last batch of activityStream refactors


packages/server/modules/core/services/projects.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { Project } from '@/modules/core/domain/streams/types'
 import { RegionalProjectCreationError } from '@/modules/core/errors/projects'
 import { StreamNotFoundError } from '@/modules/core/errors/stream'
+import { ProjectVisibility as GqlProjectVisibility } from '@/modules/core/graph/generated/graphql'
```
gjedlicska, line 15: graphql types should not be imported into services

gjedlicska (changes requested):
just needs a quick service vs gql type fix

gjedlicska:
now the gql type is in the domain... but lets not get bogged down by this rn.

PR #4023, opened 2025-02-18
feat(workspaces): user workspace join requests

alemagio:
## Checklist:


packages/server/modules/workspaces/graph/resolvers/workspaceJoinRequests.ts
```diff
@@ -74,6 +76,33 @@ export default {
       return await ctx.loaders.workspaces!.getWorkspace.load(parent.workspaceId)
     }
   },
+  User: {
+    workspaceJoinRequests: async (parent, args) => {
```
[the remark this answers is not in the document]
fabis94: What do you mean? The counterparts should definitely exist, otherwise self-hosters will get strange non-human readable errors (instead of "Workspaces not supported" something like "Unexpected null value returned for non-null field")

packages/server/modules/shared/services/paginatedItems.ts
```diff
@@ -23,6 +23,13 @@ export const getPaginatedItemsFactory =
     getTotalCount: (args: Omit<TArgs, 'cursor' | 'limit'>) => Promise<number>
   }) =>
   async (args: TArgs): Promise<Collection<T>> => {
+    if (args.limit === 0) {
+      return {
+        cursor: null,
+        totalCount: 0,
```
fabis94, line 29: that's not what I meant, the totalCount should still be calculated. the point of passing in limit===0 is to optimize execution - skip items, but still get the total count

packages/server/modules/workspaces/graph/resolvers/workspaceJoinRequests.ts
```diff
@@ -33,117 +36,160 @@ import { WorkspaceJoinRequestGraphQLReturn } from '@/modules/workspacesCore/help
...
+                })
+              }
             })
+          return await approveWorkspaceJoinRequest({
+            userId: args.input.userId,
+            workspaceId: args.input.workspaceId
+          })
+        },
+        deny: async (_parent, args) => {
+          const denyWorkspaceJoinRequest = commandFactory<DenyWorkspaceJoinRequest>({
```
fabis94, line 161: I don't get how commandFactory works. It seems to wrap things in a transaction, but the transaction doesn't actually seem to be used in any way.

fabis94 (changes requested):
left some comments

packages/server/modules/workspaces/graph/resolvers/workspaceJoinRequests.ts
fabis94: image
shouldn't the trx be passed in place of db like so? cc @gjedlicska

fabis94: i'm also concerned if just wrapping all of the "command's" DB operations in a trx is something we even want to do. you might want to batch inserts/updates in a trx, while getting selects without one

packages/server/modules/shared/services/paginatedItems.ts
alemagio: you're right, my bad

packages/server/modules/workspaces/graph/resolvers/workspaceJoinRequests.ts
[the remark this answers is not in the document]
alemagio: What I mean is that we are not doing it consistently so what I would do is create a dedicated PR to add all the missing ones, anyway, I pushed a change for these resolvers

packages/server/modules/workspaces/graph/resolvers/workspaceJoinRequests.ts
alemagio: Your are seeing this as a changed line because I added the FF check which moves everything to the right, however this is not part of this PR.

fabis94 (approved):

PR #4043, opened 2025-02-21
refactor(fe): Improve webhooks ux

andrewwallacespeckle:
- Reorder items in CreateOrEditDialog so triggers is higher, meaning there is less chance of the dropdown overflowing the dialog
- Use new FormSelectMulti
- Prevent the FormSelectMulti from jumping to the top of the option list on each selection
- Add a mapping function to map the old webhooks names to their new equivalent, without affecting functionality

### DIFF
--- a/packages/ui-components/src/components/form/select/Multi.vue
+++ b/packages/ui-components/src/components/form/select/Multi.vue
@@ -133,6 +133,7 @@
                 </label>
 
                 <div
+                  ref="optionsContainer"
                   class="overflow-auto simple-scrollbar max-h-60 gap-1 flex flex-col"
                 >
                   <div v-if="isAsyncSearchMode && isAsyncLoading" class="px-1">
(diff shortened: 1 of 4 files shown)


andrewwallacespeckle:
@Mikehrn @fabis94 
Jonathon raised that we are using the old naming in webhooks (eg streams, branches).

I added a mapping function to FE to convert these to the new names, but did it so it wouldn't affect existing functionality with webhooks, it is purely a label update. 

Can you double check this one for me, I'm right in saying this shouldn't affect existing webhooks, or cause any issues elsewhere outside of FE?

Mikehrn (approved):

packages/ui-components/src/components/form/select/Multi.vue
```diff
@@ -763,16 +764,27 @@ const isSelected = (item: SingleItem) => {
   return wrappedValue.value.some((v) => itemKey(v) === itemKey(item))
 }
 
+const optionsContainer = ref<HTMLDivElement | null>(null)
```
Mikehrn, line 767: Is this top override library behavior?

fabis94 (approved):

PR #4061, opened 2025-02-24
fix(server/logging): some subscription errors should be logged at warning level

iainsproat:
## Description & motivation

We had a few `GraphQLErrors` logged as errors. These were due to client/user input issues so should not be `error` level. However, they may indicate a malformed query in a client that we, Speckle, own and which should be fixed. Therefore they will be logged as `warn` level.

## Changes:

- add a `shouldLogAsWarnLevel` to filter some specific GraphQLError messages
- update `shouldLogAsInfoLevel` to include some specific GraphQLError messages
- refactor to move some code out of the main `app.ts` and into the logging directory, to make it manageable & readable.

## Checklist:

### DIFF
--- a/packages/server/logging/graphqlError.ts
+++ b/packages/server/logging/graphqlError.ts
@@ -32,6 +32,12 @@ export const shouldLogAsWarnLevel = (err: unknown): boolean => {
     )
   )
     return true
+  if (
+    /Cannot query field\s"([^\s]+)"\son type\s"([^\s]+)"\. Did you mean\s"([^\s]+)"\?/.test(
+      err.message
+    )
+  )
+    return true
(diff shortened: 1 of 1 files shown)


packages/server/logging/graphqlError.ts
```diff
@@ -20,3 +21,25 @@ export const shouldLogAsInfoLevel = (err: unknown): boolean => {
 
   return err instanceof ApolloError
 }
+
+export const shouldLogAsWarnLevel = (err: unknown): boolean => {
+  if (!(err instanceof GraphQLError)) return false
+
+  if (err.message.startsWith('Cannot return null for non-nullable field')) return true
```
fabis94, line 28: i would think that query/variable validation errors have a specific error subtype/code that can just be checked

fabis94 (approved):

PR #4064, opened 2025-02-25
chore: move common workspace plan/feature config to @speckle/shared

fabis94:
## Checklist:


packages/server/codegen.yml
```diff
@@ -8,6 +8,7 @@ generates:
       - 'typescript'
       - 'typescript-resolvers'
     config:
+      enumsAsConst: true
```
fabis94, line 11: TS native Enums kinda suck cause values typed as those enums don't accept the same exact values, just because they're not using the enum. So this won't work:

```ts
const Foo {
 A = 'a'
}

// ERROR: Even tho the value fits the type, I used a literal string 'a', not Foo.A, and this breaks it
const test: Foo = 'a' 
```

We tend to do mapping between GQL enums and our own domain types in a lot of places, so I decided to adjust codegen config to not use native TS enums and generate const objects instead

```ts
const Foo = {
  A: 'a'
} as const

type Foo = (typeof Foo)[keyof typeof Foo]

const test: Foo = 'a' // works!
```

packages/server/assets/gatekeeper/typedefs/gatekeeper.graphql
```diff
@@ -1,7 +1,3 @@
-extend type Query {
-  workspacePricingPlans: JSONObject!
```
fabis94, line 2: according to Apollo Studio hasn't been used by anyone since Jan 14th. the FE def doesn't use it

packages/server/codegen.yml
gjedlicska: def makes it a lot better, i'm still not behind importing these constants into the business logic parts of the app.

packages/server/modules/gatekeeperCore/domain/billing.ts
```diff
@@ -75,7 +23,7 @@ export type PaidWorkspacePlan = BaseWorkspacePlan & {
 }
 
 export type TrialWorkspacePlan = BaseWorkspacePlan & {
-  name: TrialWorkspacePlans
+  name: TrialEnabledPaidWorkspacePlans
```
gjedlicska, line 26: These also constrain the statuses each workspace plan can be on, maybe that is also usefull in the frontend.
cc @Mikehrn

gjedlicska (changes requested):
Backend looks good.
Lets remove the lodash templating from the shared naming for now. I don't think its needed.
Shared constants also look good.
Meaningful renames are appreciated.

Frontend review is deferred to @Mikehrn

packages/server/modules/gatekeeperCore/domain/billing.ts
fabis94: The FE is likely going to work with the WorkspacePlan GQL type directly. Casting it to these would require helpers/type guards (cause the types don't fit 1:1) and the FE use cases for these aren't immediately obvious to me. We can always move them to shared later, but if Mike sees a use for these, I can move em

Mikehrn (commented):
Overall looks good and everything seems to work :) Only in the security and data regions pages seem to break, probably small change that needs to be made there 

Screenshot 2025-02-26 at 16 06 45

Screenshot 2025-02-26 at 16 06 38

packages/server/modules/gatekeeperCore/domain/billing.ts
Mikehrn: This will be any obsolete in a few weeks right, as we will abandon the concept of trials existing?

fabis94:
@Mikehrn I'm not getting those errors, they seem to be backend errors coming from the GQL API. Can you check which GQL fields trigger the error?

Mikehrn (approved):

PR #4086, opened 2025-02-27
feat(workspaces): track last visited workspace and project

cdriesler:
## Description & motivation

- We'd like to navigate users back to the last workspace/project they were in
- There's some complexity wrt self-hosting and not having workspaces, or projects outside of workspaces

## Changes:

- Introduces `User.activeWorkspace` and `User.activeProject` that can be set independently. Should allow FE to express all cases in a flexible way. For example:
  - On workspace visit, `activeWorkspace = slug` & `activeProject = null`
  - On non-workspace project visit, `activeWorkspace = null` & `activeProject = id`
  - On self-hosted, nobody knows
- Tests for happy path and null behaviors


packages/server/modules/core/domain/users/operations.ts
```diff
@@ -256,3 +256,12 @@ export type AdminGetInviteList = (args: CollectionQueryArgs) => Promise<{
...
   items: ServerInviteGraphQLReturnType[]
 }>
+
+export type SetUserActiveWorkspace = (args: {
+  userId: string
+  workspaceSlug: string | null
+}) => Promise<void>
+export type SetUserActiveProject = (args: {
+  userId: string
+  projectId: string | null
```
Mikehrn, line 266: For projects we don't need the ID, it's just "you are in the projects world", for reference this is how the state is atm in the FE. So maybe this could just be a boolean?

```
  useState<{
    activeWorkspaceSlug: string | null
    isProjectsActive: boolean
  }>('navigation-state', () => ({
    activeWorkspaceSlug: null,
    isProjectsActive: false
  }))
  ```

File: packages/server/modules/core/dbSchema.ts [hunk 2]
```diff
@@ -288,7 +288,14 @@ export const StreamFavorites = buildTableHelper('stream_favorites', [
 export const UsersMeta = buildMetaTableHelper(
   'users_meta',
   ['userId', 'key', 'value', 'createdAt', 'updatedAt'],
-  ['isOnboardingFinished', 'foo', 'bar', 'onboardingStreamId'],
+  [
+    'isOnboardingFinished',
+    'foo',
```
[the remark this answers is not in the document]
cdriesler: I found them silly too but didn't want to risk them being secretly important. They have been dealt with. 🔪

packages/server/modules/core/domain/users/operations.ts
Mikehrn: Let's catch up on this tomorrow :) The problem is that you can be in the "legacy projects world" without being in a specific legacy project. Not really sure what kind of new concept this introduces? We do something similar for the workspace wizard, where we have a JSON object for the state that the FE "controls". But, I think what you are proposing will work

Mikehrn (approved):
From my side, this will do what I need 🙂

PR #4107, opened 2025-03-04
refactor(fe): New workspace actions route to avoid slug conflicts

andrewwallacespeckle:
Move `workspaces/join` and `workspaces/join` to `workspaces/actions/join` and `workspaces/actions/create`


Mikehrn (dismissed):

packages/shared/src/workspaces/errors/index.ts
```diff
@@ -42,8 +40,4 @@ export function validateWorkspaceSlug(slug: string): void {
   if (!VALID_SLUG_BOUNDARY_REGEX.test(slug)) {
     throw new InvalidWorkspaceSlugError('Short ID cannot start or end with a hyphen.')
   }
-
-  if (blockedSlugs.includes(slug)) {
```
fabis94, line 46: you still probably want to skip 'actions' from being a slug, no?

andrewwallacespeckle: Screenshot 2025-03-04 at 13 58 53
Done

fabis94 (approved):

PR #4115, opened 2025-03-04
Feat: Base for new billing settings

Mikehrn:
This adds a base for new workspace billing page. This allows to update content and functionality as they get implemented.
- Created a new composable with workspace plan logic, this we can use in various places as it will be relevant in multiple components in the future
- Without the FF the old page is shown, it's split it up in two component, one old and one new. Old is to be removed post migration, which is why I didnt spend any effort on implementing the new composable there. 
- Pricing table not included, as the content is still TBD
- I tried to cover various scenarios, but it's hard without the actual plan data, so need to test and tweak once new plans are fully implemented

Screenshot 2025-03-04 at 22 13 44

Screenshot 2025-03-04 at 22 13 56


packages/frontend-2/components/settings/workspaces/billing/Summary.vue
```diff
@@ -0,0 +1,75 @@
...
+          <template v-else>Bill</template>
+        </h3>
+        <p class="text-heading-lg text-foreground inline-block">
+          {{ totalCostFormatted }}
+          <span v-if="isPurchasablePlan">per {{ billingInterval }}</span>
+        </p>
+        <NuxtLink
+          v-if="showBillingPortalLink"
+          class="text-body-xs text-foreground-2 underline hover:text-foreground cursor-pointer mt-1"
+          @click="billingPortalRedirect('')"
```
andrewwallacespeckle, line 29: Just checking that this is an empty string because it's not ready yet, and not a mistake

andrewwallacespeckle (dismissed):

andrewwallacespeckle (approved):

andrewwallacespeckle (approved):

PR #4117, opened 2025-03-05
feat(server): prevent viewer seat from becoming project contributor/owner + extra fixes

fabis94:
## Checklist:


packages/server/modules/core/services/projects.ts
```diff
@@ -61,13 +62,15 @@ export const createNewProjectFactory =
             const replicatedProject = await getProject({ projectId })
             if (!replicatedProject) throw new StreamNotFoundError()
           },
-          { maxAttempts: 10 }
+          { maxAttempts: 10, delay: isTestEnv() ? 1000 : undefined }
```
fabis94, line 65: default settings were not enough locally, I was getting random failures here

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -482,6 +495,60 @@ const onWorkspaceCreatedFactory =
...
 
+const blockInvalidWorkspaceProjectRoleUpdatesFactory =
+  (deps: {
+    getStream: GetStream
+    getWorkspaceRoleForUser: GetWorkspaceRoleForUser
+    getWorkspaceUserSeat: GetWorkspaceUserSeat
+    getWorkspacePlan: GetWorkspacePlan
+  }) =>
+  async ({ payload }: EventPayload<typeof ProjectEvents.PermissionsBeingAdded>) => {
+    const project = await deps.getStream({ streamId: payload.projectId })
```
fabis94, line 506: moved chuck's project role protection logic here

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -482,6 +495,60 @@ const onWorkspaceCreatedFactory =
...
+
+    if (
+      currentWorkspaceRole?.role === Roles.Workspace.Guest &&
+      payload.role === Roles.Stream.Owner
+    ) {
+      // Workspace guests cannot be project owners
+      throw new WorkspaceInvalidRoleError('Workspace guests cannot be project owners.')
+    }
+
+    // Workspace seat checks
```
fabis94, line 533: + Added the new seat checks below

packages/server/modules/gatekeeper/domain/billing.ts
```diff
@@ -173,7 +173,12 @@ export type ReconcileSubscriptionData = (args: {
   applyProrotation: boolean
 }) => Promise<void>
 
-export type WorkspaceSeatType = 'viewer' | 'editor'
+export const WorkspaceSeatType = <const>{
+  Viewer: 'viewer',
+  Editor: 'editor'
+}
+export type WorkspaceSeatType =
+  (typeof WorkspaceSeatType)[keyof typeof WorkspaceSeatType]
```
alemagio, line 181: I would like to avoid these kind of changes.
I thought we agreed that union string types are fine

cc
@gjedlicska

alemagio (approved):

packages/server/modules/gatekeeper/domain/billing.ts
fabis94: it's the same type it was before, except now you can also use it as a runtime enum. I don't see what the problem is with that

fabis94: i think magic string literals are bad practice - 1) the autocomplete only works when you're putting them into a constant/param that is already typed as the expected type. if you're just doing `const a = ...` and start typing the string, its not gonna know `a`'s expected type and you won't get autocomplete. don't have that issue with enums 
2) duplicating strings over and over again makes it more difficult to change the value in the future

that said, i'm ok with someone here disagreeing - that's why there's both options available. you can still do the magic strings if you want, but at least there's the option to use the enum if you prefer that

PR #4120, opened 2025-03-05
feat(fe): Tutorials page

andrewwallacespeckle:
image


packages/frontend-2/components/tutorials/Card.vue
```diff
@@ -6,7 +6,7 @@
       <NuxtImg
         :src="tutorialItem.image"
         :alt="tutorialItem.title"
-        class="h-32 w-full object-cover"
+        class="aspect-video w-full object-cover"
```
andrewwallacespeckle, line 9: It seems that all the images are this aspect ratio, and it means the borders that are in the images are not clipped. 

It does make them slightly taller on Dashboard though, if this is a problem I can add a prop or something.

packages/frontend-2/lib/dashboard/helpers/tutorials.ts
```diff
@@ -1,52 +1,80 @@
...
+  [TutorialProduct.Rhino]: 'Rhino',
+  [TutorialProduct.AutoCad]: 'AutoCad',
+  [TutorialProduct.Automate]: 'Automate',
+  [TutorialProduct.Blender]: 'Blender',
+  [TutorialProduct.SketchUp]: 'SketchUp',
+  [TutorialProduct.Unreal]: 'Unreal',
+  [TutorialProduct.Unity]: 'Unity'
+})
+
+export const tutorialItems: TutorialItem[] = [
```
Mikehrn, line 16: There has been a new one added last month, maybe we can add it in there directly

packages/frontend-2/components/tutorials/Card.vue
Mikehrn: Nice, overall looks good, but maybe make it 4 cols? They seems quite big with 3 cols per row

packages/frontend-2/components/connectors/Page.vue
```diff
@@ -38,6 +38,7 @@
             :items="categories"
             size="base"
             color="foundation"
+            clearable
```
Mikehrn, line 41: Small papercut, because it's min width it jumps when the clear button is shown, you can change it to a fixed width instead of max

benjaminvo (changes requested):
Yes, overall looks good!

1. Agree with Mike that the 3-column grid makes the cards too big
2. I'd remove search and the product filter for now. Bring this back when we can render all tutorials in the app. At the current count it doesn't make much sense.
3. Add a "View all tutorials" button at the bottom. Probably centered.

packages/frontend-2/lib/dashboard/helpers/tutorials.ts
```diff
@@ -42,11 +48,5 @@ export const tutorials: TutorialItem[] = [
     image:
       'https://cdn.prod.website-files.com/66c31b5a50432200dc753cc4/66f9c50ac50a28b58fe0e10b_66e047f505114bd4a6854b6a_216-blocks-to-families%25400.5x.png',
     url: 'https://www.speckle.systems/tutorials/new-in-2-16-block-to-family-conversion'
-  },
```
andrewwallacespeckle, line 45: image

This seems like an old tutorial, so I removed it and added the new one instead.

Mikehrn (approved):

Mikehrn (approved):

benjaminvo (approved):

PR #4125, opened 2025-03-06
feat: remove project discoverability concept from UI

fabis94:
considering that it may come back at some point, i didn't fully eradicate the concept, just removed it from the UI and made it so that the API doesn't return the discoverability value


packages/server/assets/core/typedefs/projects.graphql
```diff
@@ -12,6 +12,14 @@ enum ProjectVisibility {
   PUBLIC
 }
 
+"""
+Visbility without the "discoverable" option
+"""
+enum SimpleProjectVisibility {
```
fabis94, line 18: Still leaving the old enum and accepting it in create/update for backwards compatibility. If consumers send in "discoverable"/"public" as a value, it'll just be converted to UNLISTED behind the scenes.

Mikehrn (dismissed):

andrewwallacespeckle (approved):
Frontend looks good

cdriesler (approved):
ditto for BE

PR #4126, opened 2025-03-06
refactor(fe): Cominbed onboarding middlewares

andrewwallacespeckle:
Combined:
- email verification
- onboarding
- require workspace

middlewares into a single file to aid with readability and performance. 

I have called this onboarding as it best captures all of the different tasks. I now refer to the questions as segmentation rather than onboarding to make it clear which is which.


packages/frontend-2/middleware/004-onboarding.global.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  const isEmailEnabled = serverInfoData?.serverInfo.configuration.isEmailEnabled
+
+  // 1. Email verification redirect
+  // Self hosters may not have emails enabled, so we skip the redirect
+  const isVerifyEmailPage = to.path === verifyEmailRoute
+  const hasUnverifiedEmails = userData.activeUser.emails.some(
+    (email) => !email.verified
+  )
+  if (isEmailEnabled) {
```
Mikehrn, line 52: I think this probably can just be something like this, a bit shorter:

```
// Skip email verification for self-hosters with email disabled & users with verified emails
if (isEmailEnabled && hasUnverifiedEmails) {
  // Redirect to verification page if not already there
  if (!isVerifyEmailPage) {
    return navigateTo(verifyEmailRoute)
  }
  
  // Don't run any other checks if the user has unverified emails
  return
}
```

packages/frontend-2/middleware/004-onboarding.global.ts
```diff
@@ -0,0 +1,113 @@
...
+    return navigateTo(onboardingRoute)
+  }
+  if (isGoingToSegmentation && isSegmentationFinished) {
+    return navigateTo(homeRoute)
+  }
+  // Don't run any other checks if the user has not finished the onboarding process
+  if (!isSegmentationFinished) return
+
+  // 3. Workspace join/create redirect
+  const isWorkspaceNewPlansEnabled = useWorkspaceNewPlansEnabled()
```
Mikehrn, line 77: Lets make sure we make a note to remove these `isWorkspaceNewPlansEnabled ` checks post-release

Mikehrn (approved):

Mikehrn (approved):

PR #4149, opened 2025-03-10
feat: removes docs link, adds usecases quickstart card

didimitrie:
Before: 

image

After: 

image

Notes:

- Rationale: docs are a broken window currently and we don't want users going into them. We replace them with links to our shiny new use case pages. 
- I know it's a bit heavy and noisy - i would've made the cards more graphic, but automate use case does not have a featured image, so tough luck
- Would've linked to one landing page for all usecases, but we don't have that - so unfolded into all current use cases. 
- Tracking would be nice, but yolo.


Mikehrn:
@didimitrie Should we either make this a separate page, or just remove the docs link? This page will be dead in like 3 weeks. On my laptop I can't even see my latest projects anymore 😅

Screenshot 2025-03-10 at 13 52 20

PR #4167, opened 2025-03-12
Alessandro/web 2741 subscriptions to upgrade for new plan types

alemagio:
## Description & motivation

- separate old and new paths
- new path is automatically chosen if the workspace is currently in a new plan
- create new path to upgrade paid plans

## Checklist:


packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
```diff
@@ -0,0 +1,381 @@
...
+import {
+  PaidWorkspacePlans,
+  PaidWorkspacePlansNew,
+  throwUncoveredError,
+  WorkspacePlanBillingIntervals,
+  xor
+} from '@speckle/shared'
+import { cloneDeep } from 'lodash'
+
+export const upgradeWorkspaceSubscriptionFactoryOld =
```
alemagio, line 35: This one is a copy/paste from main

fabis94: another downside of copy pasting the new implementation - very hard to review. i have to keep both the old and new versions side by side and go line by line to see what actually differs between them and how

packages/server/modules/gatekeeper/graph/resolvers/index.ts
```diff
@@ -301,21 +306,46 @@ export = FF_GATEKEEPER_MODULE_ENABLED
...
+                  getWorkspacePlanPriceId,
+                  getWorkspacePlanProductId,
+                  upsertWorkspacePlan: upsertPaidWorkspacePlanFactory({ db }),
+                  updateWorkspaceSubscription: upsertWorkspaceSubscriptionFactory({
+                    db
+                  })
+                })
+          await upgradeWorkspaceSubscription({
+            workspaceId,
+            targetPlan: workspacePlan as PaidWorkspacePlansNew, // This should not be casted and the cast will be removed once we will not support old plans anymore
```
fabis94, line 346: why is this cast necessary? both the old & new upgrade implementation expects targetPlan to be `PaidWorkspacePlans` not `PaidWorkspacePlansNew` which should work without the cast, no?

packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
```diff
@@ -0,0 +1,381 @@
...
+    targetPlan: PaidWorkspacePlans
+    billingInterval: WorkspacePlanBillingIntervals
+  }) => {
+    const workspacePlan = await getWorkspacePlan({
+      workspaceId
+    })
+
+    if (!workspacePlan) throw new WorkspacePlanNotFoundError()
+    if (!isNewPlanType(workspacePlan.name) || !isNewPlanType(targetPlan)) {
+      throw new NotImplementedError()
```
fabis94, line 268: In the case of `!isNewPlanType(targetPlan)` this is not a valid throw. It's not that its not implemented, its that its an explicitly unsupported operation and should show an appropriate error just like on line  120

The same issue is in the old implementation function on line 112

packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
```diff
@@ -0,0 +1,381 @@
...
+      if (xor(oldPlanFitsSchema, newPlanFitsSchema)) {
+        throw new WorkspacePlanUpgradeError(
+          'Attempting to switch between incompatible plan types'
+        )
+      }
+    }
+
+    if (isNewPlanType(targetPlan) || isNewPlanType(workspacePlan.name)) {
+      // Needs custom logic below for seats
+      throw new NotImplementedError()
```
fabis94, line 128: all of these NotImplemented throws that were added to show where new plans need logic changes should be removed now that we have the actual new implementation, no?

fabis94 (changes requested):
left some comments/questions

packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
alemagio: The point is that you should not compare the 2, the old one is exactly as it was before.
The new implementation is similar but that's not the rule, you should review the new one on its own.

packages/server/modules/gatekeeper/graph/resolvers/index.ts
alemagio: actually no
the old should expect the old only, and the new should expect the new only.
I found the old implementation changed so that it also accepts the new plans, so I left it as it was.

packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
alemagio: I wanted to leave the old implementation exactly as it was to avoid missing some changes in the CR.

packages/server/modules/gatekeeper/services/subscriptions/upgradeWorkspaceSubscription.ts
fabis94: > The point is that you should not compare the 2
I disagree. Especially as I don't have the full context & experience with this function, it is actually very important for me to compare it with a baseline - the previous implementation

fabis94 (dismissed):

fabis94 (approved):

PR #4170, opened 2025-03-12
fix(shared): ensureError should not bury original message

iainsproat:
## Description & motivation

Currently if the original cause is not an Error it is ignored by `ensureError` and replaced by the fallback message.

This PR ensures that the original cause, if not null or undefined, is included alongside the error message.

## Checklist:


fabis94 (dismissed):
What if `e` isn't JSON serializable? Might be worth serializing it in a try/catch first

fabis94 (approved):

PR #4172, opened 2025-03-12
feat(objectloader): ObjectLoader2: use indexdb as cache and batch download objects

adamhathcock:
Ported to Typescript
Still a package that can be used

Benchmarking:
pausing in v1:
Screenshot 2025-03-19 141736

v1 has memory issues (goes up to 8 gigs)
Screenshot 2025-03-19 141828
v1 download and cache usage with timing
Screenshot 2025-03-19 142050
v1 just from cache:
Screenshot 2025-03-19 142433

v2 with little to no pause:
Screenshot 2025-03-19 143042

v2 hovers around 2.5gigs on my machine:
Screenshot 2025-03-19 144713
v2 with download and cache can be slower:
Screenshot 2025-03-19 144807
v2 just from cache:
Screenshot 2025-03-19 144729

v2 with download pool and adaptive wait for downloads/indexeddb writes
Screenshot 2025-03-20 112711

v2 from cache now:
Screenshot 2025-03-20 110807

also includes https://linear.app/speckle/issue/WEB-2898/objectloader-should-include-caller-defined-headers-in-fetch-requests


packages/objectloader2/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@speckle/objectloader2",
+  "version": "1.0.0",
+  "description": "TODO",
```
fabis94, line 4: update

packages/objectloader2/readme.md
```diff
@@ -0,0 +1,3 @@
+# objectloader2
+
+TODO
```
fabis94, line 3: update

packages/objectloader2/src/BaseDatabase.ts
```diff
@@ -0,0 +1,123 @@
...
+        const db = (event.target as IDBOpenDBRequest).result
+        if (!db.objectStoreNames.contains(storeName)) {
+          db.createObjectStore(storeName, { keyPath: 'id', autoIncrement: true })
+        }
+      }
+
+      request.onsuccess = () => resolve(request.result)
+      request.onerror = () =>
+        reject(
+          request.error instanceof Error
```
fabis94, line 30: use ensureError() from @speckle/shared

fabis94: also - isnt the error supposed to come from the args of this event handler?

fabis94:
no, pino is not browser friendly

packages/objectloader2/src/types.ts
```diff
@@ -0,0 +1,32 @@
+export type CustomLogger = (message?: string, ...optionalParams: unknown[]) => void
```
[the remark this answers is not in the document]
fabis94: no, pino is not browser friendly. we can make this type fit the same function signature so that we can use the pino logger where available ,tho

packages/objectloader2/src/types/types.ts
```diff
@@ -30,3 +32,19 @@ export function chunk<T>(array: T[], size: number): T[][] {
   }
   return result
 }
+
+export function isBase(maybeBase?: unknown): boolean {
+  return (
+    maybeBase !== null &&
+    typeof maybeBase === 'object' &&
```
fabis94, line 39: @adamhathcock I've just pushed in a commit adding [lodash](https://lodash.com/docs/4.17.15) as a dep. Its a common utility lib that will help you simplify some of these checks (e.g. isObjectLike, isString).

I noticed you already started creating your own versions for some of these in helpers, you can get rid of those and use lodash instead.

Just ensure you import from `#lodash` not `lodash` to ensure the proper ESM or CJS version is imported depending on the environment.

fabis94: Also, if you need to start getting creative with more complicated TS types, I suggest looking into https://github.com/sindresorhus/type-fest it has a lot of useful utility types

adamhathcock: I can do that....I was thinking I didn't want lodash anymore as a thing.  I'll clean up tomorrow

packages/objectloader2/src/types/types.ts
```diff
@@ -30,3 +32,19 @@ export function chunk<T>(array: T[], size: number): T[][] {
   }
   return result
 }
+
+export function isBase(maybeBase?: unknown): boolean {
```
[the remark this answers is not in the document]
adamhathcock: I was doing that often so I made `asBase` as a function to share.  Maybe I'm still misunderstanding....I need schooling on best practices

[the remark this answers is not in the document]
fabis94: @adamhathcock If `isBase` is a proper type guard w/ a `arg is type` return, `asBase` is redundant. You already get the same result of maybeBase being cast as `Base` in the execution branch where `isBase()` returned true

packages/objectloader2/.npmignore
```diff
@@ -0,0 +1,4 @@
+coverage
+node_modules
+.eslintcache
+dist
```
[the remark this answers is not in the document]
adamhathcock: Installled it but I have idea what to do next 😬

packages/objectloader2/src/BaseDownloader.ts
```diff
@@ -0,0 +1,106 @@
...
+  static async downloadBatch(
+    idBatch: string[],
+    url: string,
+    headers: HeadersInit,
+    results: AsyncBuffer<Item>
+  ): Promise<void> {
+    const response = await fetch(url, {
+      method: 'POST',
+      headers: { ...headers, 'Content-Type': 'application/json' },
+      body: JSON.stringify({ objects: JSON.stringify(idBatch) })
```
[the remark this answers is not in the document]
adamhathcock: probably not as I want it to fail spectacularly

packages/objectloader2/src/operations/downloader.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+import { vi } from 'vitest'
+import AsyncGeneratorQueue from '../helpers/asyncGeneratorQueue.js'
+import { Item } from '../types/types.js'
+import Downloader from './downloader.js'
+import CacheDatabase from './database.js'
+
+const fetchMocker = createFetchMock(vi)
+
+// sets globalThis.fetch and globalThis.fetchMock to our mocked version
+fetchMocker.enableMocks()
```
fabis94, line 12: why not just make fetch injectable through the constructor?

packages/objectloader2/src/operations/downloader.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+describe('downloader', () => {
+  beforeEach(() => {
+    fetchMocker.resetMocks()
+  })
+  test('download batch', async () => {
```
fabis94, line 18: its good practice to make sure the final test titles are human readable sentences. the resulting string ends up being a combination of ancestor describe titles and the test's title.

so if you have a describe('downloader') and a test('downloads batch successfully'), that ends up building a nice human readable description for the test: downloader downloads batch successfully

packages/objectloader2/src/operations/downloader.spec.ts
adamhathcock: no idea....I don't know what's best practice anymore.  This patching was what was hot back in the day. 

I was going to ask you to a meeting to show you what I did to test things.  Sounds like you're suggesting a more typed/DI approach of wrapping fetch in a more basic type and mocking it to test ability

packages/objectloader2/src/operations/database.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, test } from 'vitest'
+import CacheDatabase from './database.js'
+import indexeddb from 'fake-indexeddb'
+import { Item } from '../types/types.js'
+
+globalThis.indexedDB = indexeddb
```
fabis94, line 6: Maybe indexedDb should be injected through constructor options too? It could default to globalThis.indexedDB, but could also be overridden through ctor options

packages/objectloader2/src/operations/downloader.spec.ts
fabis94: you can still mock it, I don't think you necessarily need to wrap it in anything, just inject it in through DI instead of mocking the global fetch import

packages/objectloader2/src/operations/database.ts
```diff
@@ -0,0 +1,198 @@
...
+
+    // Initialize
+    await this.safariFix()
+    this._cacheDB = await this.openDatabase(
+      CacheDatabase._databaseName,
+      CacheDatabase._storeName
+    )
+    return true
+  }
+  async getItems(
```
fabis94, line 82: pretty strange to have a `getItems(): void` function. so it retrieves them but doesn't return them? maybe if its useful, this can be converted to an async generator (cause results come in one at a time) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncGenerator#async_generator_iteration

packages/objectloader2/src/operations/database.ts
```diff
@@ -0,0 +1,198 @@
...
+    const base = await getBase
+    if (base === undefined) return undefined
+    if (isBase(base)) {
+      return { baseId, base }
+    } else {
+      throw new ObjectLoaderRuntimeError(`${baseId} is not a base`)
+    }
+  }
+
+  static async cacheSaveBatch(batch: Item[], cacheDB: IDBDatabase): Promise<void> {
```
fabis94, line 154: why are these 2 functions static while everything else isn't?

packages/objectloader2/src/operations/downloader.ts
```diff
@@ -0,0 +1,155 @@
...
+    this._database = database
+    this._results = results
+
+    this._serverUrl = serverUrl
+    this._streamId = streamId
+    this._objectId = objectId
+    this._token = token
+    this._options = {
+      ...{
+        fetch: (...args) => window.fetch(...args),
```
fabis94, line 40: 1. you shouldn't reference `window` anywhere, this code has to be Node compatible
2. y not just `fetch: globalThis.fetch`

packages/objectloader2/src/operations/downloader.ts
```diff
@@ -0,0 +1,155 @@
...
+    if (!response.body) {
+      throw new Error('ReadableStream not supported or response has no body.')
+    }
+
+    const reader = response.body.getReader()
+    const decoder = new TextDecoder()
+    let buffer = '' // Temporary buffer to store incoming chunks
+
+    while (true) {
+      const { done, value } = await reader.read()
```
fabis94, line 113: sometimes the requests objectloader does fail (e.g. 500 or 429 status code) and the response is no longer JSON but HTML. in those scenarios objectloader throws a confusing error about not being able to parse `<` or something. would be nice if that case would be more gracefully handled

packages/objectloader2/src/operations/database.ts
```diff
@@ -0,0 +1,198 @@
+import BatchingQueue from '../helpers/batchingQueue.js'
+import Queue from '../helpers/queue.js'
+import { ObjectLoaderRuntimeError } from '../types/errors.js'
+import { CustomLogger, Item, isBase } from '../types/types.js'
+import { ensureError, isSafari } from '@speckle/shared'
+import { BaseDatabaseOptions } from './options.js'
+import { ICache } from './interfaces.js'
+
+export default class CacheDatabase implements ICache {
+  private static _databaseName: string = 'speckle-cache'
```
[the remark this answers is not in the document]
adamhathcock: looks so weird

packages/objectloader2/src/operations/database.ts
```diff
@@ -0,0 +1,198 @@
...
+        reject(ensureError(e, 'Failed to open IndexedDB database'))
+    })
+  }
+
+  private supportsCache(): boolean {
+    return !!(this._options.enableCaching && this._options.indexedDB)
+  }
+
+  private async setupCacheDb(): Promise<boolean> {
+    if (this._cacheDB !== undefined && !this.supportsCache()) return false
```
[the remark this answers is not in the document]
adamhathcock: I thought about making a factory class but this is better I guess

packages/objectloader2/src/operations/database.ts
adamhathcock: usually I get suggestions to have things as static when they are but since you have to address static members directly then it's uglier

packages/objectloader2/src/operations/downloader.ts
adamhathcock: I was going around in circles to make this work....I think my example just used window...can use the other I guess...will need to understand the difference though

packages/objectloader2/src/operations/objectLoader2.ts
```diff
@@ -0,0 +1,92 @@
...
+      return
+    }
+    yield rootItem
+    if (!rootItem.base.__closure) return
+    const getPromise = this._database.getItems(
+      Object.keys(rootItem.base.__closure),
+      this._gathered,
+      this._downloader
+    )
+    for await (const item of this._gathered.consume()) {
```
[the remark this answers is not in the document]
adamhathcock: I'd have to talk to you about possible approaches.  It is unintuitive but I can't currently think of a better way.  

Basically I want a callback to queue an item for download then caching or returning.  It's rough to generate it.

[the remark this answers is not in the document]
fabis94: lmk if you need help

packages/objectloader2/src/operations/downloader.ts
adamhathcock: In those situations, the validateResponse should trigger an error and never get this far

fabis94:
what Iain said + align on native #private props everywhere plz

packages/objectloader2/src/operations/database.ts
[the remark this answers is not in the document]
adamhathcock: actually, I don't want to generate values because the places I use this kind of callback I don't want to pause execution.  I can explain in person better.

packages/objectloader2/src/operations/indexed-database.ts
```diff
@@ -158,6 +165,7 @@ export default class IndexedDatabase implements Cache {
     }
     await Promise.all(promises)
     transaction.commit()
+    console.log('Saved ' + batch.length)
```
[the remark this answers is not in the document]
adamhathcock: I meant to remove this!

packages/objectloader2/src/helpers/batchedPool.ts
```diff
@@ -0,0 +1,73 @@
...
+        if (duration > this.interval) {
+          this.interval = Math.min(this.interval * 1.5, this.maxInterval) // Increase if slow
+        } else {
+          this.interval = Math.max(this.interval * 0.8, this.minInterval) // Decrease if fast
+        }
+      }
+      if (wait) {
+        await this.delay(this.interval)
+        //waited so reset
+        this.interval = this.baseInterval
```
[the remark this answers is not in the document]
adamhathcock: you're right.  This actually should be a let within each function all as it changes

packages/objectloader2/src/operations/objectLoader2.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  #logger: CustomLogger
+
+  #database: Cache
+  #downloader: Downloader
+
+  #gathered: AsyncGeneratorQueue<Item>
+
+  constructor(
+    serverUrl: string,
```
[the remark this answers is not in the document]
adamhathcock: really?  seems wrong 🤔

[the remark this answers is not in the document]
fabis94: @adamhathcock https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?pvs=4#c79ca97569504d1dbd9d5d3b9978360a

packages/objectloader2/src/operations/serverDownloader.ts
```diff
@@ -0,0 +1,158 @@
...
+  #requestUrlChildren: string
+  #headers: HeadersInit
+  #options: BaseDownloadOptions
+  #fetch: Fetcher
+
+  #downloadQueue?: BatchedPool<string>
+
+  constructor(options: BaseDownloadOptions) {
+    this.#options = options
+    this.#fetch = options.fetch ?? ((...args) => globalThis.fetch(...args))
```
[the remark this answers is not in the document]
adamhathcock: I think I had some type error when I tried that.

packages/objectloader2/src/operations/indexedDatabase.ts
```diff
@@ -113,6 +115,9 @@ export default class IndexedDatabase implements Cache {
       // const endTime = performance.now()
       // const duration = endTime - startTime
       // this.#logger('Read batch ' + x + ' ' + batch.length + ' ' + duration / 1000)
+
+      // interate down here to help with pausing
+      i += maxCacheReadSize
```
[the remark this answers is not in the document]
adamhathcock: I want to continue after a pause but not increment the loop.  Might be a more clear way but I increment last to simulate the for loop without changing much

PR #4175, opened 2025-03-12
feat(fe): Improve UX around out of context comments

andrewwallacespeckle:
Adding indicators when a comment is out of context. This could be because it was on another version, or it was placed on federated models. 

This got more complex than I was expecting, especially in `commentManagment.ts` and `serialization.ts`.

Let me know if there is a simpler way to do any of this.


packages/frontend-2/components/viewer/anchored-point/Thread.vue
```diff
@@ -545,6 +518,44 @@ onMounted(() => {
...
+  return {
+    text: 'Full context',
+    icon: ArrowUpRightIcon,
+    action: handleContextClick
+  }
+})
+
+const handleContextClick = () => {
+  if (threadResourceStatus.value.isDifferentVersion) {
+    onLoadThreadVersionContext()
```
fabis94, line 554: The naming for these 2 onLoadXX functions doesn't fully make sense - it implies that this is something that happens on load, but this is what triggers the load in the first place.

Also: Seems like this whole `handleContextClick` function could be returned from your composable as its only reliant on the internals already available in the composable. Then you wouldn't need to expose those 2 onLoadXX functions at all

packages/frontend-2/lib/viewer/composables/commentManagement.ts
```diff
@@ -237,3 +246,110 @@ export function useCheckViewerCommentingAccess() {
...
+    previousState.value = state
+    await applyState(state, mode)
+  }
+
+  const onLoadThreadVersionContext = () =>
+    loadContext(StateApplyMode.ThreadFullContextOpen)
+  const onLoadFederatedContext = () => loadContext(StateApplyMode.FederatedContext)
+
+  const goBack = async () => {
+    if (!previousRoute.value) {
```
fabis94, line 337: This will only re-apply the resourceIdString, while there's a bunch of other settings that will remain as is. Instead you should store the serialized viewer state before the full context is loaded and then apply it with a new StateApplyMode.Full mode that just copies everything from the state being passed in.

packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -256,10 +257,25 @@ export function useApplySerializedState() {
...
 
+    // Handle resource string updates
     if (
-      [StateApplyMode.Spotlight, StateApplyMode.TheadFullContextOpen].includes(mode)
+      [StateApplyMode.Spotlight, StateApplyMode.ThreadFullContextOpen].includes(mode)
     ) {
       await resourceIdString.update(state.resources.request.resourceIdString)
+    } else if (mode === StateApplyMode.FederatedContext) {
+      // For federated context, append only model IDs (without versions) to show latest
+      const currentResourceString = resourceIdString.value
```
fabis94, line 267: Use `parseUrlParameters()` from speckle shared to parse string into deserialized objects that are easier to work with. And then at the end serialize it back into a single string with `createGetParamFromResources`

packages/frontend-2/lib/viewer/composables/commentManagement.ts
```diff
@@ -237,3 +246,110 @@ export function useCheckViewerCommentingAccess() {
...
+    const hasLoadedVersions =
+      commitLinks.length > 0 &&
+      commitLinks.every((commitId) =>
+        loadedResources.some((lr) => lr.versionId && lr.versionId === commitId)
+      )
+
+    // Resource is loaded, check versions and federation
+    const currentModels = state.resources.response.modelsAndVersionIds.value
+    const threadModels = threadData.viewerResources.filter(
+      (r): r is typeof r & { modelId: string; versionId: string } =>
```
fabis94, line 296: why the type cast?

fabis94 (changes requested):

packages/frontend-2/lib/viewer/composables/commentManagement.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/e385d93cf8a5930c24d5ceb56f7292eb7e4c95ff

packages/frontend-2/lib/viewer/composables/serialization.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/cef55cd66a66cc642fed927ac8e6551170402759

packages/frontend-2/lib/viewer/composables/commentManagement.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/8c4cbbd77b8add4fb02c890554701a619ff593ab

packages/frontend-2/lib/viewer/composables/commentManagement.ts
```diff
@@ -237,3 +247,125 @@ export function useCheckViewerCommentingAccess() {
...
+
+  const goBack = async () => {
+    if (!previousState.value) {
+      return
+    }
+
+    await applyState(previousState.value, StateApplyMode.ThreadFullContextOpen)
+
+    if (previousRoute.value) {
+      await router.push(previousRoute.value)
```
fabis94, line 357: do we still need to store and reset the route? when you apply the serialized state, the resourceIdString, that already triggers a route change behind the scenes

fabis94 (changes requested):
a couple of remaining ones

packages/frontend-2/lib/viewer/composables/commentManagement.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/49ee7e005ce807310a627fdf2ffb074062931caf

packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -256,10 +257,37 @@ export function useApplySerializedState() {
...
+      )
+        .map((resource) => {
+          if (resource instanceof SpeckleViewer.ViewerRoute.ViewerModelResource) {
+            // Only keep model ID, drop version
+            return new SpeckleViewer.ViewerRoute.ViewerModelResource(resource.modelId)
+          }
+          return resource
+        })
+        .filter(
+          (resource) =>
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/3f457d4f6cf3e79e42a2227932f427c5f76374a6

packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -256,10 +257,37 @@ export function useApplySerializedState() {
...
 
+    // Handle resource string updates
     if (
-      [StateApplyMode.Spotlight, StateApplyMode.TheadFullContextOpen].includes(mode)
+      [StateApplyMode.Spotlight, StateApplyMode.ThreadFullContextOpen].includes(mode)
     ) {
       await resourceIdString.update(state.resources.request.resourceIdString)
+    } else if (mode === StateApplyMode.FederatedContext) {
+      // For federated context, append only model IDs (without versions) to show latest
+      const currentResources = SpeckleViewer.ViewerRoute.parseUrlParameters(
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/4175/commits/de620c90bed833de42198a05ce8c3fd171b4ba7f

fabis94 (dismissed):

fabis94 (dismissed):

fabis94 (approved):

PR #4176, opened 2025-03-12
Bump IFC nuget

JR-Morgan:
Bumped the IFC importer nuget

We're having some fun re-designing the speckle-sharp-connectors CI right now, so the semver is a bit funky 😁 but I (and sourcelink) assure you its a newer version


JR-Morgan:
> Is this from a branch? `*-adam*`

It's from this commit on `installer-test/adam/cnx-1367-build-subset-when-tagged` branch:
 https://github.com/specklesystems/speckle-sharp-connectors/commit/bdabd10ba433665005823f3bc29d260457de597c

Branch naming and nuget naming not final, just a happy accident that Adam was testing some deploys and this commit includes the changes I want.

JR-Morgan:
This was the change that's included in that nuget https://github.com/specklesystems/speckle-sharp-connectors/pull/655

PR #4185, opened 2025-03-13
Screen Space Outlines Update

AlexandruPopovici:
## Description & motivation
Because we've had several requests like [this one](https://speckle.community/t/revit-parts-visualisation-in-speckle/16037/5) and [this one](https://speckle.community/t/edge-visibility-in-speckle-viewer/12411/8) we went and updated our screen space outlines so that edges "between different objects" are computed and visible
In depth description available [here](https://www.notion.so/speckle/Screen-Space-Outlines-formerly-Edges-b76c89d6a32846f4b76a66a3cabe73ee?pvs=4#1b5b78fc7aa680f1aab3f1c675b8b304)

## Changes:
- Implemented a way to generate per object unique id + color very fast by using already existing data
- Implemented a MRT `DepthNormalIdPass` which works similarly to `DepthNormalPass` but it also generates object color id in an additional render target
- Id gradients are binary. If they are greater than 0, they will get updated to 1. This fixes the issue where similar nearby objects that happen to have similar colors would exhibit a fainter gradient
- When running WebGL 1.0 we dynamically add a per instance attribute buffer with the object id in order to compute hash. If we're running WebGL 2.0 we do not do that and instead rely on the builtin `gl_InstanceID` so we conserve bandwidth
- Fixed small type issue with the edge generation shader when running WebGL 1.0
- The updated screen space outlines pass and shaders now work with both WebGL 1.0 and WebGL 2.0
- `Batcher` now takes all the webgl caps instead of just mac vert uniforms and float textures
- Cleaned up the edge generator shader nicely
- Extended three.js's InstancedMesh in order to add a batch index variable to it.
- Removed some unneeded non-null assertion left over from a long time ago

## Screenshots:
**Before:**
old outlines

**Now:**
outlines

**Process:**
final

## Checklist:


didimitrie (dismissed):
My only question is why changes in the `daemon.js` of the fileimport service.

AlexandruPopovici:
Prettier changed it. I'll remove those changes

benjaminvo (approved):
Tested things locally. Code wise it's a trust-approve :).

PR #4198, opened 2025-03-14
refactor(fe): simplify planPlan to always use monthly price

andrewwallacespeckle:


packages/frontend-2/components/settings/workspaces/billing/PricingTable/Plan.vue
```diff
@@ -136,19 +136,7 @@ const isUpgradeDialogOpen = ref(false)
...
-
-const finalPlanPrice = computed(() => {
-  const basePrice = prices.value?.[props.plan].monthly?.['workspace:member']
-  if (!basePrice) return undefined
-  return {
-    ...basePrice,
-    amount: props.yearlyIntervalSelected ? basePrice.amount * 0.8 : basePrice.amount
-  }
-})
+const planPrice = computed(() => prices.value?.[props.plan]?.monthly)
```
Mikehrn, line 139: I cannot test this, I need to re-setup Stripe locally, but we need the check for `yearlyIntervalSelected`. The returned price needs to be 0.8 (20%) of the monthly price, or yearlyPrice divided by 12

Mikehrn (changes requested):

Mikehrn (approved):

PR #4212, opened 2025-03-18
Fix: Decrease sidebar with in new layout

Mikehrn:


andrewwallacespeckle (approved):
Just to confirm - sidebar/new is only used when FF is true, so you don't need to check for this here?

Mikehrn:
> Just to confirm - sidebar/new is only used when FF is true, so you don't need to check for this here?

Yup the Wrapper takes care of rendering the correct one :) Only not for settings, it's within the same component (as there is very little difference)

PR #4225, opened 2025-03-19
Feat: New princing table for new plans

Mikehrn:
Partly a duplication from the existing workspace plan, but adjusted according to the needs for the new workspace plans


packages/frontend-2/components/pricingTable/PricingTable.vue
```diff
@@ -0,0 +1,60 @@
...
+graphql(`
+  fragment PricingTable_Workspace on Workspace {
+    id
+    role
+  }
+`)
+
+const props = defineProps<{
+  slug: string
+  role: MaybeNullOrUndefined<WorkspaceRoles>
```
Mikehrn, line 33: I will probably move all of these into a composable, as mentioned earlier in another PR, but I will do this as a follow-up so I can adjust it in all places

andrewwallacespeckle (approved):

andrewwallacespeckle (approved):

PR #4228, opened 2025-03-19
feat(automate): ability to delete automations

cdriesler:
## Description & motivation

- Important user(s) (and Nikos) request

## Changes:

- Adds `delete()` mutation for automations
- Adds an action menu and delete modal to automation page that invokes it
- Uses Apollo cache to update automations after delete

Screen Shot 2025-03-19 at 22 44 11

Screen Shot 2025-03-19 at 22 44 20

## Validation of changes:

Testable on testing3


gjedlicska (commented):
Have you checked, that all automation related objects are getting cleared? I remember having some missing foreign keys in the DB, that would leave some items not deleted with the cascade.

cdriesler:
> Have you checked, that all automation related objects are getting cleared? I remember having some missing foreign keys in the DB, that would leave some items not deleted with the cascade.

Yes, db records all cascade nicely: (I think this was handled when fixing "projects with automations can't be deleted")

image

We are not clearing automation tokens in Automate's db, but we're planning to drop those entirely so I left them alone for now.

But after auditing I realized we're not deleting files associated with runs. Will address

PR #4231, opened 2025-03-20
refactor(fe): workspace sidebar members section

andrewwallacespeckle:


packages/frontend-2/components/workspace/header/Header.vue
```diff
@@ -77,6 +77,7 @@
         v-if="!isWorkspaceGuest"
         :workspace-info="workspaceInfo"
         :is-workspace-admin="isWorkspaceAdmin"
+        :is-workspace-guest="isWorkspaceGuest"
```
Mikehrn, line 80: It's not directly relevant, but I've been thinking about creating a composable that holds this basic WS information, now you will always be in 1 active workspace. Something like `useActiveWorkspace()` so we can just do `activeWorkspace.role` or `activeWorkspace.isAdmin`. wdyt?

packages/ui-components/src/components/layout/sidebar/menu/group/Group.vue
```diff
@@ -61,13 +62,14 @@ import Plus from '~~/src/components/global/icon/Plus.vue'
...
 import CommonBadge from '~~/src/components/common/Badge.vue'
+import { ChevronRightIcon } from '@heroicons/vue/24/outline'
 
 defineProps<{
   tag?: string
   title?: string
   collapsible?: boolean
   collapsed?: boolean
-  icon?: 'add' | 'edit'
+  icon?: 'add' | 'edit' | 'view'
```
Mikehrn, line 72: Maybe we can do something like FormButton, `icon?: Nullable<PropAnyComponent>` so it's a bit more flexible, can be a follow-up if cumbersome to do now

Mikehrn (approved):

PR #4235, opened 2025-03-21
feat: enable domain discoverability on workspace creation

fabis94:
checkbox is only available if user only has 1 verified non-blocked (not gmail) domain, which then gets used. the BE API itself does support picking from any of the user's available verified non-blocked domains


packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -437,37 +437,90 @@ export = FF_WORKSPACES_MODULE_ENABLED
...
+            logo,
+            slug,
+            enableDomainDiscoverabilityForDomain
+          } = args.input
+
+          const createWorkspace = commandFactory({
+            db,
+            eventBus,
+            operationFactory: ({ db, trx, emit }) => {
+              const createWorkspace = createWorkspaceFactory({
```
fabis94, line 452: Create + update + addDomain in a trx, which ensures that if any of those fail the workspace never gets committed

packages/frontend-2/components/workspace/wizard/step/Invites.vue
```diff
@@ -29,10 +29,16 @@
...
+      <div class="flex flex-col gap-2 w-full">
+        <FormCheckbox
+          v-model="enableDomainDiscoverabilityModel"
+          name="enableDomainDiscoverability"
+          :label="`Allow users with the @${verifiedDomain} domain to request to join workspace`"
+        />
+      </div>
+
       <div class="flex flex-col gap-3 mt-4 w-full md:max-w-96">
-        <FormButton size="lg" submit full-width>
```
Mikehrn, line 33: @benjaminvo @michalspeckle This is more of a design question; but now the main CTA is not "skip" anymore we should probably have another way to make clear that adding invites is not required 

Screenshot 2025-03-24 at 13 04 51

Mikehrn (approved):

packages/frontend-2/components/workspace/wizard/step/Invites.vue
michalspeckle: Not ideal but we could do it like this: 
image

benjaminvo: Good point Mike 🤔. What if the blue button said "Continue without inviting"? It's long – but clear :)

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
gjedlicska: wait, why not pass in the transaction as the db argument?

fabis94: @gjedlicska  what do you mean? that is what i'm doing here - passing in the trx as the db argument. Its only 1-2 lookup queries that use the main db connection, cause they don't need to be in a transaction

packages/frontend-2/components/workspace/wizard/step/Invites.vue
fabis94: updated

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: i've made it only use the trx for all of them, if thats what u prefer

Mikehrn (dismissed):

alemagio (dismissed):

PR #4243, opened 2025-03-21
feat(fe): Add pagination and search to Move Projects dialog

andrewwallacespeckle:
image


packages/frontend-2/components/workspace/MoveProjectsDialog.vue
```diff
@@ -105,8 +128,27 @@ const props = defineProps<{
...
+  onInfiniteLoad
+} = usePaginatedQuery({
+  query: moveProjectsDialogQuery,
+  baseVariables: computed(() => ({
+    cursor: null as string | null,
+    filter: search.value?.length ? { search: search.value } : null
+  })),
+  resolveKey: (vars) => [vars.filter?.search || ''],
+  resolveCurrentResult: (res) => res?.activeUser?.projects,
+  resolveNextPageVariables: (baseVars, cursor) => ({
```
Mikehrn, line 146: Can we add a filter for `workspaceId: null` here? That way we can get rid of this manual filtering (and fetching extra projects that are already in a workspace):

```
const projectsWithWorkspace = computed(() =>
  userProjects.value
    .filter((project) => !!project.workspace?.id)
    .map((project) => project.id)
)
const moveableProjects = computed(() =>
  userProjects.value.filter(
    (project) =>
      !workspaceProjects.value.includes(project.id) &&
      !projectsWithWorkspace.value.includes(project.id)
  )
)
```

Mikehrn: Should probably add `onlyWithRoles: Roles.Stream.Owner` as well, so we only get projects you are the owner of (meaning you can actually move them)

Mikehrn (changes requested):
Nice

Mikehrn (approved):

PR #4244, opened 2025-03-21
Fix: Various pricing plan fixes

Mikehrn:
Some small fixes
- Redirect to checkout if no active subscription 
- Update plan naming
- Update button text


benjaminvo (approved):
If not too much work it would be great to align the internal and public plan names when the old plans are deprecated :)

PR #4253, opened 2025-03-24
chore(authz): hoist authorizeresolver

cdriesler:
## Description & motivation

- WIP auth work is complicated by `authorizeResolver` calls "buried" in services.
- Some of these are easy to hoist, some of these are in use in what are effectively "policies" already.

## Changes:

- Hoists easy-to-remove calls into resolvers
- Leaves "policy-like" calls for when we write those particular policies.


gjedlicska (approved):
what could go wrong

PR #4257, opened 2025-03-24
Feat: Update plans with new metadata

Mikehrn:
Limits needs to be updated when it's actually available


packages/shared/src/workspaces/helpers/features.ts
```diff
@@ -25,11 +25,25 @@ export const WorkspacePlanFeatures = <const>{
   CustomDataRegion: 'workspaceDataRegionSpecificity'
 }
 
+export const WorkspacePlanFeaturesNew = <const>{
+  // Core features
+  UnlimtedSeats: 'unlimitedSeats',
```
benjaminvo, line 30: Typo here. `Unlimted`

benjaminvo (commented):
Noticed a typo. Why not include the actual project, model and version history limits now? Because it should come from the backend?

packages/shared/src/workspaces/helpers/features.ts
Mikehrn: Sharp eyes 😎
Yeah we didn't fully settle on what that will look like, but are working it, we can do it separately once we've done that

benjaminvo: > Sharp eyes 😎

I actually read it!

benjaminvo (approved):

andrewwallacespeckle (approved):

PR #4260, opened 2025-03-25
chore(automate): improve error message

cdriesler:
## Description & motivation

- There's a technically-correct but hard to understand error when you try to trigger an automation with a model that has no version
automate-error

## Changes:

- Changes it

### DIFF
--- a/packages/server/modules/automate/services/trigger.ts
+++ b/packages/server/modules/automate/services/trigger.ts
@@ -522,7 +522,7 @@ export const manuallyTriggerAutomationFactory =
     )
     if (!latestCommit) {
       throw new TriggerAutomationError(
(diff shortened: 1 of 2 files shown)


packages/server/modules/automate/services/trigger.ts
```diff
@@ -522,7 +522,7 @@ export const manuallyTriggerAutomationFactory =
     )
     if (!latestCommit) {
       throw new TriggerAutomationError(
-        'No version to trigger on found for the available triggers'
+        'Selected model has no versions and cannot be used to trigger an automation.'
```
benjaminvo, line 525: May I suggest "so it cannot" instead of "and cannot"? Yes, I'm not picking.

gjedlicska (approved):

PR #4262, opened 2025-03-25
feat: true-myth result structures & other auth policy improvements

fabis94:
## Checklist:


packages/server/modules/core/authz/loaders/index.ts
```diff
@@ -1,27 +1,31 @@
...
 import { getStreamFactory } from '@/modules/core/repositories/streams'
-import { defineLoaders } from '@/modules/loaders'
 import { getFeatureFlags } from '@/modules/shared/helpers/envHelper'
 import { db } from '@/db/knex'
 import { getUserServerRoleFactory } from '@/modules/shared/repositories/acl'
+import { err, ok } from 'true-myth/result'
+import { Authz } from '@speckle/shared'
 
-export const defineModuleLoaders = () => {
+export default defineModuleLoaders(async () => {
```
fabis94, line 9: Simplified & improved DX of the way server modules define auth loaders. Previously you had to manually invoke defineModuleLoaders() somewhere and you also had to manually invoke defineLoaders() inside of it and you were not even forced to do it.

Now loaders are dynamically auto-loaded from /<module>/authz/loaders/index.ts and you just have to `export default defineModuleLoaders()` which will return the loaders to auto-load and provide proper TS typing

packages/server/type-augmentations/true-myth.d.ts
```diff
@@ -0,0 +1,10 @@
+// Only need to do this because our CommonJS app does not support true-myth's export maps
```
fabis94, line 1: another reason to move to an ESM build ASAP - libraries are relying on "exports" maps in package.json more and more, and we can't use them (at least their TS types)

packages/shared/src/authz/domain/loaders.ts
```diff
@@ -8,11 +10,47 @@ import type {
...
+      [key in AuthCheckContextLoaderKeys]: unknown
+    },
+    Mapping
+  >
+>
+
+/**
+ * All loaders must be listed here for app startup validation to work properly
+ */
+export const AuthCheckContextLoaderKeys = <const>{
```
fabis94, line 39: 1. didn't really make sense that a runtime collection of the expected loader keys lived in `server` not `shared`, so i moved it over
2. also made it an enum so that (optionally) loaders can be referenced through the enum without string literals

packages/shared/src/authz/domain/loaders.ts
```diff
@@ -8,11 +10,47 @@ import type {
...
+  getWorkspace: 'getWorkspace',
+  getWorkspaceRole: 'getWorkspaceRole',
+  getWorkspaceSsoProvider: 'getWorkspaceSsoProvider',
+  getWorkspaceSsoSession: 'getWorkspaceSsoSession'
+}
+
+export type AuthCheckContextLoaderKeys =
+  (typeof AuthCheckContextLoaderKeys)[keyof typeof AuthCheckContextLoaderKeys]
+
+export type AllAuthCheckContextLoaders = AuthContextLoaderMappingDefinition<{
```
fabis94, line 53: This wrapping type ensures the loaders follow an expected schema. For example, it forces loaders to be async, even if their return types aren't. Loaders should **always** be async cause they will load things different ways in different environments and there may be caching added ontop

packages/shared/src/authz/policies/canQueryProject.ts
```diff
@@ -18,44 +16,45 @@ import {
...
-  }: UserContext & ProjectContext): Promise<
-    AuthResult<
-      | typeof ProjectNotFoundError
-      | typeof ProjectNoAccessError
-      | typeof WorkspaceNoAccessError
-      | typeof WorkspaceSsoSessionInvalidError
-    >
-  > => {
-    const { FF_ADMIN_OVERRIDE_ENABLED, FF_WORKSPACES_MODULE_ENABLED } = loaders.getEnv()
+export const canQueryProjectPolicyFactory: AuthPolicyFactory<
```
fabis94, line 23: Instead of all of the arg, loader & return types being typed out next to the actual args, loaders & return, I instead created an AuthPolicyFactory type that the policy factory constant is typed at.

This ensures that the policy factory follows an expected schema, that it can't randomly decided to have more positional args next to loaders, or return a non-AuthResult etc.

All of the loader, arg & return type requirements are also defined here as generic args of AuthPolicyFactory

packages/shared/src/authz/policies/canQueryProject.ts
```diff
@@ -18,44 +16,45 @@ import {
...
+  | typeof AuthCheckContextLoaderKeys.getEnv
+  | typeof AuthCheckContextLoaderKeys.getProject
+  | typeof AuthCheckContextLoaderKeys.getProjectRole
+  | typeof AuthCheckContextLoaderKeys.getServerRole
```
fabis94, line 27: Using the enum makes it easier to find references of loaders across policies and check functions. It was annoying that Right Clicking on a loader key and clicking "Find References" wouldn't find anything

packages/shared/src/authz/policies/canQueryProject.ts
```diff
@@ -18,44 +16,45 @@ import {
...
+  | typeof ProjectNotFoundError
+  | typeof ProjectNoAccessError
+  | typeof WorkspaceNoAccessError
+  | typeof WorkspaceSsoSessionInvalidError
+> =
+  (loaders) =>
+  async ({ userId, projectId }) => {
+    const env = await loaders.getEnv()
+    if (!isOk(env)) {
+      throw new LogicError('Failed to load environment variables')
```
fabis94, line 41: I was debating whether this should be a throw or a true-myth err() result. I thought that this is essentially a LogicError - something that should never happen during runtime and so it would be annoying to have to check for a `EnvVarsNotFoundError` structure everywhere.

It's still annoying that getEnv() has to have isOk() checked on it, even tho it doesn't have any kind of error cases, but true-myth does not support returning an `ok()` struct that you don't have to check `.isOk` on. And similarly I did not want to leave one of the loaders without a true-myth return

packages/shared/src/authz/policies/canQueryProject.ts
```diff
@@ -18,44 +16,45 @@ import {
...
+    }
 
     // All users may read public projects
     const isPublicResult = await requireExactProjectVisibilityFactory({ loaders })({
       projectId,
       projectVisibility: 'public'
     })
     if (isPublicResult) {
-      return authorized()
+      return ok(true)
```
fabis94, line 57: Because we need true-myth structures for loaders (e.g. getProject can not only not find the project, but also just not be able to retrieve it through GQL because the user is not authorized), it also didn't make sense anymore to have our custom authorized()/unauthorized() helpers and I moved to true-myth structs here as well

packages/shared/eslint.config.mjs
```diff
@@ -36,6 +36,12 @@ const configs = [
     rules: {
       '@typescript-eslint/no-explicit-any': 'off'
     }
+  },
+  {
+    files: ['**/*.spec.ts'],
+    rules: {
+      '@typescript-eslint/require-await': 'off' // so we can easily make sync mocked loaders -> async
```
gjedlicska, line 43: thank you.

packages/shared/src/authz/domain/authErrors.ts
```diff
@@ -25,12 +25,42 @@ export const ProjectNoAccessError = defineAuthError({
   message: 'You do not have access to the project'
 })
 
+export const ProjectRoleNotFoundError = defineAuthError({
```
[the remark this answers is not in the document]
gjedlicska: Fine for now, lets see how it goes

gjedlicska (approved):
good start, lets learn while adding new stuff

packages/shared/src/authz/checks/projects.ts
```diff
@@ -1,34 +1,34 @@
 import { StreamRoles, throwUncoveredError } from '../../core/index.js'
 import { ProjectNotFoundError } from '../domain/errors.js'
-import { AuthCheckContext } from '../domain/loaders.js'
+import { AuthCheckContext, AuthCheckContextLoaderKeys } from '../domain/loaders.js'
 import { isMinimumProjectRole } from '../domain/projects/logic.js'
 import { ProjectVisibility } from '../domain/projects/types.js'
 
 export const requireExactProjectVisibilityFactory =
-  ({ loaders }: AuthCheckContext<'getProject'>) =>
+  ({ loaders }: AuthCheckContext<typeof AuthCheckContextLoaderKeys.getProject>) =>
```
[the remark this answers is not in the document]
fabis94: this was the main reason for these specific changes https://github.com/specklesystems/speckle-server/pull/4262#discussion_r2012367864

PR #4264, opened 2025-03-25
Feat: Adapt workspace invites for new plans

Mikehrn:
There are quite some conditionals in there in order to support both old and new flows, can be cleaned up significantly post-release

Screenshot 2025-03-25 at 23 25 07

Screenshot 2025-03-25 at 23 25 36


andrewwallacespeckle:
Strange edge case here, but if I add an empty row between emails it fails, even though the emails are valid. 

image
image

andrewwallacespeckle:
> Strange edge case here, but if I add an empty row between emails it fails, even though the emails are valid.
> 
> image image

Same happens with FF New plans off too

packages/frontend-2/components/invite/dialog/Workspace.vue
```diff
@@ -1,13 +1,22 @@
 <template>
-  <div>
+  <LayoutDialog
```
andrewwallacespeckle, line 2: When FF for new plans is off, open the invite dialog (it opens the old as expected). 

Hit cancel and you'll go to the new view:
image

andrewwallacespeckle (dismissed):

andrewwallacespeckle (approved):

PR #4275, opened 2025-03-27
feat: authz frontend foundation + reworked errors

fabis94:
## Checklist:


packages/shared/src/authz/domain/authErrors.ts
```diff
@@ -1,17 +1,42 @@
...
-  code: ErrorCode
-  message: string
+export type AuthError<ErrorCode extends string = string, Payload = undefined> = {
+  readonly code: ErrorCode
+  readonly message: string
+  readonly payload: Payload
 }
 
-export const defineAuthError = <ErrorCode extends string>(params: {
+export const defineAuthError = <
```
fabis94, line 7: Had to rework errors cause I arrived at a use case that needed them to be customizable w/ custom payloads (like VError.info().

This is the best DX I could achieve with these requirements in mind:
* the error types are actual dynamic classes so you can instantiate them like ones (e.g. new Error()
* you're forced to define the code and its accessible both on the instance and also the class itself
* you can attach a dynamic payload, and you don't have to even specify it if there is none (type = undefined)

only ugly thing about this is having to do `InstanceType<typeof FooError>` wherever you want to reference the error type

packages/shared/src/authz/domain/authErrors.ts
```diff
@@ -45,16 +70,21 @@ export const WorkspaceSsoProviderNotFoundError = defineAuthError({
...
-
 export const WorkspaceSsoSessionNotFoundError = defineAuthError({
   code: 'WorkspaceSsoSessionNotFound',
   message: 'Your workspace SSO session was not found'
 })
 
+export const WorkspaceSsoSessionNoAccessError = defineAuthError<
+  'WorkspaceSsoSessionNoAccess',
+  {
+    workspaceSlug: string
```
fabis94, line 81: We need to pass along the workspaceSlug cause this error should cause a redirect to the workspace in FE

packages/shared/src/authz/domain/workspaces/operations.ts
```diff
@@ -3,28 +3,40 @@ import { WorkspaceRoles } from '../../../core/constants.js'
...
+    | InstanceType<typeof WorkspaceSsoSessionNoAccessError>
+    | InstanceType<typeof WorkspaceNoAccessError>
+  >
+>
 
 export type GetWorkspaceRole = (args: {
   userId: string
   workspaceId: string
-}) => Promise<Result<WorkspaceRoles, typeof WorkspaceRoleNotFoundError>>
+}) => Promise<Result<WorkspaceRoles, InstanceType<typeof WorkspaceRoleNotFoundError>>>
```
fabis94, line 28: I made the decision not to surface WorkspaceNotFound/WorkspaceNoAccess/WorkspaceSsoSessionNoAccess errors out of this one, even tho they can pop up from the loader on the frontend. This is because I wanted to avoid the complication of surfacing all of these error types any time anything related to workspaces is loaded on the FE. I think it's better to assume that the developer will make sure to invoke GetWorkspace first, before moving on and checking more granular aspects of the workspace.

Same idea also in Project loaders

packages/shared/package.json
```diff
@@ -78,15 +79,17 @@
...
   "tshy": {
     "exports": {
+      ".": "./src/index.ts",
       "./package.json": "./package.json",
       "./pinoPrettyTransport.cjs": "./pinoPrettyTransport.cjs",
       "./environment": "./src/environment/index.ts",
       "./observability": "./src/observability/index.ts",
-      "./dist/*": "./dist/*",
-      ".": "./src/index.ts"
+      "./authz": "./src/authz/index.ts",
```
fabis94, line 87: Anything running ESM can import authz stuff directly from `@speckle/shared/authz` now

packages/frontend-2/middleware/requireValidProject.ts
```diff
@@ -1,80 +1,47 @@
...
-  }
-
-  if (errors?.length) {
-    return abortNavigation(
-      createError({
-        statusCode: 500,
-        message: getFirstErrorMessage(errors) || 'An unexpected error occurred'
-      })
-    )
+  const authPolicies = useAuthPolicies()
```
fabis94, line 16: Project access check middleware rewritten with the new policy

packages/frontend-2/middleware/requireValidProject.ts
```diff
@@ -1,80 +1,47 @@
...
+  const canAccess = await authPolicies.noCache.project.canQuery({
+    projectId,
+    userId: ActiveUserId
+  })
+  if (canAccess.isErr) {
+    switch (canAccess.error.code) {
+      case WorkspaceSsoSessionNoAccessError.code: {
+        // Redirect to the SSO error page
+        const workspaceSlug = canAccess.error.payload.workspaceSlug
+        return navigateTo(`/workspaces/${workspaceSlug}/sso/session-error`)
```
fabis94, line 26: here's the redirect that relies on error metadata

fabis94: cc @Mikehrn / @andrewwallacespeckle i'd appreciate help in testing this out, i dont have a functioning SSO setup

gjedlicska (dismissed):
errors are a  bit more tricky to work with right now, but they get the job done.
We'll try to come up with a better pattern if possible.

packages/frontend-2/middleware/requireValidProject.ts
Mikehrn: I dont have this locally either, I think @andrewwallacespeckle can help you with it, otherwise seems to work as expected

Mikehrn (approved):

packages/frontend-2/middleware/requireValidProject.ts
andrewwallacespeckle: Just tested and this works as it did before. It successfully redirects to the sso page when a user tries to navigate to a workspace that requires sso, and they have not logged in via sso.

PR #4281, opened 2025-03-27
fix(shared): workspace sso fragment must be able to load limited workspace data

gjedlicska:
## Checklist:

### DIFF
--- a/packages/shared/src/authz/domain/workspaces/operations.ts
+++ b/packages/shared/src/authz/domain/workspaces/operations.ts
@@ -7,7 +7,6 @@ import {
   WorkspaceNotFoundError,
   WorkspaceRoleNotFoundError,
   WorkspaceSsoProviderNotFoundError,
-  WorkspaceSsoSessionNoAccessError,
   WorkspaceSsoSessionNotFoundError
 } from '../authErrors.js'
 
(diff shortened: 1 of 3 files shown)


packages/shared/src/authz/fragments/workspaceSso.spec.ts
```diff
@@ -12,6 +14,39 @@ import {
...
+      })
+      await expect(result).resolves.toStrictEqual(
+        just(err(new WorkspaceNoAccessError()))
+      )
+    }
+  )
+  it('throws Uncovered error for unknown loader errors', async () => {
+    await expect(
+      maybeMemberRoleWithValidSsoSessionIfNeeded({
+        // @ts-expect-error testing the unexpected error case here
```
fabis94, line 38: if you put a @ts-expect-error here, its not gonna show errors even if there are completely missing dependencies in the factory, which is not good

maybe if you put it lower, inside of the actual body of the getWorkspace implementation, then it won't apply to the entire factory. if not, maybe its better to do `error as AcceptableErrorType`

packages/shared/src/authz/domain/workspaces/operations.ts
```diff
@@ -17,7 +16,6 @@ export type GetWorkspace = (args: {
   Result<
     Workspace,
     | InstanceType<typeof WorkspaceNotFoundError>
-    | InstanceType<typeof WorkspaceSsoSessionNoAccessError>
```
fabis94, line 20: that error has to remain there, because it can be thrown when the FE invokes the getWorkspace loader

fabis94 (changes requested):
left comments

PR #4284, opened 2025-03-28
feat: Workspace/ProjectCollaborator/WorkspaceCollaborator seatType

fabis94:


packages/server/assets/gatekeeperCore/typedefs/workspaceSeats.graphql
```diff
@@ -4,7 +4,21 @@ enum WorkspaceSeatType {
 }
 
 extend type WorkspaceCollaborator {
-  seatType: WorkspaceSeatType!
+  seatType: WorkspaceSeatType
```
fabis94, line 7: made it nullable in case module is disabled

packages/server/observability/logging.ts
```diff
@@ -6,7 +6,7 @@ const { getLogger, extendLoggerComponent } = Observability
 
 export const logger = getLogger(
   process.env.LOG_LEVEL || 'info',
-  process.env.LOG_PRETTY === 'true'
+  process.env.LOG_PRETTY === 'true' && !process.env.FORCE_NO_PRETTY
```
fabis94, line 9: on node 22 pino pretty causes debugger to freeze on startup sometime. i don't want to change the env vars in the `yarn dev` command every time, so it's cleaner to just introduce an extra flag that disables it forcefully irregardless of what LOG_PRETTY is so I can just do `FORCE_NO_PRETTY=1 yarn dev`

packages/server/assets/gatekeeperCore/typedefs/workspaceSeats.graphql
gjedlicska: we should pro move things around. If you are on a new plan workspace, you will 100% have a seat type, no matter if the gatekeeper module is enabled or not.

But for now, if you are on an old plan, we do not assign seats, so the typing change is technically correct.

packages/server/modules/gatekeeper/repositories/workspaceSeat.ts
```diff
@@ -136,3 +141,67 @@ export const getWorkspaceRoleAndSeatFactory =
...
+      userId: string
+      projectId: string
+    }>
+  }) => {
+    const { requests } = params
+    const idPairs = requests.map(({ userId, projectId }) => [userId, projectId])
+
+    const q = tables
+      .streamAcl(deps.db)
+      .whereIn([StreamAcl.col.userId, StreamAcl.col.resourceId], idPairs)
```
fabis94, line 181: `WHERE (userId, resourceId) in ((x, y), (x, y), (x,y))` SQL syntax so we can get results for multiple workspaceIds with 1 SQL query

cdriesler (approved):

PR #4292, opened 2025-03-31
Alessandro/web 2906 list invitable workspace members in a project

alemagio:
## Checklist:


packages/server/modules/workspaces/repositories/users.ts
```diff
@@ -12,3 +20,87 @@ export const setUserActiveWorkspaceFactory =
...
+    limit
+  }: {
+    filter: {
+      workspaceId: string
+      projectId: string
+      search?: string
+    }
+    cursor?: string
+    limit: number
+  }): Promise<WorkspaceTeamMember[]> => {
```
fabis94, line 38: count & get items have the same base query that should be extracted out to its own function

packages/server/modules/workspaces/repositories/users.ts
```diff
@@ -12,3 +20,87 @@ export const setUserActiveWorkspaceFactory =
...
+      .where(WorkspaceAcl.col.workspaceId, workspaceId)
+      .whereNotIn(
+        Users.col.id,
+        await tables
+          .streamAcl(db)
+          .select(StreamAcl.col.userId)
+          .where(StreamAcl.col.resourceId, projectId)
+      )
+    if (search) {
+      query.andWhere((w) =>
```
fabis94, line 54: shouldnt UserEmails only be joined if `search` is set?

packages/server/modules/workspaces/tests/integration/repositories/users.spec.ts
```diff
@@ -0,0 +1,292 @@
...
+} from '@/modules/workspaces/tests/helpers/creation'
+import { createTestUser } from '@/test/authHelper'
+import { createTestStream } from '@/test/speckle-helpers/streamHelper'
+import { Roles } from '@speckle/shared'
+import { expect } from 'chai'
+import { pick } from 'lodash'
+
+const { FF_WORKSPACES_MODULE_ENABLED } = getFeatureFlags()
+
+;(FF_WORKSPACES_MODULE_ENABLED ? describe : describe.skip)(
```
fabis94, line 20: unnecessary, the entire workspace module tests get skipped if this FF is not set

packages/server/modules/workspaces/repositories/users.ts
alemagio: unfortunately I couldn't make it work since I need an async function to build the query because of the whereNotIn clause.
I belive the way knex works doesn't fit with async functions since the query is immediately resolved and I can't use the query builder.

fabis94: that async clause doesn't need to be async either

cdriesler (dismissed):
LGTM (and looks like you adequately addressed Fabians comments)

packages/server/modules/workspaces/graph/resolvers/projects.ts
```diff
@@ -0,0 +1,52 @@
...
+          // TODO: add authz policy
+          if (!parent.workspaceId) {
+            return {
+              totalCount: 0,
+              items: [],
+              cursor: null
+            }
+          }
+
+          return await getPaginatedItemsFactory<
```
fabis94, line 26: if limit===0 only get count, dont invoke item retrieval query at all (same convention we have in other paginated resolvers)

packages/server/modules/workspaces/repositories/users.ts
```diff
@@ -12,3 +20,89 @@ export const setUserActiveWorkspaceFactory =
...
+    limit: number
+  }): Promise<WorkspaceTeamMember[]> => {
+    const { workspaceId, projectId, search } = filter
+    const query = buildInvitableCollaboratorsByProjectIdQueryFactory({ db })({
+      workspaceId,
+      projectId,
+      search
+    })
+    if (cursor) {
+      query.andWhere(Users.col.createdAt, '<', cursor)
```
fabis94, line 81: ideally we should use decodeCompositeCursor/encodeCompositeCursor and order by both a date and also user id, to ensure we don't run into issues when multiple users have the same date

fabis94 (changes requested):

fabis94 (approved):

PR #4294, opened 2025-03-31
gergo/web 2888 workspace project cancreate

gjedlicska:
- WIP can create project
- WIP can create project more work

## Checklist:


packages/shared/src/authz/fragments/workspaceSso.ts
```diff
@@ -21,64 +23,46 @@ export const maybeMemberRoleWithValidSsoSessionIfNeeded: AuthPolicyFragment<
...
-        case 'WorkspaceNoAccess':
-        case 'WorkspaceNotFound':
-          return just(err(new WorkspaceNoAccessError()))
-        case 'WorkspaceSsoSessionNoAccess':
-          return just(err(workspace.error))
-        default:
-          throwUncoveredError(workspace.error)
-      }
-    }
+    // hides the fact, that the workspace does not exist
```
fabis94, line 26: considering that we're no longer running this in the FE and that loader results are cached I think we should make fragments as self-contained as possible to simplify using them - you should not need to prepare any checks before hand /have any assumptions about what the state of the workspace/policy is, before you invoke the fragment

so, judging by the figma and current usage of this, the fragment should 1) do the hasAnyWorkspaceRole check inside of it, instead of having to remember to invoke it before the check 2) maybe it should even accept a projectId as an arg, not a workspaceId, so you don't even have to do the `if (WORKSPACES_ENABLED && project.workspaceId)` check beforehand, and you can just invoke this fragment without wrapping it in any preparatory conditonals

packages/shared/src/authz/policies/index.ts
```diff
@@ -1,9 +1,9 @@
 import { AllAuthCheckContextLoaders } from '../domain/loaders.js'
-import { canQueryProjectPolicy } from './canQueryProject.js'
+import { canReadProjectPolicy } from './canReadProject.js'
 
 export const authPoliciesFactory = (loaders: AllAuthCheckContextLoaders) => ({
   project: {
-    canQuery: canQueryProjectPolicy(loaders)
+    canRead: canReadProjectPolicy(loaders)
```
fabis94, line 6: should add the new policy here

packages/shared/src/authz/policies/canCreateWorkspaceProject.ts
```diff
@@ -0,0 +1,103 @@
...
+import { hasMinimumServerRole } from '../checks/serverRole.js'
+import { Roles } from '../../core/constants.js'
+import { maybeMemberRoleWithValidSsoSessionIfNeeded } from '../fragments/workspaceSso.js'
+import { hasAnyWorkspaceRole } from '../checks/workspaceRole.js'
+import { throwUncoveredError } from '../../core/index.js'
+import { hasEditorSeat } from '../checks/workspaceSeat.js'
+import { MaybeUserContext, WorkspaceContext } from '../domain/context.js'
+
+export const canCreateWorkspaceProjectPolicy: AuthPolicy<
+  | 'getEnv'
```
fabis94, line 22: I'm wondering if this should just be `canCreateProjectPolicy` instead, and then inside if it we differentiate between whether it's a workspace or non-workspace policy? Would simplify the permission API a bit

cdriesler: I have the non-workspace one WIP, will see how much they overlap after. (The non-workspace one is stupid simple compared to this.)

packages/server/modules/workspaces/authz/loaders/index.ts
```diff
@@ -1,46 +1,62 @@
...
+      )
+    },
+    getWorkspaceProjectCount: async ({ workspaceId }, { dataLoaders }) => {
+      return await dataLoaders.workspaces!.getProjectCount.load(workspaceId)
+    },
+    getWorkspacePlan: async ({ workspaceId }) => {
+      return await getWorkspacePlan({ workspaceId })
+    },
+    getWorkspaceLimits: async () => {
+      // TODO: Do real implementation
```
fabis94, line 55: There's no existing repo fn and I don't know how this is supposed to be resolved, so y'all need to do the real implementation here

packages/server/modules/workspaces/graph/resolvers/permissions.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export default {
+  Workspace: {
+    permissions: (parent) => ({
+      workspaceId: parent.id
+    })
+  },
+  WorkspacePermissionChecks: {
+    canCreateProject: async (parent, _args, ctx) => {
+      const canCreateProject = await ctx.authPolicies.workspace.canCreateProject({
```
fabis94, line 12: I already exposed the new policy through the GQL API, so when this is merged it should be working correctly

PR #4306, opened 2025-04-01
feat(workspaces): add query for fetching usage

cdriesler:
## Description & motivation

- Requires #4294 
- FE needs to know how the workspace is currently using its quotas. Some of this is technically possible with various existing endpoints, but not in a performant or simple way.

## Changes:

- Adds `usage` field to `WorkspacePlan` that emits current usage values:
  - `projectCount`
  - `modelCount`


alemagio (changes requested):
regional data and query for single workspace

alemagio (approved):

PR #4308, opened 2025-04-01
fix(gatekeeper): missing priceId-s should stop the server from booting

gjedlicska:
The missing env vars were only validated at request time, not at server startup time.

Also included some log level lowerings, they are not relevant right now.


gjedlicska:
this is not needed any more

PR #4319, opened 2025-04-04
feat(server): workspace roles taken into account in project queries

fabis94:
subsequent PRs will focus on fixing up the frontend


packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -113,7 +113,7 @@ const {
   filter: {
     search: (search.value || '').trim() || null,
     onlyWithRoles: selectedRoles.value?.length ? selectedRoles.value : null,
-    workspaceId: isWorkspaceNewPlansEnabled ? (null as Nullable<string>) : undefined
+    personalOnly: isWorkspaceNewPlansEnabled.value
```
fabis94, line 116: there's no such thing as undefined in GQL, its either a set value or null. so the current implementation of "only get personal projects" is busted, which is why I introduced a properly working flag

packages/server/modules/core/graph/resolvers/projects.ts
```diff
@@ -284,6 +284,13 @@ export = {
         throw new RateLimitError(rateLimitResult)
       }
 
+      const canCreate = await context.authPolicies.project.canCreateLegacy({
+        userId: context.userId
+      })
+      if (canCreate.isErr) {
+        throw canCreate.error
```
fabis94, line 291: Just an example of how the switch mapping could be avoided and auth errors could be thrown as is

packages/server/modules/core/graph/setup.ts
```diff
@@ -35,29 +36,34 @@ export function buildErrorFormatter(params: {
       }
     }
 
-    // If error isn't a VError child, don't do anything extra
-    if (!(realError instanceof VError)) {
-      return formattedError
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    let extensions: { [key: string]: any } = {
```
fabis94, line 40: Support for properly formatting AuthErrors in GQL responses

packages/server/modules/shared/repositories/roles.ts
```diff
@@ -28,6 +32,13 @@ export const getRolesFactory =
     return roles
   }
 
+export const getCachedRolesFactory = wrapFactoryWithCache({
```
fabis94, line 35: these will never change while the app is running

packages/server/modules/shared/services/auth.ts
```diff
@@ -106,7 +114,27 @@ export const authorizeResolverFactory =
       : null
 
     if (!userAclRole) {
-      throw new ForbiddenError('You are not authorized to access this resource.')
+      // TODO: Could be more optimized (caching?) but we're moving away from this towards
```
fabis94, line 117: Fixed up authorizeResolver to take into account implicit workspace roles. Ofc we need new policies for mutations anyway, because the frontend needs them for checks

packages/server/modules/workspacesCore/domain/events.ts
```diff
@@ -7,7 +7,7 @@ export const workspaceEventNamespace = 'workspace' as const
 const eventPrefix = `${workspaceEventNamespace}.` as const
 
 export const WorkspaceEvents = {
-  Authorized: `${eventPrefix}authorized`,
+  Authorizing: `${eventPrefix}authorizing`,
```
fabis94, line 10: the previous name implied that the workspace was already authorized, so any errors thrown from this event handler wouldn't actually cancel the authorization

packages/server/modules/shared/repositories/roles.ts
fabis94: (just have to make sure not to invoke this before the init role migrations are done...currently its fine, this is only used in authorizeResolver)

packages/server/modules/shared/index.ts
```diff
@@ -1,11 +1,12 @@
 import { db } from '@/db/knex'
 import { getStreamFactory } from '@/modules/core/repositories/streams'
+import { getWorkspaceRoleAndSeatFactory } from '@/modules/gatekeeper/repositories/workspaceSeat'
```
gjedlicska, line 3: if we need this in the shared module, pls move this repo function into the foss licensed `workspaceCore` module

packages/shared/src/authz/policies/project/canCreate.spec.ts
```diff
@@ -0,0 +1,66 @@
+import { describe, expect, it } from 'vitest'
+import { canCreateProjectPolicy } from './canCreate.js'
+import { parseFeatureFlags } from '../../../environment/index.js'
+import {
+  ProjectNoAccessError,
+  ServerNoAccessError,
+  ServerNoSessionError
+} from '../../domain/authErrors.js'
+
+const buildSUT = (overrides?: Partial<Parameters<typeof canCreateProjectPolicy>[0]>) =>
```
gjedlicska, line 10: whats a SUT?

gjedlicska (changes requested):
most importantly, lets fix the licensing by moving the repo function. Otherwise im ok with this

packages/shared/src/authz/policies/project/canCreate.spec.ts
fabis94: System Under Test, common way to refer to the thing being tested in a unit test

gjedlicska: til

packages/shared/src/authz/policies/project/canCreate.ts
```diff
@@ -0,0 +1,43 @@
...
+    | typeof ProjectNoAccessError
+    | typeof WorkspaceSsoSessionNoAccessError
+  >
+> =
+  (loaders) =>
+  async ({ userId }) => {
+    const env = await loaders.getEnv()
+    if (!userId?.length) return err(new ServerNoSessionError())
+    if (env.FF_WORKSPACES_MODULE_ENABLED) {
+      // TODO: We're not ready to enforce this yet, there's a bunch of tests that would break
```
fabis94, line 29: cc @gjedlicska this particular check should be its own ticket

gjedlicska (dismissed):

gjedlicska (approved):

PR #4343, opened 2025-04-08
feat(shared): new policies for minor fragment refactor

fabis94:
## Checklist:


packages/shared/src/authz/domain/policies.ts
```diff
@@ -1,36 +1,65 @@
...
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  ExpectedAuthErrors extends AuthError<any, any>
+  ExpectedAuthErrors extends AuthError<any, any>,
+  Return
 > = (
   loaders: AuthCheckContextLoaders<LoaderKeys>
-) => (args: Args) => Promise<Maybe<Result<Unit, ExpectedAuthErrors>>>
+) => (args: Args) => Promise<Result<Return, ExpectedAuthErrors>>
+
+export type AuthPolicyEnsureFragment<
```
fabis94, line 29: The fact that fragments could have 3 outcomes: nothing, ok() and error, was too complicated to reason about. The fact that a fragment could decide that the policy should `ok()` was also very limiting - what if the policy has extra checks it wants to do after that fragment? And also, fragments relied on pre-existing checks too much, making them even harder to reason about.

The overly complicated nature was also shown by the name of the only fragment - `maybeMemberRoleWithValidSsoSessionIfNeeded()`. It was so difficult to name precisely because it has too many concerns and is only usable within specific circumstances (because of being able to ok() the parent policy).

I made a minor pivot here:
1. Fragments still remain (composable policy parts that return a Result structure), but there's 2 different kinds now
2. An Ensure fragment - it ensures something is the case (e.g. user is logged in), or returns an error structure
3. A check fragment - it checks if something is the case and returns true or false (similar to basic checks), or returns an error structure if the boolean value couldn't even be resolved. This solves the issue of bool checks suppressing error cases

On top of that, I tried to make fragments as self contained as possible. It doesn't matter if they invoke some of the loaders/checks that previous fragments already did - we have caching. This self-contained nature makes them a lot easier to understand and reuse.

With these 2 kinds of fragments, I was able update all existing policies to be shorter and more easily understandable. The `maybeMemberRoleWithValidSsoSessionIfNeeded()` is still 99% what it was before, except now its an `ensureWorkspaceRoleAndSessionFragment()`

packages/shared/src/authz/fragments/workspaces.ts
```diff
@@ -24,47 +27,58 @@ export const maybeMemberRoleWithValidSsoSessionIfNeeded: AuthPolicyFragment<
...
+    if (!hasAnyRole) return err(new WorkspaceNoAccessError())
 
     const hasMinimumMemberRole = await requireMinimumWorkspaceRole(loaders)({
       userId,
       workspaceId,
       role: 'workspace:member'
     })
     // only members and above need to use sso
-    if (!hasMinimumMemberRole) return nothing()
+    if (!hasMinimumMemberRole) return ok()
```
fabis94, line 41: no more `Maybe` results -a fragment only makes a decision for itself, not for the parent policy. it either succeeds or it fails, so no `maybe` necessary

packages/shared/src/authz/policies/workspace/canCreateWorkspaceProject.ts
```diff
@@ -43,44 +49,45 @@ export const canCreateWorkspaceProjectPolicy: AuthPolicy<
...
 
-    const memberWithSsoSession = await maybeMemberRoleWithValidSsoSessionIfNeeded(
-      loaders
-    )({
-      userId,
-      workspaceId
-    })
     // guests cannot create projects in the workspace
-    if (memberWithSsoSession.isNothing)
+    const isNotGuest = await requireMinimumWorkspaceRole(loaders)({
```
fabis94, line 78: The previous logic of a `nothing()` actually being an error case for us was too complicated to reason about. Now `maybeMemberRoleWithValidSsoSessionIfNeeded/ ensureWorkspaceRoleAndSessionFragment` just ensures you have workspace access (any role). And the separate check of not allowing guests is its own thing.

packages/shared/src/authz/policies/project/canUpdate.ts
```diff
@@ -0,0 +1,48 @@
...
+import { MaybeUserContext, ProjectContext } from '../../domain/context.js'
+import { AuthPolicy, ErrorsOf, LoadersOf } from '../../domain/policies.js'
+import { Roles } from '../../../core/constants.js'
+import { ensureMinimumServerRoleFragment } from '../../fragments/server.js'
+import {
+  ensureMinimumProjectRoleFragment,
+  ensureProjectWorkspaceAccessFragment
+} from '../../fragments/projects.js'
+
+export const canUpdateProjectPolicy: AuthPolicy<
```
fabis94, line 11: i will add tests here after, i just wanted to validate my refactors first

packages/shared/src/authz/policies/workspace/canCreateWorkspaceProject.ts
fabis94: that, or ensureWorkspaceRoleAndSessionFragment() could have an optional `role` arg that allows you to specify the minimum role you expect

packages/shared/src/authz/fragments/workspaces.ts
gjedlicska: Ok, with the implicit project role moved into where it is now, this can work.

packages/shared/src/authz/policies/project/canReadProject.ts
```diff
@@ -1,95 +1,76 @@
...
-    const project = await loaders.getProject({ projectId })
-    if (!project) return err(new ProjectNotFoundError())
-
     // All users may read public projects
-    if (await isPubliclyReadableProject(loaders)({ projectId })) return ok()
+    const isPubliclyReadable = await checkIfPubliclyReadableProjectFragment(loaders)({
+      projectId
+    })
+    if (isPubliclyReadable.isErr) {
+      return err(isPubliclyReadable.error)
```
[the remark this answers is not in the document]
gjedlicska: As long as they are explicity defined in the return type of the policy. If we're just blindly accepting and passing along, that is not ok. See above.

gjedlicska (changes requested):
I get the changes and the pattern is evolving in the right way, but we need to have the type guarantees somewhere. ErrorsOf will not work as is. Now that we're just passing the policy result to the client, if someone changes an error in a check / fragment, it immediately lands in the clients, without anyone getting notified by type checking changes. That is not ok.

 And I do not get the need for LoadersOf, its not as bad as errors of, its just an unnecessary abstraction.

fabis94:
got rid of LoadersOf/ErrorsOf

PR #4355, opened 2025-04-09
feat(fe): disable domain protection switch if not on business plan

andrewwallacespeckle:
image


packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
```diff
@@ -187,6 +208,12 @@ const workspace = computed(() => result.value?.workspaceBySlug)
 const workspaceDomains = computed(() => {
   return workspace.value?.domains || []
 })
+const isBusinessPlan = computed(
```
Comment (Mikehrn) [inline packages/frontend-2/pages/settings/workspaces/[slug]/security.vue:211]: There is a query you can do for this :) 

```
query Workspace($featureName: WorkspaceFeatureName!, $workspaceId: String!) {
  workspace(id: $workspaceId) {
    hasAccessToFeature(featureName: $featureName)
  }
}
```

```
{
  "featureName": "domainBasedSecurityPolicies",
}
```

Mikehrn (changes requested):

packages/frontend-2/pages/settings/workspaces/[slug]/security.vue
Comment (andrewwallacespeckle) [reply packages/frontend-2/pages/settings/workspaces/[slug]/security.vue]: Nice, done :)

PR #4357, opened 2025-04-09
Feat: Update pricing plan features

Mikehrn:
Also cleans up some old stuff


packages/frontend-2/components/pricingTable/Plan.vue
```diff
@@ -56,29 +56,45 @@
       </div>
     </div>
     <ul class="flex flex-col gap-y-2 mt-4 pt-3 border-t border-outline-3">
-      <li
+      <PricingTablePlanFeature
```
Mikehrn, line 59: Some of these features a just "cosmetic", only this component cares about rendering, they are not really features so didnt add them to the shared object

packages/shared/src/workspaces/helpers/features.ts
```diff
@@ -97,63 +76,42 @@ export type WorkspacePlanConfig<Plan extends WorkspacePlans = WorkspacePlans> =
...
-  WorkspacePlanFeatures.DomainSecurity,
-  WorkspacePlanFeatures.SSO,
-  WorkspacePlanFeatures.CustomDataRegion,
-  WorkspacePlanFeatures.PrioritySupport
-]
-
 export const WorkspacePaidPlanConfigs: {
   [plan in PaidWorkspacePlans]: WorkspacePlanConfig<plan>
 } = {
   // Old
```
Mikehrn, line 86: If would like to remove these old ones, seems they are not really used on the BE yet, FE doesnt use them anymore, but Im not sure @gjedlicska

gjedlicska: for now we cannot do it, post release, we can clean up all refs to the old plans

gjedlicska (dismissed):

gjedlicska (approved):

PR #4360, opened 2025-04-09
fix(authz): branches are models too

cdriesler:
It's true


packages/server/modules/core/graph/resolvers/branches.ts
```diff
@@ -70,12 +71,14 @@ export = {
   },
   Mutation: {
     async branchCreate(_parent, args, context) {
-      await authorizeResolver(
-        context.userId,
-        args.branch.streamId,
-        Roles.Stream.Contributor,
-        context.resourceAccessRules
-      )
+      const canCreate = await context.authPolicies.project.canCreateModel({
```
gjedlicska, line 74: coordinate with Fabs on this, the token access rules must be checked.

gjedlicska (approved):

PR #4361, opened 2025-04-09
feat(add createFromObjects and createFromJSON to ObjectLoader2)

adamhathcock:
Missing methods from OL2

Added DefermentManager to better test and assure getObject will work
Switched to snapshot testing
Added createFromJSON/createFromObjects which uses Memory implementations of cache and downloads to be static
Added Traverser that will traverse Base objects correctly to build the tree 
Added interfaces needed for the above


AlexandruPopovici:
`getAndConstructObject` might be legacy and unneeded anymore. Unless PowerBI does? @oguzhankoral

oguzhankoral:
We are not using `getAndConstructObject` in PowerBI, but we still use it in `dui3` for the connectors do not have receive mechanism @AlexandruPopovici @adamhathcock

adamhathcock:
> We are not using `getAndConstructObject` in PowerBI, but we still use it in `dui3` for the connectors do not have receive mechanism @AlexandruPopovici @adamhathcock

How many connectors is this?  Do we want to support this going forward?  Just seeing if I should implement this

oguzhankoral:
> How many connectors is this? Do we want to support this going forward? Just seeing if I should implement this

Sketchup and Archicad currently but we will elimate Archicad soon, regardless I can use OL1 in `dui3` if it is a big hussle to implement

AlexandruPopovici (approved):

PR #4363, opened 2025-04-09
fix(authz): allow optional project id

cdriesler:
## Description & motivation

- It's still useful to check if a project move to workspace is allowed even before we know the specific project we're trying to move.

## Changes:

- Allow optional project id


fabis94 (approved):
the same also needs to happen in the Project.permissions checks when workspaceId is unknown, is that gonna be a separate PR?

PR #4368, opened 2025-04-10
feat: comment read/write auth policies in BE & FE

fabis94:
## Checklist:


packages/server/modules/comments/tests/comments.graph.spec.js
```diff
@@ -855,7 +855,7 @@ describe('Graphql @comments', () => {
             [archiveMyComment, true],
             [archiveOthersComment, true],
             [editMyComment, true],
-            [editOthersComment, true],
+            [editOthersComment, false],
```
fabis94, line 858: doesn't make sense that anyone other than the comment owner should be able to edit it. otherwise you can have project/workspace owners impersonating others without a trace

packages/server/modules/comments/tests/comments.graph.spec.js
```diff
@@ -978,8 +978,8 @@ describe('Graphql @comments', () => {
             [archiveOthersComment, false],
             [editOthersComment, false],
             [replyToAComment, false],
-            [queryComment, false],
-            [queryComments, false],
+            [queryComment, true],
```
fabis94, line 981: querying on public projects should be possible even for archived users, that's the same logic we have for Project.canRead

packages/shared/src/authz/fragments/projects.spec.ts
```diff
@@ -16,11 +17,12 @@ import {
   WorkspaceSsoSessionNoAccessError
 } from '../domain/authErrors.js'
 import { OverridesOf } from '../../tests/helpers/types.js'
+import { getProjectFake } from '../../tests/fakes.js'
 
 describe('ensureMinimumProjectRoleFragment', () => {
   const buildSUT = (overrides?: OverridesOf<typeof ensureMinimumProjectRoleFragment>) =>
     ensureMinimumProjectRoleFragment({
-      getProject: async () => ({
+      getProject: getProjectFake({
```
fabis94, line 25: we should use getProjectFake/getComentFake/GetXXXFake everywhere to avoid having to update a bunch of places in code just because we've added a new prop to the underlying resource types

PR #4376, opened 2025-04-10
Feat: Add search for filter values

Mikehrn:
Should still work with longer lists, but I dont have a good model to test, will test on latest

Screenshot 2025-04-10 at 19 32 00


cdriesler (approved):
🔥🔥🔥🔥

PR #4419, opened 2025-04-14
feat(authz): `Workspace.canInvite` and `Project.canInvite`

cdriesler:
## Description & motivation

- Adds these policies _and_ uses them


packages/shared/src/authz/policies/project/canInvite.ts
```diff
@@ -0,0 +1,58 @@
...
+export const canInvitePolicy: AuthPolicy<PolicyLoaderKeys, PolicyArgs, PolicyErrors> =
+  (loaders) =>
+  async ({ userId, projectId }) => {
+    const ensuredServerRole = await ensureMinimumServerRoleFragment(loaders)({
+      userId,
+      role: Roles.Server.User
+    })
+    if (ensuredServerRole.isErr) return err(ensuredServerRole.error)
+
+    const ensuredProjectRole = await ensureMinimumProjectRoleFragment(loaders)({
```
fabis94, line 50: Is this really enough? Don't we also need to check that the user has access to the workspace that the project is in + has an SSO session in it?

I think this entire thing can be replaced w/ `ensureImplicitProjectMemberWithWriteAccessFragment()` w/ the Owner role passed in through args

fabis94 (changes requested):
see comment

packages/shared/src/authz/policies/project/canInvite.ts
cdriesler: Fair, updated to use fragment.

fabis94 (approved):

PR #4433, opened 2025-04-15
fix(fe): Only show "why am I seeing this" when forced workspace creation

andrewwallacespeckle:
image


packages/frontend-2/middleware/004-onboarding.global.ts
```diff
@@ -129,7 +129,10 @@ export default defineNuxtRouteMiddleware(async (to) => {
       return navigateTo(workspaceJoinRoute)
     }
     if (!hasDiscoverableWorkspaces && !isGoingToCreateWorkspace) {
-      return navigateTo(workspaceCreateRoute())
+      return navigateTo({
```
Mikehrn, line 132: This will still show for new users, right? This should only show for existing users that dont have a workspace yet. If you are a new user this should never show

Mikehrn (changes requested):

packages/frontend-2/components/workspace/moveProject/Manager.vue
```diff
@@ -28,11 +28,7 @@
     />
     <template #buttons>
       <div class="-my-1 w-full flex justify-end">
-        <FormButton
-          v-if="!selectedProject"
-          color="outline"
-          @click="navigateTo(workspaceCreateRoute())"
-        >
+        <FormButton v-if="!selectedProject" color="outline" @click="open = false">
```
andrewwallacespeckle, line 31: Fix - unrelated to this branch, but noticed it here

packages/frontend-2/middleware/004-onboarding.global.ts
andrewwallacespeckle: You are right. 

I forgot about the useRegisteredThisSession composable that we use for verify-email page. 

I used this here too, and reverted changes to middleware

Mikehrn (approved):

PR #4438, opened 2025-04-15
fix(fe): Update dismiss button in workspace card

andrewwallacespeckle:


Mikehrn:
You still need to be able to dismiss them. The issue is that the buttons dont work, but they should

Mikehrn (approved):

PR #4444, opened 2025-04-15
chore(server): ADMIN_OVERRIDE_ENABLED flag to FF_ADMIN_OVERRIDE_ENABLED

iainsproat:
## To-do before merge:

- [ ] update relevant deployments to use new feature flag. Merge https://github.com/specklesystems/gitOps/pull/1479

## Checklist:


fabis94:
i don't think we should do this, I was already trying to in a different PR a couple of weeks back and it 1) broke a bunch of tests 2) made testing more complicated:

1) you need to get the original feature flags, because to override one key you still need to return the rest of them 2) you still need to mock adminOverrideEnabled() also, because you can't mock getFeatureFlags() when its referenced directly from adminOverrideEnabled() in the same module 3) if we ever need to mock multiple feature flags, you won't be able to re-use any utilities that just mock a specific key, because it rewrites all of the other feature flags to be set to their defaults
image

i just don't see any real benefit to this and a bunch of downsides

PR #4450, opened 2025-04-16
Feat: Add workspace empty state

michalspeckle:
## Checklist:

### DIFF
--- a/packages/frontend-2/components/workspace/WorkspaceEmptyState.vue
+++ b/packages/frontend-2/components/workspace/WorkspaceEmptyState.vue
@@ -1,0 +1,151 @@
+<template>
+  <svg
+    width="259"
+    height="231"
+    viewBox="0 0 259 231"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      d="M7.49495 103.869L166.728 50.7911C170.436 49.5552 175.387 51.0542 177.786 54.1393L253.207 151.108C255.606 154.193 254.546 157.696 250.838 158.932L91.6048 212.01C87.897 213.246 82.946 211.747 80.5464 208.662L5.12605 111.693C2.72652 108.608 3.78711 105.105 7.49495 103.869Z"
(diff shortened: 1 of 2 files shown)


packages/frontend-2/components/workspace/WorkspaceEmptyState.vue
```diff
@@ -0,0 +1,151 @@
+<template>
+  <svg
```
benjaminvo, line 2: Should this SVG perhaps go in the assets/images folder instead and imported as an asset? I'm not sure what's best practice here.

michalspeckle: I thought about adding it to components/global/emptyStates folder (the plan is to have more of those, right?)
I'm using tailwind colors in the SVG code for path fills/stroke colors so I thought it has to be a vue component with a template.

Mikehrn: `assets` gets processed by the build tools, usually I put assets that don't need to be processed in `public`

andrewwallacespeckle: Had a quick chat with Michal on this one. 

This empty state uses our tailwind classes so it works with dark mode. I'd much rather this than having to export 2 pngs. 

We'll need to keep it as a vue component for this to work. Right now it makes sense to keep it in `/components/workspaces/EmptyState`, but we discussed creating a `/components/empty-states/` when we have more to add.

packages/frontend-2/components/workspace/WorkspaceEmptyState.vue
```diff
@@ -0,0 +1,151 @@
+<template>
```
andrewwallacespeckle, line 1: Only comment from me is a reminder to update the name from `WorkspaceEmptyState` to `EmptyState`

andrewwallacespeckle (approved):

Mikehrn (approved):

PR #4454, opened 2025-04-16
feat(fe) Add Limits dialog to AddProjectMenu

andrewwallacespeckle:


packages/frontend-2/components/workspace/header/AddProjectMenu.vue
```diff
@@ -54,27 +89,41 @@ const props = defineProps<{
...
+        : undefined
     }
   ]
 ])
 
 const onActionChosen = (params: { item: LayoutMenuItem; event: MouseEvent }) => {
   const { item } = params
 
+  if (isLimitReached.value) {
+    limitReachedWorkspace.value = {
```
Mikehrn, line 120: Do you need this typecasting if the structure is the same?

Mikehrn (dismissed):

packages/frontend-2/components/workspace/header/AddProjectMenu.vue
andrewwallacespeckle: Fixed this, I removed this ref completely, it was leftover from another component

Mikehrn (approved):

PR #4464, opened 2025-04-16
fix(gatekeeper): try catch the plan migration

gjedlicska:
makes sure the transaction is canceled and not left in a dangling state

@iainsproat feel free to merge this


gjedlicska:
superseeded by https://github.com/specklesystems/speckle-server/pull/4468/files

PR #4468, opened 2025-04-17
fix(fe2): not being able to remove member from workspace

fabis94:
plus: fix for workspace new plan migration causing knex pool connection leak on failures
plus: fixed withTransaction API, the DX wasn't great

### DIFF
--- a/packages/server/modules/shared/helpers/dbHelper.ts
+++ b/packages/server/modules/shared/helpers/dbHelper.ts
@@ -3,7 +3,7 @@ import { Knex } from 'knex'
 import { isString } from 'lodash'
 import { postgresMaxConnections } from '@/modules/shared/helpers/envHelper'
 import { EnvironmentResourceError } from '@/modules/shared/errors'
-import { isNonNullable } from '@speckle/shared'
+import { isNonNullable, MaybeAsync } from '@speckle/shared'
 
 export type BatchedSelectOptions = {
   /**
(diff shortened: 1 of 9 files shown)


packages/server/modules/shared/helpers/dbHelper.ts
```diff
@@ -106,6 +106,9 @@ export const numberOfFreeConnections = (knex: Knex) => {
   return Math.max(pgMaxConnections - demand, 0)
 }
 
+/**
+ * @deprecated use withOperationTransaction for a better API
```
fabis94, line 110: requires you to manually create the trx beforehand which makes it less useful than it could be. the trx object should be an internal implementation detail, you don't need it outside of the function

Mikehrn (dismissed):

packages/server/modules/shared/helpers/dbHelper.ts
```diff
@@ -107,17 +107,20 @@ export const numberOfFreeConnections = (knex: Knex) => {
 }
 
 export const withTransaction = async <T>(
-  callback: Promise<T> | T,
-  trx: Knex.Transaction
-): Promise<T> => {
+  operation: (args: { db: Knex }) => MaybeAsync<T>,
```
fabis94, line 110: the old implementation requires you to manually create the trx beforehand which makes it less useful than it could be. the trx object should be an internal implementation detail, you don't need it outside of the function.

i also made the API more similar to withOperationLogging, with the goal of possibly merging withTransaction+withOperationLogging+commandFactory together into 1 utility in the future

Mikehrn (approved):

PR #4470, opened 2025-04-17
chore(server): fixed/improved commandFactory

fabis94:
commandFactory replacement that does what it already did, but also:
* makes sure `db` is the actual transaction. this misleading aspect of the original caused many usages to not use the trx correctly
* also merges withOperationLogging into the logic

### DIFF
--- a/packages/server/modules/shared/command.ts
+++ b/packages/server/modules/shared/command.ts
@@ -1,8 +1,15 @@
+import { mainDb } from '@/db/knex'
 import { withTransaction } from '@/modules/shared/helpers/dbHelper'
-import { EmitArg, EventBus, EventBusEmit } from '@/modules/shared/services/eventBus'
+import {
+  EmitArg,
+  EventBus,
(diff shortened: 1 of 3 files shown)


fabis94:
made db & eventBus optional (defaults to main db and main eventBus), also made it so that eventBus transactionality can be disabled, cause sometimes we only need the DB to be a transaction

packages/server/modules/shared/command.ts
```diff
@@ -40,3 +48,75 @@ export const commandFactory =
...
+    eventBus = getEventBus(),
+    logger,
+    name,
+    description,
+    transaction
+  } = params
+
+  return await withOperationLogging(
+    async () => {
+      if (!transaction) {
```
cdriesler, line 94: Possibly true by default? Maybe after wider usage?

cdriesler (approved):
FA8I6ZpXoAIerTO

PR #4473, opened 2025-04-17
fix(fe): "Move this project" warning shown for public projects and non-project owner

andrewwallacespeckle:


packages/frontend-2/pages/projects/[id]/index.vue
```diff
@@ -304,6 +306,18 @@ const activePageTab = computed({
...
 
+const shouldShowWorkspaceAlert = computed(
+  () =>
+    isWorkspacesEnabled.value &&
+    isLoggedIn.value &&
+    !project.value?.workspace &&
+    hasRole.value
+)
+
+const disableLegacyMoveProjectButton = computed(
```
Comment (Mikehrn) [inline packages/frontend-2/pages/projects/[id]/index.vue:317]: We should prob use the canMove permission for this, but we can run with this for now and update that afterwards

Mikehrn (approved):

Mikehrn (dismissed):

Mikehrn (approved):

PR #4475, opened 2025-04-17
refactor(fe): Workspace member actions updates

andrewwallacespeckle:
Always show the seat change in a separate section alone in the top: 

In the admin menu: Hide the "Make guest…" action and move the "Revoke admin access…" up to the top of the second section.

### DIFF
--- a/packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
+++ b/packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
@@ -1,6 +1,6 @@
 <template>
   <section>
-    <div class="md:max-w-5xl md:mx-auto pb-6 md:pb-0">
+    <div class="md:max-w-5xl md:mx-auto pb-16">
       <SettingsSectionHeader
         hide-divider
         title="People"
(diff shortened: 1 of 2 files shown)


packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <section>
-    <div class="md:max-w-5xl md:mx-auto pb-6 md:pb-0">
+    <div class="md:max-w-5xl md:mx-auto pb-16">
```
Comment (andrewwallacespeckle) [inline packages/frontend-2/pages/settings/workspaces/[slug]/members.vue:3]: Added this padding because the actions menu sometimes overflows the viewport leaving unreachable items.

Comment (Mikehrn) [reply packages/frontend-2/pages/settings/workspaces/[slug]/members.vue:3]: Ran into this before as well, we need to make a few improvements to this table, we also need inf loading because some workspace have a lot of members, it's pretty slow now, we can follow up with improvements to this :)

Mikehrn (approved):

PR #4493, opened 2025-04-17
fix(invites): projectinvitablecollaborators should only return members

cdriesler:
## Description & motivation

- Invitable collaborators logic was returning too many members

## Changes:

- Adjust query to return expected list of:
  - Workspace users
  - Without project roles
- Update tests to reflect updated behavior


Mikehrn (approved):
habbo-dance

cdriesler:
ygnnb39p94r91

Mikehrn (approved):

PR #4498, opened 2025-04-18
feat(previews): retry errored previews if feature flag enabled

iainsproat:
## Description & motivation

For various reasons the preview service can return an error response. It will not be retried and the preview will forever remain broken.

This provides a simple retry mechanism. Running as a scheduled task ("cronjob"), every 23 minutes it finds an errored preview (one for each connected regional database). It resets the preview request to pending and queues a job to be processed by the preview service.

A maximum number of attempts is set for each object preview, beyond which we will not retry.

## Changes:

- adds a new feature flag `FF_RETRY_ERRORED_PREVIEWS_ENABLED`
- adds an `attempts` column to the `object_preview` table, to keep track of the number of attempts

## To-do before merge:

- [x] test in a testing environment
   - [x] test it works for no previews to retry
   - [x] test it works for some errored previews to retry

## Screenshots:

Retries until maximum number of attempts is reached:

Screenshot 2025-07-15 at 09 04 42

## Checklist:

### DIFF
--- a/packages/server/modules/previews/tests/integration/retryErrors.spec.ts
+++ b/packages/server/modules/previews/tests/integration/retryErrors.spec.ts
@@ -149,7 +149,6 @@ describe('Previews services @previews', () => {
       expect(maxAttemptsResults.items.map((item) => item.objectId)).to.not.contain(
         originalFirstItem.objectId
       )
-      expect(maxAttemptsResults.items[0].attempts).to.equal(2)
     })
   })
 })


packages/server/modules/core/migrations/20250701164407_object_preview_attempts.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const tableName = 'object_preview'
+const attemptsCol = 'attempts'
+
+export async function up(knex: Knex): Promise<void> {
+  await knex.schema.alterTable(tableName, (table) => {
+    table.integer(attemptsCol).defaultTo(0).notNullable()
+
+    table.dropIndex(['previewStatus', 'priority', 'lastUpdate'])
+    table.index(['previewStatus', 'priority', 'lastUpdate', attemptsCol])
```
danielgak, line 11: https://www.postgresql.org/docs/current/indexes-multicolumn.html

> Multicolumn indexes should be used sparingly. In most situations, an index on a single column is sufficient and saves space and time. Indexes with more than three columns are unlikely to be helpful unless the usage of the table is extremely stylized.

fabis94: @danielgak surely if you're often going to be filtering by X, Y and Z, then a multicolumn index on X, Y and Z is going to be more useful than indices on X, Y and Z individually?

danielgak (approved):

PR #4512, opened 2025-04-21
refactor(fe): members settings usePaginatedQuery

andrewwallacespeckle:


packages/frontend-2/pages/settings/workspaces/[slug]/members.vue
```diff
@@ -21,31 +21,22 @@ import { graphql } from '~/lib/common/generated/gql'
...
-        id
-      }
+      id
     }
     adminWorkspacesJoinRequests {
       items {
         id
         status
       }
-      totalCount
```
Comment (Mikehrn) [inline packages/frontend-2/pages/settings/workspaces/[slug]/members.vue:48]: Will the total count for join requests still work if you remove this?

Mikehrn (approved):

PR #4517, opened 2025-04-22
Feat: Add tooltip to add-on button

Mikehrn:


packages/frontend-2/components/settings/workspaces/billing/addOns/AddOns.vue
```diff
@@ -73,10 +73,19 @@ const contactButton = computed(() => ({
   }
 }))
 
+const addOnButtonTooltip = computed(() => {
+  if (!isAdmin.value)
+    return 'You must be a workspace admin in order to purchase the add-on'
+  if (hasUnlimitedAddon.value) return 'You already have an unlimited add-on'
+  if (!isPaidPlan.value) return 'Only available for starter & business plans'
```
benjaminvo, line 80:
```suggestion
  if (!isPaidPlan.value) return 'Only available for Starter & Business plans'
```

packages/frontend-2/components/settings/workspaces/billing/addOns/AddOns.vue
```diff
@@ -73,10 +73,19 @@ const contactButton = computed(() => ({
   }
 }))
 
+const addOnButtonTooltip = computed(() => {
+  if (!isAdmin.value)
```
benjaminvo, line 77: Should this perhaps be the last check so the other two are shown before this? Even if you are an admin, you still won't be able to add this add-on if the others are tru

packages/frontend-2/components/settings/workspaces/billing/addOns/AddOns.vue
```diff
@@ -73,10 +73,19 @@ const contactButton = computed(() => ({
...
+    return 'You must be a workspace admin in order to purchase the add-on'
+  if (hasUnlimitedAddon.value) return 'You already have an unlimited add-on'
```
benjaminvo, line 79:
```suggestion
  if (hasUnlimitedAddon.value) return 'The add-on is already included in your subscription'
```

benjaminvo (changes requested):

packages/frontend-2/components/settings/workspaces/billing/addOns/AddOns.vue
Mikehrn: Whatever you prefer, in the pricing table "non admin" takes priority over all other states, so I went with the same logic here, but if you want to change the order just lmk

benjaminvo: Ah, hmm. Let's for now just keep to that logic then

benjaminvo (approved):

andrewwallacespeckle (approved):

PR #4524, opened 2025-04-22
fix(fe): Move project card - pass correct id so move dialog has project preselected

andrewwallacespeckle:


packages/frontend-2/components/projects/Dashboard.vue
```diff
@@ -118,7 +118,7 @@ const filterProjectsToMove = ref(false)
 const openNewProject = ref(false)
 const showLoadingBar = ref(false)
 const showMoveProjectDialog = ref(false)
-const emittedProjectId = ref('')
+const emittedProjectId = ref<Nullable<string>>(null)
```
Mikehrn, line 121: Could this not be `ref<string | undefined>(undefined)`  in order to avoid the piping for the prop?

Mikehrn (approved):

PR #4534, opened 2025-04-22
refactor(fe): Disable seat type change if expired or cancelled

andrewwallacespeckle:


andrewwallacespeckle:
@Mikehrn Ticket only mentions cancelled, but I added expired here too - i'm not sure the difference though, let me know if this isn't right.

PR #4561, opened 2025-04-23
Feat: Add seat purchased event

Mikehrn:


packages/frontend-2/components/settings/workspaces/members/actions/UpdateAdminDialog.vue
```diff
@@ -150,6 +152,14 @@ const handleConfirm = async () => {
       workspaceId: props.workspace.id
     })
 
+    if (!isUnpaidPaidUpgrade.value) {
+      mixpanel.track('Workspace Seat Purchased', {
+        location: 'upgrade_admin_dialog',
```
benjaminvo, line 157: Should we for future proofing sake add a "type: editor" in case we add more seat types in the future?

benjaminvo (approved):

andrewwallacespeckle (approved):

PR #4577, opened 2025-04-23
Feat: Sort move project by role

Mikehrn:


packages/frontend-2/components/workspace/moveProject/SelectWorkspace.vue
```diff
@@ -92,7 +92,7 @@ graphql(`
         ...WorkspaceMoveProjectManager_Workspace
       }
     }
-    projects(cursor: $cursor, filter: $filter) {
+    projects(cursor: $cursor, filter: $filter, sortBy: $sortBy) {
```
Mikehrn, line 95: @andrewwallacespeckle We should probably split this query up in two, one for the project and one for the workspaces step, else it can potentially fetch unrelated stuff (and it's a bit confusing the fragment used for projects is in the workspace step)

andrewwallacespeckle (dismissed):

andrewwallacespeckle (approved):

PR #4582, opened 2025-04-23
fix(workspaces): no inferior roles for workspaces owners

cdriesler:
Admins are not peasants


Mikehrn (approved):
helemaal skitterend mooi chef

PR #4594, opened 2025-04-24
fix: blob upload should use auth policies

fabis94:


packages/frontend-2/components/form/file-upload/ProgressRow.vue
```diff
@@ -16,6 +14,12 @@
       :class="[' w-full mt-2', progressBarClasses]"
       :style="progressBarStyle"
     />
+    <div v-if="errorMessage" class="flex">
+      <span class="text-tiny text-danger">
+        {{ errorMessage }}
```
fabis94, line 19: adding errorMessage to failing uploads

andrewwallacespeckle (approved):

PR #4599, opened 2025-04-25
fix(server): project role updates after workspace role/seat changes

fabis94:
- ensured project roles update correctly after both role & seat type changes
- ensured there's never a situation where a project gets its last owner removed because of aforementioned changes
- rewrote seat/role update tests to work with new plans


packages/server/modules/shared/helpers/envHelper.ts
```diff
@@ -374,7 +374,7 @@ export function isEmailEnabled() {
 }
 
 export function postgresMaxConnections() {
-  return getIntFromEnv('POSTGRES_MAX_CONNECTIONS_SERVER', '4')
+  return getIntFromEnv('POSTGRES_MAX_CONNECTIONS_SERVER', '8')
```
fabis94, line 377: i was running into connection errors just because of a few assignToWorkspace() calls wrapped in a Promise.all. 4 is a ridiculously small number, and 8 is also, but it does resolve the issue I had

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
```diff
@@ -727,37 +726,35 @@ export = FF_WORKSPACES_MODULE_ENABLED
           })
 
           if (!role) {
-            // this is currently not working with the command factory
-            // TODO: include the onWorkspaceRoleDeletedFactory listener service
-            await withOperationLogging(
```
fabis94, line 732: @cdriesler  this approach where we wrap DB calls in a trx and yet push through eventBus emits outside a transaction is a hack that we should definitely avoid - either both db & eventEmit should be in a transaction, or none of them should.

making eventEmit emit before the trx has commited made GQL subs fire before the operation was commited, for example. event bus listeners have no idea whether they were fired from within a DB trx or not, so you can't really avoid this unless if we'd start feeding in the trx object through event emit calls which would become very messy very fast (believe me, I tried)

i found the following to be a better approach: anything happening in event listeners (secondary activities) should assume that the primary activity succeeded and its success is irreversible. what this means for workspaceRole updated/deleted handlers is that they should not ever throw if the person being removed from the workspace is the last owner of the project. instead, the handlers should recover the situation by assigning a new owner - one of the workspace admins

packages/server/modules/workspaces/services/management.ts
```diff
@@ -493,10 +495,10 @@ export const updateWorkspaceRoleFactory =
           workspaceId,
           role: nextWorkspaceRole
         },
+        updatedByUserId,
         flags: {
           skipProjectRoleUpdatesFor: skipProjectRoleUpdatesFor ?? []
```
fabis94, line 500: @cdriesler we can't get rid of skipProjectRoleUpdatesFor and the old role assignment logic yet, otherwise a bunch of tests that use old plans break. i'm handling that in a separate PR, so this PR should only fix logic for new plans without breaking logic for old ones

packages/server/modules/workspaces/graph/resolvers/workspaces.ts
fabis94: generally speaking I am curious about being able to pass through a transaction context down into event listeners, if we could do that with nice DX, we could actually wrap all of the logic (direct or handled in event bus listeners) in transactions which would be nice.

but there does not appear to be a simple way of doing so. it requires a bunch of changes all over the place. you need everything to be transactifiable - DB calls, event bus emits, GQL sub publication etc.

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -256,29 +262,73 @@ export const onWorkspaceAuthorizedFactory =
...
+  }: {
+    acl: { userId: string; workspaceId: string }
+    updatedByUserId: string
+  }) => {
+    // Resolve a fallback admin
+    const [admin] = await deps.getWorkspaceCollaborators({
+      workspaceId,
+      limit: 1,
+      filter: {
+        roles: [Roles.Workspace.Admin],
```
alemagio, line 284: is it fine to pick one random Admin?
I mean, even if it is fine on a business perspective I would still add a sorting criteria if it is not too much work.

alemagio (approved):
@fabis94 LGTM I approved but I see there are explicit points for Chuck.
So in case you want to merge it is approved, otherwise you can wait for him

packages/server/modules/workspaces/events/eventListener.ts
fabis94: If all admins are already owners in practice, then is it really that important? cc @benjaminvo

fabis94: this is essentially a technical implementation detail that we only do to ensure integrity. it makes more sense that we resolve the issue of having to have explicit project owners, than start emailing people about this odd behavior

PR #4604, opened 2025-04-25
featu(fe): Add limit_type to Limit Reached Dialog Viewed mixpanel event

andrewwallacespeckle:


packages/frontend-2/components/workspace/plan/ProjectModelLimitReachedDialog.vue
```diff
@@ -80,7 +80,9 @@ watch(dialogOpen, (value) => {
       type: props.type,
       location: props.location,
       // eslint-disable-next-line camelcase
-      workspace_id: props.workspaceSlug
+      workspace_id: props.workspaceSlug,
+      // eslint-disable-next-line camelcase
+      limit_type: props.type || 'project/model'
```
andrewwallacespeckle, line 85: There are sometimes when we don't know if it is the project or model limit hit. We just know a limit was hit. In this case we don't pass the prop, and it will fallback to project/model

packages/frontend-2/components/viewer/limits/Dialog.vue
```diff
@@ -132,7 +132,9 @@ watch(dialogOpen, (value) => {
       type: props.limitType === 'version' ? 'version' : 'model',
       location: 'viewer',
       // eslint-disable-next-line camelcase
-      workspace_id: props.workspaceSlug
+      workspace_id: props.workspaceSlug,
+      // eslint-disable-next-line camelcase
+      limit_type: props.limitType
```
Mikehrn, line 137: Could you make this limitType? snake_case shouldnt be use for event property names, the workspace_id is the outlier in this case

PR #4612, opened 2025-04-25
chore(server/previews): handle case where project deleted while preview generated

iainsproat:
## Description & motivation

There is an edge case where the previews module receives a response from the preview service, but the project to which the response relates cannot be found - most likely because it has been deleted.

This PR prevents false positive errors being thrown & logged.

## Changes:

- update preview record with response, not upsert. If the project no longer exists, the preview record was likely cascade deleted. Plus if the project is not present, we cannot determine which region to store the preview record in.
- improve response handling of previews
- adds test
- refactors previews module
  - makes the creation of the request and response queue a little more generic, and pushes function into client directory
  - move some factory instantiation out of the service, allowing the service to be unit tested more easily

## Checklist:

### DIFF
--- a/packages/server/modules/previews/services/responses.ts
+++ b/packages/server/modules/previews/services/responses.ts
@@ -5,6 +5,7 @@ import type {
   BuildConsumePreviewResult,
   ObserveMetrics
 } from '@/modules/previews/domain/operations'
+import { StreamNotFoundError } from '@/modules/core/errors/stream'
 import { fromJobId, previewResultPayload } from '@speckle/shared/workers/previews'
 
 const parseMessage = (data: string) =>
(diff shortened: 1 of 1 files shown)


packages/server/modules/previews/services/responses.ts
```diff
@@ -0,0 +1,91 @@
...
+        projectId,
+        objectId,
+        previewResult: parsedPayload
+      })
+    } catch (e) {
+      const err = ensureError(e, 'Unknown error when consuming preview result')
+
+      if (
+        err instanceof StreamNotFoundError ||
+        (err instanceof DatabaseError &&
```
danielgak, line 72: if on `consumePreviewResult` we always do updates, we never should have `StreamNotFoundError` bc the update just will be silent right? and either a foreign key violation 🤔

danielgak (dismissed):
Generally it looks good, for what I understood (but maybe I'm wrong here) I had trouble understanding how the StreamNotFoundError, of the foreign key violation happens if we only do updates

danielgak (dismissed):

PR #4615, opened 2025-04-27
Fix: Minor fixes for canceled payment state

Mikehrn:
Fixes some minor things I found while testing the wizard state after canceling the Stripe payment. @benjaminvo I still cant on any env or browser reproduce the bug where it gets stuck in the loading state..


packages/frontend-2/pages/workspaces/[id]/create.vue
Comment (Mikehrn) [inline packages/frontend-2/pages/workspaces/[id]/create.vue:1]: This is old/redundant, this is from when we originally still supported continuing with the wizard later

benjaminvo (approved):

PR #4616, opened 2025-04-27
Feat: Bulk invites for projects and workspaces

Mikehrn:
@benjaminvo After some thinking, I thought of a way to handle the pasted input rather than making it a seperate flow. It allows you to paste a list of emails in any input (it will perserve other input you already have), parse the list and get all valid emails from it

Screenshot 2025-04-27 at 22 28 20
Screenshot 2025-04-27 at 22 26 23


benjaminvo:
Nice! I'll test later today

benjaminvo (dismissed):
Really nice!

PR #4624, opened 2025-04-29
chore: get rid of all old workspace plan code

fabis94:
## Checklist:


packages/dui3/lib/common/generated/gql/graphql.ts
```diff
@@ -1913,11 +1913,8 @@ export type OnboardingCompletionInput = {
 };
 
 export enum PaidWorkspacePlans {
-  Business = 'business',
-  Plus = 'plus',
   Pro = 'pro',
```
fabis94, line 1916: Backwards incompatible GQL API change, hopefully we're fine with that (we can't remove old constants without doing this)

packages/server/modules/core/repositories/users.ts
```diff
@@ -508,9 +524,32 @@ export const lookupUsersFactory =
 
     // limit to given project
     if (projectId) {
+      // Workspace implicit roles logic:
```
fabis94, line 527: fix for users() query not supporting implicit workspace roles

packages/server/modules/gatekeeper/domain/billing.ts
```diff
@@ -131,7 +124,7 @@ export const SubscriptionData = z.object({
   status: z.union([
     z.literal('incomplete'),
     z.literal('incomplete_expired'),
-    z.literal('trialing'),
+    z.literal('trialing'), // TODO: Should we get rid of trial related states?
```
fabis94, line 127: should we? this is about the stripe type, not ours

packages/server/modules/gatekeeperCore/migrations/20250424064652_migrate_old_plans.ts
```diff
@@ -0,0 +1,37 @@
+import { Knex } from 'knex'
+
+/**
+ * The full stripe+db migration should've already executed, this is a fallback for dev/test envs to migrate broken plans
```
fabis94, line 4: we're already "sure" that nothing is on old plans, so this should be safe to have

packages/server/modules/workspaces/services/invites.ts
```diff
@@ -223,7 +228,16 @@ export const collectAndValidateWorkspaceTargetsFactory =
       await deps.validateWorkspaceMemberProjectRoleFactory({
         workspaceId,
         userId: targetUser.id,
-        projectRole
+        projectRole,
+        workspaceAccess: workspaceRoleAndSeat
```
fabis94, line 232: fixed issue where invites were not validating against the planned role & seat (we were only checking against the real/existing one)

packages/shared/src/workspaces/helpers/plans.ts
```diff
@@ -62,38 +28,6 @@ export const WorkspacePlans = <const>{
 
 export type WorkspacePlans = (typeof WorkspacePlans)[keyof typeof WorkspacePlans]
 
-// TODO: Remove this post workspace migration
-export const WorkspaceGuestSeatType = 'guest'
```
fabis94, line 66: we no longer need `guest` either, I assume

Mikehrn (approved):
FE looks good :)

packages/server/modules/core/tests/integration/objects.rest.spec.ts
```diff
@@ -108,7 +108,7 @@ describe('Objects REST @core', () => {
         slug: ''
       }
       await createTestWorkspace(workspace, user, {
-        addPlan: { name: 'business', status: 'expired' }
+        addPlan: { name: PaidWorkspacePlans.Team, status: 'canceled' }
```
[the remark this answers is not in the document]
fabis94: controversial topic. some members of the team prefer using magic strings and don't want to stop doing it. i only changed these in some places where I explicitly had to make changes, but intentionally didn't change everything

gjedlicska:
@fabis94 i've not fully gone through it but conceptually we're in the right direction. we can delete all remnants of the old plan names. Also, trial state should go too

fabis94:
@gjedlicska ok let me know when ur done with a full review. all of the changes are done on my end, no further edits planned, so it's only about validating the PR and the questions I raised

packages/server/modules/gatekeeperCore/migrations/20250424064652_migrate_old_plans.ts
gjedlicska: Why do we need this migration?
Prod should be up to date, is it for local envs?

packages/server/modules/core/repositories/users.ts
fabis94: when looking up project collaborators in mentions & i think invite dialogs as well?

packages/server/modules/gatekeeperCore/migrations/20250424064652_migrate_old_plans.ts
fabis94: yes, see comment

gjedlicska (dismissed):

gjedlicska (approved):

PR #4631, opened 2025-04-30
Fix: Always show reset section box tool reset CTA when enabled

Mikehrn:


Mikehrn:
Closing this for now, needs a bit more thought on how this UX works

PR #4637, opened 2025-04-30
Feat: Set onboarding questions to user profile

Mikehrn:


packages/frontend-2/lib/auth/composables/onboarding.ts
```diff
@@ -101,6 +101,12 @@ export const useProcessOnboarding = () => {
     const user = activeUser.value
     if (!user) throw new OnboardingError('Attempting to onboard unidentified user')
 
+    mixpanel.people.set_once({
+      onboardingSurveyRole: onboardingData?.role,
+      onboardingSurveyPlans: onboardingData?.plans,
```
Mikehrn, line 106: Some of these answers are very long, like "Business intelligence (data analytics and visualisation)". Maybe we just create an unique key for each of them that we use rather than the actual text?

Mikehrn: Nvm these are actually already the unique keys

benjaminvo (approved):

PR #4640, opened 2025-04-30
Feat: Enable Intercom for in-app support

benjaminvo:
Experimenting with enabling Intercom for in-app support (not confirmed we want to do it).

- Created a new plugin for initializing Intercom
- Trigger the Intercom chat bubble from the "Give us feedback" actions in the sidebar and in the account dropdown
- Send workspace name, id, and plan to Intercom. It's created as "Companies" in Intercom as that's their built in logic. Luckily a user can be in multiple companies so I watch for the currently active workspace and send that + resend when the active workspace changes.  
- Added a feature flag to only enable Intercom if workspaces are enabled on the server.

Still to do:
1. Check logic for initializing and shitting down Intercom + sending active workspace
2. Send server id as an attribute to company and user
3. Send project, model, version count to companies and users.
4. Don't show the chat bubble when in the viewer and in the user creation flow.
5. Remove pricing announcement so it doesn't clash with Intercom 
6. Introduce a feature flag for only enabling Intercom on app.speckle.systems. We don't want to send a lot of crap data from all our local and test environments.
8. Send new chats to new #intercom Discord channel using Zapier.  

CleanShot 2025-05-01 at 00 02 25@2x


Mikehrn:
> Is this feature flagged (or behind an existing workspace feature flag)?

It is, and we also need to move the app_id to a FF flag, so it should only show when workspaces are enabled and the intercom env var is set

Mikehrn (dismissed):

Mikehrn (approved):

Mikehrn (approved):

PR #4648, opened 2025-05-01
WIP: chore(package.json): do not use yarn lifecycle hooks

iainsproat:
Running the yarn lifecycle hook (`postinstall`) for DUI3 & FE2 were slowing down CI by 1 minute each time `yarn` was called.

We only need the DUI3 & FE2 `postinstall` commands for the CircleCI jobs that actually use them.

By removing the yarn lifecycle hooks for DUI3 & FE2, we now have to explicitly call the underlying commands prior to building these packages.

README is updated as additional manual step is required to install husky.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -441,6 +441,14 @@ jobs:
             - .yarn/cache
             - .yarn/unplugged
       - run:
+          name: Run prebuild hook for dui3
+          working_directory: 'packages/dui3'
+          command: 'yarn prebuild'
+      - run:
+          name: Run prebuild hook for frontend
+          working_directory: 'packages/frontend-2'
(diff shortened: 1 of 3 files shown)


fabis94 (commented):
i dont know, these steps are there for a reason. its not just a post-build step, its a post-install step because it (re-)generates nuxt types that, if not there or invalid, would not even allow you to build the apps.

at the very least i'd still want this to happen everywhere outside of CI, and CI could have some kind of env var based exemption

fabis94:
i would rather us look into why this step takes so long and/or optimizing/fixing caching (maybe once we move to github actions?) so that these installs are properly cached and dont have to happen over and over again in each CI step

PR #4664, opened 2025-05-05
Feat: Allow adding members directly from workspace members section

Mikehrn:
Screenshot 2025-05-05 at 17 25 37


packages/frontend-2/components/project/page/collaborators/generalAccess/Members.vue
```diff
@@ -0,0 +1,194 @@
...
+const onAddClick = async (
+  userId: string,
+  workspaceRole: MaybeNullOrUndefined<string>,
+  userName: string
+) => {
+  await createInvite(
+    projectId.value,
+    [
+      {
+        role: Roles.Stream.Reviewer,
```
andrewwallacespeckle, line 177: Just checking this is correct - is user always set to reviewer?

andrewwallacespeckle (approved):
1 small question

packages/frontend-2/components/project/page/collaborators/generalAccess/Members.vue
Mikehrn: Yes :) We want to always add these as reviewer; same as the invite modal

benjaminvo (approved):

PR #4665, opened 2025-05-05
feat(workspaces): added cron job to delete stale workspaces

danielgak:
## Description & motivation

When a workspace creation fails for whatever reason, we leave it in an incomplete state. 
These can never be resumed, as it stands currently and they will pollute the database.

This workspaces also use a slug that can be reused, and as it does not show to the customer, the slug acts as "used".

## Changes:

- Adding to workspace module a cronjob to trigger workspace deletion
- Select query for workspaceId 
- Service useCase

## To-do before merge:

- Add coverage 

## Checklist:


danielgak:
Since last month, there were some new incomplete workspaces - none of them have projects or any relevant info.

packages/server/modules/workspaces/domain/operations.ts
```diff
@@ -83,6 +83,10 @@ export type GetWorkspacesBySlug = (args: {
   userId?: string
 }) => Promise<WorkspaceWithOptionalRole[]>
 
+export type GetWorkspacesNonComplete = (args: {
```
fabis94, line 86: (minor issue, can stay as is) Incomplete*

packages/server/modules/workspaces/services/workspaceCreationState.ts
```diff
@@ -0,0 +1,30 @@
...
+    const workspaces = await getWorkspacesNonComplete({
+      createdAtBefore: thirtyMinutesAgo.toDate()
+    })
+    if (!workspaces?.length) return
+
+    const workspaceIds = workspaces.map((workspace) => workspace.workspaceId)
+    logger.info({ workspaceIds }, 'Deleting non complete workspaces')
+
+    for (const workspaceId of workspaceIds) {
+      await deleteWorkspace({ workspaceId })
```
fabis94, line 28: Why do we retrieve and then delete, instead of just deleting immediately?

fabis94 (approved):

PR #4704, opened 2025-05-12
feat: "workspace" project visibility

fabis94:
1. finally introduced streams.visibility column. isPublic/isDiscoverable to be removed after
2. migrated workspace project 'private' visibility to now be 'workspace'. 'private' now will always mean 'only accessible to explicit collaborators' irregardless of whether its a workspace or non-workspace project
3. new 'workspace' visibility means - only accessible/visible to workspace members. same as what 'private' was before for workspace projects


packages/server/modules/core/graph/resolvers/streams.ts
```diff
@@ -312,6 +323,10 @@ export = {
   },
 
   Stream: {
+    isPublic(parent) {
+      return parent.visibility === ProjectRecordVisibility.Public
```
fabis94, line 327: @JR-Morgan this PR may require adjustments to SDK tests. Project.visibility UNLISTED will now return PUBLIC instead. Other than that the API should be more or less the same.

benjaminvo:
@fabis94 Do you also make sure to change a personal project's visibilty to Workspace if it's Private when moved into the workspace? I haven't had time to test and can't see it from the code

fabis94:
@benjaminvo yes https://github.com/specklesystems/speckle-server/pull/4704/files#diff-3cccc94bcd2adca5d1d87b7cd6a9496ff9ae095c2a20bc09e147c33a6d9ca549R214

packages/server/modules/core/graph/resolvers/streams.ts
fabis94: this is about access not about where you're able or not able to find them - whichever way you find the project, through link or through your dashboard, its accessible by the public (literally anyone).

public/private terminology is way more intuitive when it comes to access levels than 'unlisted'

fabis94: where the project is listed or not should be a separate setting altogether, like how we may want to introduce link shareability for any visibility through a URL token or something

andrewwallacespeckle (approved):

PR #4710, opened 2025-05-12
trying to add a separate geometry conversion progress as well as data loading progress

adamhathcock:
progress is more than loading the data!

### DIFF
--- a/packages/viewer-sandbox/src/main.ts
+++ b/packages/viewer-sandbox/src/main.ts
@@ -111,7 +111,7 @@ const getStream = () => {
   return (
     // prettier-ignore
     // Revit sample house (good for bim-like stuff with many display meshes)
-    // 'https://app.speckle.systems/streams/da9e320dad/commits/5388ef24b8'
+    'https://app.speckle.systems/streams/da9e320dad/commits/5388ef24b8'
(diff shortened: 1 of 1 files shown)


AlexandruPopovici (approved):

packages/viewer/src/modules/tree/RenderTree.ts
```diff
@@ -29,10 +29,13 @@ export class RenderTree {
...
 
-  public buildRenderTree(geometryConverter: GeometryConverter): Promise<boolean> {
+  public buildRenderTree(
+    geometryConverter: GeometryConverter,
+    callback: () => void
+  ): Promise<boolean> {
     const p = this.tree.walkAsync((node: TreeNode): boolean => {
       let start = performance.now()
-      const rendeNode = this.buildRenderNode(node, geometryConverter)
+      const rendeNode = this.buildRenderNode(node, geometryConverter, callback)
```
AlexandruPopovici, line 38: You'll need to be careful what this callback does. Large streams have hundreds of thousands of nodes that get built

adamhathcock: It's mean to just be a callback for the event and calculation.  I don't readily see a different way to ensure it's not used for nefarious things.

AlexandruPopovici (approved):

PR #4732, opened 2025-05-13
feat(fe): Disable file upload menu item in some cases

andrewwallacespeckle:
Disable the "Upload new version..." menu item when:
- User doesn't have write permissions
- Workspace has been cancelled


packages/frontend-2/components/project/page/models/Actions.vue
```diff
@@ -105,6 +110,8 @@ const { copy } = useClipboard()
 const menuId = useId()
 const { isLoggedIn } = useActiveUser()
 const router = useRouter()
+const mp = useMixpanel()
+const { statusIsCanceled } = useWorkspacePlan(props.project.workspace?.slug || '')
```
andrewwallacespeckle, line 114: @Mikehrn Is it ok calling useWorkspacePlan here? 

It relies on the `useIsBillingIntegrationEnabled` FF, so this shouldn't cause any problems for self hosters, right?

Mikehrn (approved):

PR #4737, opened 2025-05-14
feat: increase resources for publish npm job

danielgak:
## Description & motivation

Job sometimes fails as the machine runs out of memory

## Changes:

- Increased machine size for specific job while we work in the more general topic


danielgak:
https://github.com/specklesystems/speckle-server/pull/4736

PR #4749, opened 2025-05-15
feat(workspaces): backfill workspace tracking

danielgak:
## Description & motivation

- Some data gets obsolete, as for a workspace sync to happen there is only eventListeners for very specific workspace events.
- As a quick fix, we are addding a batch update for all workspaces to keep data in sync specially for those entities that are not in workspaces: models, projects etc


packages/server/modules/workspaces/services/tracking.ts
```diff
@@ -0,0 +1,229 @@
...
+    while (true) {
+      const batchedWorkspaces = await deps.getAllWorkspaces({ cursor, limit: 25 })
+      cursor = batchedWorkspaces.cursor
+
+      if (!batchedWorkspaces.items.length) return
+
+      await Promise.all(
+        batchedWorkspaces.items.map(bultPropertiesAndPushThenToMixpanel)
+      )
+    }
```
danielgak, line 191: Any better idea how to "elegantly" do this without adding a `while(true)` loop?

packages/server/modules/workspaces/services/tracking.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    let cursor = null
+    let items = []
+    do {
+      const batchedWorkspaces = await deps.getAllWorkspaces({ cursor, limit: 25 })
+      cursor = batchedWorkspaces.cursor
+      items = batchedWorkspaces.items
+
+      await Promise.all(items.map(bultPropertiesAndPushThenToMixpanel))
+    } while (cursor && items.length)
```
danielgak, line 189: not sure if there is a "safer" way of doing this 🤔

packages/server/modules/workspaces/index.ts
```diff
@@ -10,10 +11,22 @@ import { initializeEventListenersFactory } from '@/modules/workspaces/events/eve
...
+import {
+  acquireTaskLockFactory,
+  releaseTaskLockFactory
+} from '@/modules/core/repositories/scheduledTasks'
+import { scheduleUpdateAllWorkspacesTracking } from '@/modules/workspaces/services/tracking'
+import { getClient } from '@/modules/shared/utils/mixpanel'
 
-const { FF_WORKSPACES_MODULE_ENABLED, FF_WORKSPACES_SSO_ENABLED } = getFeatureFlags()
+const {
+  FF_WORKSPACES_MODULE_ENABLED,
```
danielgak, line 23: I just realised that this only works in GA, what index module can I use so I do not have the `FF_WORKSPACES_MODULE_ENABLED` restriction?

packages/server/modules/shared/test/hepers/creation.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable camelcase */
```
danielgak, line 1: fake implementation of Mixpanel to inject it and read results

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -601,9 +546,9 @@ export const workspaceTrackingFactory =
         })
         if (!updatedPlanWorkspace) break
         mixpanel.groups.set(
-          'workspace_id',
+          WORKSPACE_TRACKING_ID_KEY,
```
danielgak, line 549: avoid magic strings

packages/server/modules/workspaces/index.ts
```diff
@@ -41,13 +54,27 @@ const workspacesModule: SpeckleModule = {
     if (FF_WORKSPACES_SSO_ENABLED) app.use(getSsoRouter())
 
     if (isInitial) {
+      const mixpanel = getClient()
+      const scheduleExecution = scheduleExecutionFactory({
+        acquireTaskLock: acquireTaskLockFactory({ db }),
+        releaseTaskLock: releaseTaskLockFactory({ db })
+      })
+
+      if (FF_BILLING_INTEGRATION_ENABLED && mixpanel)
```
danielgak, line 63: for envs where there is no mixpanel, the cron is not needed

packages/server/modules/workspaces/services/tracking.ts
```diff
@@ -0,0 +1,229 @@
...
+            db
+          }
+        )
+      }),
+      getWorkspacesProjectCount: getWorkspacesProjectsCountsFactory({ db }),
+      getWorkspaceSeatCount: getWorkspaceSeatCountFactory({ db }),
+      getAllWorkspaces: getAllWorkspacesFactory({ db })
+    })
+
+  const dailyAt2AM = '0 2 * * *'
```
danielgak, line 222: After talking with benjamin, let start by daily schedule

packages/server/modules/workspaces/tests/integration/tracking.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+          }
+        )
+      }),
+      getWorkspacesProjectCount: getWorkspacesProjectsCountsFactory({ db }),
+      getWorkspaceSeatCount: getWorkspaceSeatCountFactory({ db }),
+      getAllWorkspaces: getAllWorkspacesFactory({ db })
+    })
+
+  before(async () => {
+    await truncateTables([Workspaces.name])
```
danielgak, line 59: ci had 2 extra workspaces, so this test was failing

gjedlicska (approved):

packages/server/modules/workspaces/services/tracking.ts
cdriesler: Almost exactly this, but I think we tend to wrap them in generator functions (I defer to @fabis94 on this). See `queryAllWorkspaceProjects` as a reference

packages/server/modules/workspaces/services/tracking.ts
```diff
@@ -0,0 +1,229 @@
...
+      countWorkspaceRole: deps.countWorkspaceRole,
+      getDefaultRegion: deps.getDefaultRegion,
+      getWorkspacePlan: deps.getWorkspacePlan,
+      getWorkspaceSubscription: deps.getWorkspaceSubscription,
+      getWorkspaceModelCount: deps.getWorkspaceModelCount,
+      getWorkspacesProjectCount: deps.getWorkspacesProjectCount,
+      getWorkspaceSeatCount: deps.getWorkspaceSeatCount
+    })
+
+    const bultPropertiesAndPushThenToMixpanel = async (workspace: Workspace) => {
```
cdriesler, line 173: nit: typo `buildPropertiesAndPushThemToMixpanel`

cdriesler (changes requested):
LGTM, two small changes then ready to approve

cdriesler:
I was too slow lmao

packages/server/modules/workspaces/services/tracking.ts
fabis94: 1. while() loops should always have a failsafe, in case there's a bug and the condition keeps returning `true` forever. the failsafe can be stopping after X pages and logging an error. X can be a high, practically unreachable amount, and it would still ensure that the app doesn't get stuck forever in case of an error
2. Generators are nice DX, but there's batching here anyway, so they're not absolutely necessary.

fabis94 (commented):
minor issues, looks good for the most part

danielgak:
@fabis94 @cdriesler https://github.com/specklesystems/speckle-server/pull/4764/files

PR #4753, opened 2025-05-15
Fix: Enable grasshopper in connectors

Mikehrn:


packages/frontend-2/components/connectors/Card.vue
```diff
@@ -4,12 +4,19 @@
...
-          v-if="connector.image"
-          :src="connector.image"
-          :alt="`${connector.title} logo`"
-          class="w-[48px] -ml-1"
-        />
+        <div v-if="connector.images" class="relative flex items-start mr-2">
+          <div
+            v-for="(image, index) in connector.images"
+            :key="image"
+            :class="[
```
Mikehrn, line 11: We can probably improve this, but somebody needs to improve these icons

oguzhankoral: yup, you are the boss

PR #4766, opened 2025-05-19
ci(docker): use registry instead of workspace

iainsproat:
## Description & motivation

In https://github.com/specklesystems/speckle-server/pull/4652 we established that CircleCI workspaces were a bottleneck for high IO processes.

Additionally, using the CircleCI workspace to store Docker images instead of a registry means that we are not benefiting from layer de-duplication and similar features of registries.

As a result, this PR experimented with replacing the mechanism for transferring built Docker images between CircleCI jobs. Instead we push to a Docker registry.

## Changes:

- no longer store built Docker images to CircleCI workspace; instead we push to Docker registry
  - from our experiments, the time taken is approximately equivalent.
- no longer load Docker images in to the Docker cache of the publish steps; instead we pull from Docker registry
  - we expect the time taken to be approximately equivalent.
- no longer load Docker images to the Kind cluster (for deployment testing job) from CircleCI workspace; instead we pull from Docker registry
  - we experienced approximately 120 second reduction in duration of deployment testing job.
- build step now publishes on every run (this was already been done for the Postgres container image, so we know the Docker tag publishing mechanism is robust to this use)

Note that this may not translate to a reduction of 120 seconds for the total pipeline, as the deployment testing job may no longer be on the critical path for the entirety of this period.

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -925,7 +925,7 @@ jobs:
           version: default
           docker_layer_caching: true
       - run:
-          name: Build and Save
+          name: Build and Push
           command: ./.circleci/build.sh
 
   docker-build-server:
@@ -954,7 +954,7 @@ jobs:
           version: default
           docker_layer_caching: true
       - run:
-          name: Build and Save
+          name: Build and Publish sourcemaps
           command: ./.circleci/build_publish_fe2_sourcemaps.sh
 
   docker-build-previews:


.circleci/deployment/Tiltfile.load
```diff
@@ -1,21 +1,19 @@
-# In CI environment we expect the Images to exist as tarball files in the workspace
```
danielgak, line 1: why we where doing this in the first place? 🤔

danielgak (approved):

PR #4771, opened 2025-05-20
fix(server): Add user to mailchimp onboarding audience if they are not in it

andrewwallacespeckle:


packages/server/modules/auth/services/mailchimp.ts
```diff
@@ -55,18 +53,8 @@ export async function updateMailchimpMemberTags(
   initializeMailchimp()
   const subscriberHash = md5(user.email.toLowerCase())
 
-  // Check if user is already in audience (meaning they consented to marketing emails)
```
andrewwallacespeckle, line 58: Do you think this is ok? We now add all users to the onboarding list when they complete the segmentation questions. (This will only add members who aren't already on the list).

I removed the try catch, but maybe we still want this too?

gjedlicska: If it works for users already on the list, it's fine.

packages/server/modules/auth/services/mailchimp.ts
```diff
@@ -55,18 +53,8 @@ export async function updateMailchimpMemberTags(
...
-    throw new MailchimpResourceError(
-      'User not found in Mailchimp audience. They should have been added during registration.',
-      {
-        info: { userEmailHash: subscriberHash },
-        cause: ensureError(e, 'Mailchimp API error')
-      }
-    )
-  }
+  // Add user to audience if they are not already in it
+  await addToMailchimpAudience(user, listId)
```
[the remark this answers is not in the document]
andrewwallacespeckle: image
Yeah, we are using `setListMember` instead of `addListMember` which will only add the user when they don't exist.

PR #4786, opened 2025-05-21
feat(gatekeeper): add updatedAt field to workspace plans

danielgak:
## Description & motivation

Right now we do not record how long the plan was active since, as a change or workspace_plan upserts and does not update the createdAt field since it was created, and subscriptions get updated monthly/yearly to update the `currentBillingCycleEnd`;

## Changes:

- Added migration to table and default value
- Updated type and all tests 
- Add the field to mixpanel update


danielgak:
Tested locally with stripe integration

gjedlicska (approved):

PR #4799, opened 2025-05-22
feat(fe): placeholder updates to workspace settings

andrewwallacespeckle:


andrewwallacespeckle:
@Mikehrn There may be some copy changes but maybe we can go ahead with the code review so it's ready to go?

@benjaminvo I've updated the screenshots in this ticket with the current copy. Let me know if there are any changes or anything blocking this going in. This is just the temporary branch and it'll be replaced with the new settings page, so we could do a review of copy in that branch

packages/frontend-2/components/settings/workspaces/security/ConfirmJoinPolicyDialog.vue
```diff
@@ -0,0 +1,49 @@
+<template>
```
Mikehrn, line 1: We should standardize some of these modals, most of them are just a title + text and buttons, so we can easily create a generic one. We do have one for unsaved changes but it has hardcoded copy, we could update that one. Not blocking for this one tho

Mikehrn (approved):

andrewwallacespeckle:
> I updated the copy and also moved the domain protection setting to be last.
> 
> @andrewwallacespeckle I discovered a weird behavior: If I remove the verified domain, then domain discoverability and auto joining still appear enabled, but toggles are disabled. When I re-add the domain, the domain discoverability is disabled while auto-joining keeps being enabled.
> 
> Should we just disable all of the domain features if you remove the last domain?
> 
>  CleanShot.2025-05-27.at.15.37.42.mp4

Yeah, think that's the best solution. I'll do that now

packages/frontend-2/lib/workspaces/composables/discoverableWorkspaces.ts
```diff
@@ -140,30 +145,52 @@ export const useDiscoverableWorkspaces = () => {
...
-        description: 'Your request to join the workspace has been sent.',
-        type: ToastNotificationType.Success
-      })
+        triggerNotification({
+          title: 'Workspace joined',
+          description: `You have joined ${workspace.name}.`,
+          type: ToastNotificationType.Success
+        })
+        mutateActiveWorkspaceSlug(workspace.slug)
+        navigateTo(workspaceRoute(workspace.slug))
```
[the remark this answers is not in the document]
Mikehrn: Yeah right now you need both, one to update the active workspace and one to navigate there, I'm working on a refactor that will simplify this a lot, but didnt to open the PR yet

packages/frontend-2/components/workspace/discoverableWorkspaces/Modal.vue
```diff
@@ -16,6 +16,9 @@
         :request-status="workspace.requestStatus"
         show-dismiss-button
         location="workspace_switcher"
+        @auto-joined="workspace.requestStatus = 'approved'"
```
Mikehrn, line 19: Non-blocking, but in the follow up maybe use the WorkspaceJoinRequestStatus enum values instead of a string

andrewwallacespeckle: This is worth doing, I'll add it now

Mikehrn (dismissed):

PR #4802, opened 2025-05-22
feat(viewer): include selected object application ids in viewer state

cdriesler:
## Description & motivation
- Similar to #4797 we'd like comments to maintain a record their associated objects' application ids. This will allow connectors to interact with native geometry when working with comments, if the ids are present.

## Changes:

- Serialize `selectedObjectApplicationIds` in state
- Update state version and associated upgrade utilities

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,6 +22,18 @@ aliases:
       name: Install Dependencies
       command: YARN_ENABLE_HARDENED_MODE=0 PUPPETEER_SKIP_DOWNLOAD=true PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD=1 yarn
 
+  - &filters-allow-all
+    tags:
+      # run tests for any commit on any branch, including any tags
+      only: /.*/
+
+  - &filters-only-main-hotfix-testing
+    branches:
+      only: &branches-special
+        - main
+        - /^hotfix.*$/
+        - /^testing\d*$/
+
 workflows:
   test-build:
     jobs:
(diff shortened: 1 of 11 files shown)


gjedlicska:
i think we should do a more comprehensive change storing the selection ids in a `Record<string, string | null>` type, where the object's id is the key and the application id is the value. It would make it easier to correlate individual objects with their appicationId-s. 
I know this requires a bit more changes, but its prob worth it

gjedlicska (commented):
I think we should store `objectIds` and `applicationIds` in a record, rather than two separate arrays. It is impossible to correlate the object id and application id of an individual object this way.

packages/frontend-2/lib/viewer/composables/serialization.ts
```diff
@@ -95,7 +96,11 @@ export function useStateSerialization() {
         filters: {
           isolatedObjectIds: state.ui.filters.isolatedObjectIds.value,
           hiddenObjectIds: state.ui.filters.hiddenObjectIds.value,
-          selectedObjectIds: [...state.ui.filters.selectedObjectIds.value.values()],
+          selectedObjectIds: selectedObjects.value.map((obj) => obj.id),
+          selectedObjectApplicationIds: selectedObjects.value.reduce((ret, obj) => {
```
fabis94, line 100: as long as selectedObjectApplicationIds always includes all selectedObjectIds as keys, it should be OK to drop the old selectedObjectIds key. you can always handle breaking changes across serialized state versions in that initializeMissingData function

fabis94 (approved):

fabis94 (approved):

PR #4827, opened 2025-05-27
fix(fe): use loadedVersion for limits dialog and improve resource string handling

andrewwallacespeckle:
use loadedVersion for limits dialog and improve resource string handling


packages/frontend-2/components/viewer/PreSetupWrapper.vue
```diff
@@ -314,4 +320,12 @@ watch(
   },
   { immediate: true }
 )
+
+watch(
```
fabis94, line 324: this seems to duplicate the existing watcher 
image

fabis94 (approved):

PR #4851, opened 2025-05-30
feat(workspaces): add workspace level default seat property

cdriesler:
## Description & motivation

- We'd like to expose the ability to set the default seat type for new users in a workspace

## Changes:

- Add `defaultSeatType` column, which can be null
  - Null falls back to a viewer seat (for now)
- Emit non-null fallback from resolver in all cases


packages/server/modules/workspaces/tests/integration/workspaceSeat.spec.ts
```diff
@@ -112,6 +116,9 @@ describe('Workspace workspaceSeat services', () => {
     const sut = ensureValidWorkspaceRoleSeatFactory({
       createWorkspaceSeat,
       getWorkspaceUserSeat,
+      getWorkspaceDefaultSeatType: getWorkspaceDefaultSeatTypeFactory({
```
fabis94, line 119: 50% of the PR (and PRs generally) are the same factory DI setup adjustments

(this is just me venting again about how much I hate this manual DI setup)

fabis94 (approved):

PR #4853, opened 2025-05-30
fix(workspaces): use correct user id on `User.workspaces`

cdriesler:
use correct id when querying `User.workspaces` (currently surfaces requester in all cases)

Not usually user-visible because we almost always use `activeUser` these days, so it matches. But some custom gql queries behave silly.


fabis94 (changes requested):
this is sort of a valid fix, but also - you should not be using the only query that returns the User type for anyone but yourself (the active user), its deprecated

i don't think we should merge this, as a way to hopefully nudge API consumers  into not using the user() query

PR #4865, opened 2025-05-30
feat(gatekeeper): track seat updates

danielgak:
## Description & motivation

Improve mixpanel metrics around seat changes and user actions over them

## Changes:

- Add metrics with more granual metrics around Editor seats:
  - *Seat purchases*: when customers start paying for them + how many
  - *Seat downscales*: when customers stop paying for them + how many
  - *Seat editor addition*: what user was added as editor
  - *Seat editor removal*: what user was removed from editor


gjedlicska (changes requested):
The subscription data should not be leaked into the events, lets create an event payload, that contains data what we really care about in the app, which are the paid / editor seat numbers

PR #4868, opened 2025-06-02
fix(fe): Discoverable workspace card not dismissing

andrewwallacespeckle:
Add dismissed event to the card, and remove the worskpace from the local ref of workspaces.


packages/frontend-2/components/workspace/discoverableWorkspaces/Modal.vue
```diff
@@ -64,6 +65,10 @@ const dialogButtons = computed((): LayoutDialogButton[] => {
   ]
 })
 
+const onWorkspaceDismissed = (workspaceId: string) => {
+  localWorkspaces.value = localWorkspaces.value.filter((w) => w.id !== workspaceId)
```
Mikehrn, line 69: @andrewwallacespeckle Why did we have do this localWorkspaces again? Is it because we want certain workspace to still show in the UI, even though they are actually joined?

PR #4882, opened 2025-06-03
feat(gatekeeper): intoduce the enterprise plan

gjedlicska:
introduces a new workspace plan that will have more features than anyone else


fabis94 (dismissed):
there are some linting issues tho

PR #4884, opened 2025-06-04
fix: object history when quering by project

danielgak:
## Description & motivation

- when querying object through projects, we weren't showing the latest objects for every modal

## Changes:

- edited the query so it loads every latest commit for all the projects models


packages/server/modules/core/graph/resolvers/versions.ts
```diff
@@ -134,18 +134,23 @@ export = {
         project
       })
 
-      let lastVersion: Version | null
+      let latestVersion: Version | null = null
+      let latestVersions: Array<Version> | null = null
       if (getTypeFromPath(info) === 'Model') {
-        lastVersion = await ctx.loaders
+        latestVersion = await ctx.loaders
```
fabis94, line 140: this branching based on whether we're querying for the version through Model or not is no longer necessary

packages/server/modules/core/graph/resolvers/versions.ts
```diff
@@ -134,18 +134,23 @@ export = {
...
           .forRegion({ db: projectDB })
           .branches.getLatestCommit.load(parent.branchId)
       } else {
-        lastVersion = await ctx.loaders
+        latestVersions = await ctx.loaders
```
fabis94, line 144: i'm a bit concerned w/ having to pull all of the project's latest versions just to answer the question of whether it is the latest version.

it may be better to have a dataloader like: streams.checkIsLatestVersion({projectId, versionId}), that would batch up these requests and then check them all w/ a single query
or if its too complicated to do 1 query for different versions of different projects, we sometimes have dataloaders scoped to specific projects (check forStream())
that way we don't actually need to pull all of the versions in memory, we only get back a bool essentially

packages/server/modules/core/repositories/versions.ts
```diff
@@ -1,31 +1,26 @@
...
   }: {
     projectIds: readonly string[]
-  }): Promise<Record<string, Version & { projectId: string }>> => {
-    const results = await tables
-      .versions(db)
-      .join(StreamCommits.name, StreamCommits.col.commitId, Commits.col.id)
-      .whereIn(StreamCommits.col.streamId, projectIds)
-      .distinctOn(StreamCommits.col.streamId)
-      .select([...Commits.cols, knex.raw(`stream_commits."streamId" as "projectId"`)])
+  }): Promise<Record<string, Array<Version & { projectId: string }>>> => {
```
fabis94, line 12: `{ [something: string]: X}` is preferable to `Record<string, X>`, because it also explains the semantic meaning of the key

fabis94 (approved):

packages/server/modules/core/repositories/versions.ts
fabis94: also, its probably `Array | undefined` not `Array`. with this kind of type you're saying that any arbitrary string key will always have an array value

PR #4890, opened 2025-06-04
feat(automate): add profile:email scope to automation run tokens

gjedlicska:
c# account management needs access to the profile email to initialize an account


gjedlicska:
@JR-Morgan turns out we do not need this right?

PR #4899, opened 2025-06-06
Progress for Traversal and Rendering alongside Data loading

adamhathcock:
Also fixes the deferment logging

image
Screenshot 2025-06-06 133436
Screenshot 2025-06-06 135443


adamhathcock:
Traversal now looks like this 
Screenshot 2025-06-06 144328

packages/viewer/src/modules/tree/RenderTree.ts
```diff
@@ -31,11 +32,11 @@ export class RenderTree {
 
   public buildRenderTree(
     geometryConverter: GeometryConverter,
-    callback?: () => void
+    countCallback: (count: number) => void
```
AlexandruPopovici, line 35: I would make this optional, so that when no callback is wanted, none is provided instead of an empty one (like in the case of ObjLoader)

packages/viewer/src/modules/loaders/Speckle/SpeckleConverter.ts
AlexandruPopovici, line 1: I can't tell to well by looking at the diff what changed, but as long as there are no major changes in conversion behavior it should be all good

packages/viewer/src/modules/loaders/OBJ/ObjConverter.ts
```diff
@@ -56,7 +56,6 @@ export class ObjConverter {
     if (this.directNodeConverterExists(object)) {
       try {
         await this.convertToNode(object, childNode)
-        await callback()
```
AlexandruPopovici, line 59: Should this be removed?

AlexandruPopovici (approved):

PR #4905, opened 2025-06-09
feat: release to gha, disabled release on circle ci

danielgak:
## Changes

- Moved release pipelines to github actions

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,10 +29,12 @@ aliases:
 
   - &filters-only-main-hotfix-testing
     branches:
-      only: &branches-special
-        - main
-        - /^hotfix.*$/
-        - /^testing\d*$/
+      ignore: /.*/
+
+  - &branches-special
+    - main
+    - /^hotfix.*$/
+    - /^testing\d*$/
 
 workflows:
   test-build:


.github/workflows/builds.yml
```diff
@@ -6,60 +6,96 @@ on:
       IMAGE_VERSION_TAG:
         required: true
         type: string
+      DOCKERHUB_USERNAME:
+        required: true
+        type: string
+      PUSH_IMAGES:
+        required: true
+        type: boolean
```
danielgak, line 14: extracted push as imput, so the workflow trigger actually decides if we need to push or not. _(release pushes, pr no)_

danielgak:
Moved away some changes
https://github.com/specklesystems/speckle-server/pull/4914

.github/workflows/scripts/common.sh
```diff
@@ -4,17 +4,14 @@ set -eo pipefail
 # shellcheck disable=SC2034
 DOCKER_IMAGE_TAG="speckle/speckle-${SPECKLE_SERVER_PACKAGE}"
 
-# shellcheck disable=SC2034,SC2086
-IMAGE_VERSION_TAG="${IMAGE_VERSION_TAG:-${GITHUB_SHA}}"
-
```
danielgak, line 9: We do not need it in github actions, it only overrides it with wrong data when empty

.github/workflows/publish.yml
```diff
@@ -0,0 +1,95 @@
...
+        required: true
+      DOCKERHUB_TOKEN:
+        required: true
+      GH_DEVOPS_PAT:
+        required: true
+jobs:
+  helm-chart-oci:
+    runs-on: blacksmith
+    name: Helm chart oci
+    if: false
```
danielgak, line 26: disabled for now

PR #4908, opened 2025-06-10
feat: seat type selection in invite flows

fabis94:


packages/ui-components/src/components/form/select/Base.vue
```diff
@@ -469,6 +474,14 @@ const props = defineProps({
   tooltipText: {
     type: String,
     default: undefined
+  },
+  /**
+   * Optionally make the menu width wider/narrower than the button width by specifying the max width in pixels.
+   * Only supported when `mountMenuOnBody` is true.
+   */
+  menuMaxWidth: {
```
Mikehrn, line 482: Very nice to have this supported 😄

packages/server/modules/workspaces/repositories/users.ts
```diff
@@ -80,10 +91,35 @@ export const getInvitableCollaboratorsByProjectIdFactory =
     if (cursor) {
       query.andWhere(Users.col.createdAt, '<', cursor)
     }
-    return await query
+    const res = await query
```
fabis94, line 94: this query was busted and wasn't returning the data that the return type said it would

packages/server/modules/workspaces/services/invites.ts
```diff
@@ -166,7 +169,7 @@ export const collectAndValidateWorkspaceTargetsFactory =
     const targetWorkspaceSeatType =
       targetWorkspaceRole === Roles.Workspace.Admin
         ? WorkspaceSeatType.Editor
-        : WorkspaceSeatType.Viewer
+        : chosenWorkspaceSeatType
```
danielgak, line 172: Do we need to protect this from a Guest having an editor role?

danielgak (commented):
Generally server side looks okey,
my only question is around if Guests can be editors

packages/server/modules/workspaces/services/invites.ts
fabis94: guests can have editor seat types

Mikehrn (approved):
Nice, FE looks good 😄

benjaminvo (dismissed):
Tested, works well! I wasn't able to test it on a paid plan because I don't have billing set up locally. But, I'm happy to trust merge on that and test later.

packages/server/modules/workspaces/tests/integration/invites.graph.spec.ts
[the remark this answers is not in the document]
danielgak: but we do not check what seat type was assigned if not mistaken

PR #4930, opened 2025-06-11
feat(ci): improved caching around frontend2 builds

danielgak:
## Description & motivation

- Improved ci caching strategies for FE builds:
   - FE build on pull request consumes different caches than FE on deploy 
   - Sourcemaps stores a max cache
  
  
With this strategy: _(cache hit)_
https://github.com/specklesystems/speckle-server/actions/runs/15604028369

Without: _(both examples should be hits)_ 
(on main) https://github.com/specklesystems/speckle-server/actions/runs/15590454040/job/43908724547
(on pr) https://github.com/specklesystems/speckle-server/actions/runs/15590770112/job/43909051420

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -46,10 +46,22 @@ jobs:
         with:
           username: ${{ inputs.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - name: Build and push
+      - name: Build and push w/ version tag
         uses: useblacksmith/build-push-action@v1
+        if: ${{ inputs.PUSH_IMAGES }}
         with:
-          push: ${{ inputs.PUSH_IMAGES }}
+          push: true
+          tags: speckle/speckle-frontend-2:${{ inputs.IMAGE_VERSION_TAG }}
+          file: ./packages/frontend-2/Dockerfile
+          build-args: |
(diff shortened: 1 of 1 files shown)


.github/workflows/builds.yml
```diff
@@ -52,8 +52,6 @@ jobs:
           push: ${{ inputs.PUSH_IMAGES }}
           tags: speckle/speckle-frontend-2:${{ inputs.IMAGE_VERSION_TAG }}
           file: ./packages/frontend-2/Dockerfile
-          build-args: |
-            SPECKLE_SERVER_VERSION=${{ inputs.IMAGE_VERSION_TAG }}
```
danielgak, line 56: @fabis94 

On every run,this parameter is different and it invalidates most of the Dockerbuild on every single run. 

@iainsproat and me understand that we need it for building source maps (and we are keeping it there) but the question for you is:
- do we actually need this parameter in the FE build at buildtime if we can assure it will be available at runtime in the deployment?

fabis94: i dont remember much about how the build & sourcemap-build steps interact, but yes - you need env vars to be there on build, because that's the point in time when they get baked into the built code which changes the final file hashes which must link up exactly w/ the app running in prod

so the sourcemap app & the app built for production have to have the same exact set of env vars available, except the sourcemap one must also pass in the BUILD_SOURCEMAPS env var to actually enable sourcemaps to be built

PR #4944, opened 2025-06-16
feat(activity): add workspace events to activity

danielgak:
## Description & changes

Created new `activity` table that stores:
- `plan create`
- `plan update`
- `subscription update` 

to build a billing history around those items for a workspace. 
Every event carries the info of the user who caused it, but some as `plan update` or `subscription update` might happen without a user _(eg. paymentFailed from stripe)_.

A history will look something like this:

image

Minor tweaks in the event emission to make sure we do not emit duplicate events, or report duplicate info to mixpanel. 

Manual tests done:

- [x] on create workspace -> workspacePlan created
- [x] create a new subscription
- [x] update an existing one
- [x] schedule a subcription cancelation
- [x] cancel a subscription
- [x] renew a canceled subcription
- [x] renew a subscription scheduled from stripe
- [x] create subscription + change the card to a card that attaches but fails on payment + advance a billing interval
- [x] send same event from stripe twice (idempotence)
- [x] assign a new editor seat (purchase) to someone who joined
- [x] assign an editor seat that does exist 
- [x] invite someone as editor
- [x] add seats from stripe (edit subscription quantity)
- [x] unassign an editor seat and downscale the subcription
- [x] upgrade subscription having unassigned but paid editor seats


packages/server/modules/activitystream/domain/operations.ts
```diff
@@ -270,3 +274,10 @@ export type AddBranchDeletedActivity = (params: {
   userId: string
   branchName: string
 }) => Promise<void>
+
+export type SaveActivity = <
+  T extends keyof ResourceEventsToPayloadMap,
+  R extends keyof ResourceEventsToPayloadMap[T]
+>(
+  args: Omit<Activity<T, R>, 'createdAt' | 'id'>
+) => Promise<{ id: string }>
```
fabis94, line 283: Why not return the entire activity record?

packages/server/modules/activitystream/domain/types.ts
```diff
@@ -11,11 +11,68 @@ import {
...
+
+export type ResourceEventsToPayloadMap = {
+  workspace: {
+    workspace_plan_created: z.infer<typeof WorkspacePlanCreatedActivity>
+    workspace_plan_updated: z.infer<typeof WorkspacePlanUpdatedActivity>
+    workspace_subscription_updated: z.infer<typeof WorkspaceSubscriptionUpdatedActivity>
+  }
+}
+
+const workspacePlan = z.object({
```
fabis94, line 26: 1. The activitystream domain should not own the concept of a WorkspacePlan, that type/schema should be imported from the appropriate domain (workspaces? gatekeeper)
2. WorkspacePlan already has a type in @speckle/shared. I suppose this gives the extra benefit of having a Zod schema, but I think it should not be the concern of the acitivitystream to validate workspace plans.

Do we really need Zod here at all? Is it not enough to rely on TypeScript's type errors? It's not like any of this data is coming from user input, all of these are our own internal data structures that we can ensure are valid

packages/server/modules/activitystream/repositories/index.ts
```diff
@@ -270,3 +272,16 @@ export const saveActivityFactory =
...
+  ({ db }: { db: Knex }): SaveActivity =>
+  async (activity): Promise<{ id: string }> => {
+    const id = cryptoRandomString({ length: 10 })
+
+    await db(Activity.name).insert({
+      ...activity,
+      id
+    })
+
+    return { id }
```
fabis94, line 286: use .insert(item, '*') to get the full row back from the DB and return the full entity

fabis94 (changes requested):
looks good for the most part, ive some questions tho

packages/server/modules/activitystream/domain/types.ts
danielgak: I think Activity Stream wants to own the type of the snapshot that it's storing, I do not want to consume a type from workspace plan because I want to make clear that this the "todays workspace plan snapshot" and every time its changed, you need to specifically update the snapshot if needed.

Regarding if we need zod, I do not have a strong preference. I added it as @gjedlicska wanted to do some inferences and type exclusions with it, but that's more for exposing the activities part.

fabis94: As discussed in the call, I have concerns about the verbosity & DX difficulty of having duplicate types for all domain entities + for all of their subsequent versions, but that's something that can be also addressed later on

PR #2703, opened 2024-08-20
fix(workspaces): add email to existing account workflow bugfix

fabis94:

PR #4949, opened 2025-06-16
fix(fe): workspace members pagination not loading additional pages

andrewwallacespeckle:
I fixed the workspace members pagination by changing the Apollo cache merge policy from `merge: (_existing, incoming) => incoming` to `buildAbstractCollectionMergeFunction('WorkspaceCollaboratorCollection')`.

@fabis94 Do you remember why you had it replacing the data originally? It was done in [this branch](https://github.com/specklesystems/speckle-server/pull/2703). 

Are there any side effects of this I should worry about?


fabis94:
@andrewwallacespeckle because at that point in time Workspace.team was not a paginable collection, but just an array. Someone switched the return type later on and no one updated the merge policy 

image

fabis94 (approved):

PR #4961, opened 2025-06-19
fix(fe2): fixed model upload validation & bunch of other things

fabis94:
- Fixed new model still being created if FE upload validation fails, in which case the error message disappeared very quickly too. Now model only gets created after initial validation, and if the upload still fails on the serverside, it shows up as before through a toast notification (will be improved further [here](https://linear.app/speckle/issue/WEB-3610/highlight-failed-upload-job-on-a-model-card))
- Made it so that the upload progress UI always shows if there's an upload in progress, even if the model already has existing versions in it. Up until now, the upload progress bar would not show and we'd show the existing version's preview image.
- Fixed preview panoramic view being shown even if its just the blue cube placeholder image
msedge_T40SO7gNmU
- Fixed broken model card height (it not being a square) during importing.
image
- Fixed a bunch of spacing/height issues in model list view items
- Fixed double scrollbar on tables w/ limited height
image
- Fixed Model Uploads dialog not refetching query on every open
- Fixed model card actions menu being clipped in projects dashboard
image
- Fixed Model Upload table bottom loader taking up space even when there's no active loading
image


Mikehrn (approved):
Nice!

PR #4964, opened 2025-06-19
feat(fe): Allow disabling workspace creation for workspace members

andrewwallacespeckle:


packages/frontend-2/components/workspace/JoinPage.vue
```diff
@@ -64,6 +64,8 @@
         </FormButton>
         <FormButton
           v-if="!hasApprovedWorkspace"
+          v-tippy="!canClickCreate ? cantClickCreateReason : undefined"
```
Mikehrn, line 67: @andrewwallacespeckle @gjedlicska In a realistic usecase atm, will new users always see other workspace in the list? If their list is empty and this button disabled it might be a bad experience

packages/frontend-2/components/header/WorkspaceSwitcher/List.vue
```diff
@@ -33,7 +35,11 @@ defineProps<{
 
 const route = useRoute()
 const mixpanel = useMixpanel()
-const { isGuest } = useActiveUser()
+const { activeUser } = useActiveUser()
+
+const { canClickCreate, cantClickCreateReason } = useCanCreateWorkspace({
```
Mikehrn, line 40: We should add this same check to the middleware, else you can still access it directly

packages/frontend-2/components/projects/Add.vue
```diff
@@ -36,6 +36,7 @@ graphql(`
       }
     }
     ...UseCanCreatePersonalProject_User
+    ...UseCanCreateWorkspace_User
```
Mikehrn, line 39: Shouldn't this be on the activeUser, if we pass that into the composable?

packages/frontend-2/components/projects/add-dialog/Workspace.vue
```diff
@@ -6,20 +6,33 @@
...
 
 const emit = defineEmits<{
   'workspace-selected': [WorkspaceMoveProjectSelectWorkspace_WorkspaceFragment]
   canceled: []
 }>()
 
+const { activeUser } = useActiveUser()
+
+const { canClickCreate, cantClickCreateReason } = useCanCreateWorkspace({
+  activeUser: computed(() => activeUser.value)
```
Mikehrn, line 33: Not sure if we should add this to activeUser, we do this check only in two places (workspace switcher and join page), but we use activeUser in quite a lot of places so this will always be in that query, and we dont really do this with other permissions either. Also feels a bit weird to pass the active user into the composable, maybe just do a separate query for it?

packages/frontend-2/components/header/WorkspaceSwitcher/List.vue
andrewwallacespeckle: We don't want middleware here. If the user goes to the link, Benjamin said not to redirect them, but instead just disable the buttons and add a message saying why. This is done.

packages/frontend-2/components/projects/add-dialog/Workspace.vue
andrewwallacespeckle: New query added :)

packages/frontend-2/components/header/WorkspaceSwitcher/List.vue
Mikehrn: Cool :)

Mikehrn (approved):

PR #4967, opened 2025-06-20
feat(activity): added user info to checkout_subscription and subscription upgrade

danielgak:
## Description & motivation

As a topic from Workspace Billing Auditing recording plan changes or subscription changes must contain the user who caused it. 

Those events are emmited when stripe calls our webhook with the confirmation of a successful checkout, or subscription update, and under that rest billing endpoint, we can't trace it back to the user who triggered with the current data structures.

## Changes:

- Added userId to workpace_checkout_session
- Added updateIntent with userId to subscription for subscription target

With this changes it moves the logic in the BillingSubscriptionMutation resolver, so it no longer updates the workspace_plan, (nor the subscription billing interval) but rather it leaves this responsibility to the stripe callback that once the payment confirmation is received; we do make the update of the plan together with the desired subcription. 

Manual tests done:

- [x] create a new subscription
- [x] update an existing one
- [x] schedule a cancelation
- [x] cancel a current one
- [x] udpate a canceled subcription
- [x] after one year update
- [x] double events
- [x] add seats (purchase)
- [x] add seats from stripe
- [x] unassign,  and downscale seats
- [x] bump subscription with a card that will fail
- [x] upgrading w/ unassigned seats


packages/server/modules/gatekeeper/services/subscriptions.ts
```diff
@@ -79,52 +86,110 @@ export const handleSubscriptionUpdateFactory =
...
+            updateIntentProductId,
+            targetProductId,
+            workspaceId: subscription.workspaceId,
+            targetPlanName: planName,
+            planName: workspacePlan.name
+          },
+          'Fatal: Stripe product ID mismatch with subscription update intent'
+        )
       }
+    }
```
danielgak, line 140: if there is intent, we recognize it as the what we want to achieve; 

this handler recieves other updates apart from subscription bumps. In those cases there wont be any updateIntent.

it should never happen, but if productId given by stripe does not match the productId of the intent, we are in trouble.

PR #4991, opened 2025-06-26
fix(server): not firing upload created/processed subs in next gen uploads

fabis94:


packages/server/modules/fileuploads/domain/events.ts
```diff
@@ -10,10 +13,7 @@ export const FileuploadEvents = {
 
 export type FileuploadEvents = (typeof FileuploadEvents)[keyof typeof FileuploadEvents]
 
-type FileuploadStartedPayload = Pick<
-  FileUploadRecordV2,
-  'userId' | 'projectId' | 'fileSize' | 'fileType'
->
+type FileuploadStartedPayload = { upload: FileUploadRecordV2 & FileUploadRecord }
```
fabis94, line 16: @danielgak 

1. If the target entity is set as the entire event payload, then when you have to add new props you have to pollute the target entity type (e.g. `type CreatedPayload = FileUploadRecord & { branchCreatedAt: Date }`
2. If we're already retrieving the entire FileUpload entity, then preferably we should just emit the entire record instead of only a few of the props (for easier reuse)

fabis94: also - we have values for both v1 and v2 record available, so lets emit both for easier handling in listeners

packages/server/modules/fileuploads/helpers/types.ts
```diff
@@ -46,4 +46,4 @@ export type FileUploadRecordV2 = {
   metadata: Nullable<FileUploadRecordMetadata>
 }
 
-export type FileUploadGraphQLReturn = FileUploadRecord
+export type FileUploadGraphQLReturn = FileUploadRecord | FileUploadRecordV2
```
fabis94, line 49: another DX improvement - GQL returns can now accept either record type

packages/server/modules/fileuploads/services/management.ts
```diff
@@ -83,25 +50,10 @@ export const insertNewUploadAndNotifyFactoryV2 =
     queues: Pick<FileImportQueue, 'scheduleJob' | 'supportedFileTypes'>[]
     pushJobToFileImporter: PushJobToFileImporter
     saveUploadFile: SaveUploadFileV2
-    publish: PublishSubscription
     emit: EventBusEmit
   }): InsertNewUploadAndNotifyV2 =>
   async (upload) => {
     const file = await deps.saveUploadFile(upload)
-
```
fabis94, line 91: the bug was here - v2 version of function did not emit all of the necessary GQL subs

packages/server/modules/fileuploads/events/subscriptionListeners.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+const reportFileUploadUpdatedFactory =
+  (deps: { publish: PublishSubscription; getBranchById: GetBranchById }) =>
+  async (payload: EventPayload<typeof FileuploadEvents.Updated>) => {
+    const {
+      payload: { upload, isNewModel }
+    } = payload
+
+    if (isNewModel || !upload.modelId) {
```
fabis94, line 60: fixed similar issue I found for 'upload updated' events too

danielgak (approved):

PR #5004, opened 2025-06-27
feat(gatekeeper): add error log on seat mismatch

danielgak:
## Changes
- extracted stripe from reconciliation method as di
- created `stipe fake` and added basic tests to reconciliation logic 
- added specific error logs at downscale


danielgak:
checked with @fabis94 we are log it in the downscale moment

packages/server/modules/gatekeeper/services/subscriptions/manageSubscriptionDownscale.ts
```diff
@@ -110,9 +111,15 @@ export const manageSubscriptionDownscaleFactory =
           log.info('Did not need to downscale the workspace subscription')
         }
       } catch (err) {
-        log.error({ err }, 'Failed to downscale workspace subscription')
+        log.error(
+          {
+            err,
+            workspaceId: workspaceSubscription.workspaceId
+          },
+          'Failed to downscale workspace subscription'
```
danielgak, line 119: @iainsproat I think we should get notified by this error. Any subscription weirdness should manifest at this moment.

PR #5013, opened 2025-07-01
feat(tokens): create embed-specific tokens

cdriesler:
## Description & motivation

- Creates `embed_api_tokens` table and exposes mutation for creating them. Behaves similarly to PAT, but intended for private embeds.
- Adds policies for new operations
- Adds tests for token behavior


packages/server/modules/core/domain/tokens/operations.ts
```diff
@@ -86,4 +96,15 @@ export type CreateAndStorePersonalAccessToken = (
...
 ) => Promise<string>
 
+export type CreateAndStoreEmbedToken = (args: {
+  projectId: string
+  userId: string
+  /**
+   * The models (and optional versions) included in the embed.
+   * @example 'foo123,bar456@baz789'
+   */
+  modelIds: string
```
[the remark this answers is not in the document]
cdriesler: This is meant for surfacing/reconstructing the embed url when listed. It doesn't affect scopes but is useful for showing which project/model the embed is for.

packages/server/assets/core/typedefs/apitoken.graphql
```diff
@@ -55,6 +55,28 @@ input AppTokenCreateInput {
...
+  lifespan: BigInt!
+  lastUsed: DateTime!
+}
+
+input EmbedTokenCreateInput {
+  projectId: String!
+  """
+  The model(s) and version(s) string used in the embed url
+  """
+  modelIds: String!
```
fabis94, line 76: Is this the resourceIdString? The one formatted like 'modelId,modelId@versionId,objectId'. If it is, we should use this terminology everywhere and we even have utils for serializing & deserializing

packages/server/assets/core/typedefs/apitoken.graphql
```diff
@@ -77,3 +99,12 @@ extend type Mutation {
     @hasServerRole(role: SERVER_USER)
     @hasScope(scope: "tokens:write")
 }
+
+extend type ProjectMutations {
+  createEmbedToken(token: EmbedTokenCreateInput!): String!
```
fabis94, line 104: do these need any scope checks?

packages/server/modules/core/graph/resolvers/embedTokens.ts
```diff
@@ -0,0 +1,75 @@
...
+  createEmbedTokenFactory,
+  createTokenFactory
+} from '@/modules/core/services/tokens'
+import { mapAuthToServerError } from '@/modules/shared/helpers/errorHelper'
+import { removeNullOrUndefinedKeys } from '@speckle/shared'
+
+export = {
+  Project: {
+    embedTokens: async (parent, _args, context) => {
+      const canReadEmbedTokens =
```
fabis94, line 21: all of these should check resourceAccessRules to ensure the user even is allowed to work w/ embed tokens for this particular project

packages/server/modules/core/migrations/20250630073647_add_embed_tokens.ts
```diff
@@ -0,0 +1,30 @@
...
+      .references('id')
+      .inTable('streams')
+      .onDelete('cascade')
+    table
+      .string('userId')
+      .notNullable()
+      .references('id')
+      .inTable('users')
+      .onDelete('cascade')
+    table.string('modelIds').notNullable()
```
fabis94, line 23: why not use a string array column for this? better DX

packages/server/assets/core/typedefs/apitoken.graphql
fabis94: also: it would be nicer DX for the consumer if this would actually return the newly created EmbedToken entity. it probably returns the actual secret token string, which is also useful, but then you could return both

```
type CreateEmbedTokenResult {
  token: String!
  tokenMetadata: EmbedToken!
}
```

packages/server/assets/core/typedefs/apitoken.graphql
```diff
@@ -55,6 +55,28 @@ input AppTokenCreateInput {
...
+A token used to enable an embedded viewer for a private project
+"""
+type EmbedToken {
+  tokenId: String!
+  projectId: String!
+  userId: String!
+  modelIds: String!
+  createdAt: DateTime!
+  lifespan: BigInt!
+  lastUsed: DateTime!
```
fabis94, line 68: I'm wondering if the token itself shouldnt be a prop here, or at the very least a partially obfuscated version. Cause we'll need to list these in the FE at some point.

Maybe if its the real secret token, we add checks that only allow showing it in a non obfuscated version to project owners, and everyone else sees asterisks?

packages/server/modules/core/graph/resolvers/embedTokens.ts
```diff
@@ -0,0 +1,75 @@
...
+        await context.authPolicies.project.tokens.canReadEmbedTokens({
+          userId: context.userId,
+          projectId: parent.id
+        })
+      if (!canReadEmbedTokens.isOk) {
+        throw mapAuthToServerError(canReadEmbedTokens.error)
```
fabis94, line 27: throwIfAuthNotOk quicker

packages/server/modules/core/graph/resolvers/embedTokens.ts
```diff
@@ -0,0 +1,75 @@
...
+        throw mapAuthToServerError(canRevokeEmbedToken.error)
+      }
+
+      return await revokeEmbedTokenByIdFactory({ db })({
+        tokenId: args.token,
+        projectId: args.projectId
+      })
+    }
+  }
+} as Resolvers
```
fabis94, line 75: `export {} as Resolvers` has that issue you surfaced a while back where it allows minor type mismatches, just cause you're efffectively force casting the object as `Resolvers`

to get rid of the force cast do
```
const resolvers: Resolvers = {}
export default resolvers
```

packages/server/modules/core/repositories/tokens.ts
```diff
@@ -74,6 +80,30 @@ export const storePersonalApiTokenFactory =
     return newToken
   }
 
+export const storeEmbedApiTokenFactory =
+  (deps: { db: Knex }): StoreEmbedApiToken =>
+  async (token) => {
+    const [newToken] = await tables.embedApiTokens(deps.db).insert(token).returning('*')
```
fabis94, line 86: shouldnt this be in its own repo considering this is all for a new/separate table? our repos/resolvers/services tend to get too big

packages/server/modules/core/repositories/tokens.ts
```diff
@@ -143,6 +173,18 @@ export const revokeUserTokenByIdFactory =
...
 
+export const revokeEmbedTokenByIdFactory =
+  (deps: { db: Knex }): RevokeEmbedTokenById =>
+  async ({ tokenId: token, projectId }) => {
+    const tokenId = token.slice(0, 10)
+    const delCount = await tables
+      .embedApiTokens(deps.db)
+      .where({ tokenId, projectId })
+      .delete()
+    if (delCount === 0) throw new UserInputError('Did not revoke embed token')
```
fabis94, line 184: Shouldnt it say Not Found? Vague message

packages/server/modules/core/tests/embedTokens.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    await createTestWorkspace(workspace, adminUser)
+
+    projectA.workspaceId = workspace.id
+    projectB.workspaceId = workspace.id
+
+    await createTestStream(projectA, adminUser)
+    await createTestStream(projectB, adminUser)
+
+    const adminApollo = await testApolloServer({
+      context: await createTestContext({
```
fabis94, line 67: its not necessary to manually build the entire context, you can just pass in `authUserId` which will authenticate the context w/ that user id

packages/shared/src/authz/policies/project/tokens/canCreateEmbedToken.spec.ts
```diff
@@ -0,0 +1,133 @@
...
+import { canCreateEmbedTokenPolicy } from './canCreateEmbedToken.js'
+import { assert, describe, expect, it } from 'vitest'
+import {
+  ProjectNotEnoughPermissionsError,
+  ServerNoAccessError,
+  WorkspacePlanNoFeatureAccessError
+} from '../../../domain/authErrors.js'
+
+const buildCanCreateEmbedToken = (
+  overrides?: Partial<Parameters<typeof canCreateEmbedTokenPolicy>[0]>
```
fabis94, line 14: these should test both workspaced non-workspaced contexts

packages/shared/src/authz/policies/project/tokens/canCreateEmbedToken.ts
```diff
@@ -0,0 +1,101 @@
...
+      return ok()
+    }
+
+    const project = await loaders.getProject({ projectId })
+    if (!project?.workspaceId) {
+      // Project is not in a workspace, no plan to check
+      return ok()
+    }
+
+    const plan = await loaders.getWorkspacePlan({ workspaceId: project.workspaceId })
```
fabis94, line 84: If there is a workspace attached, shouldnt we check ensureProjectWorkspaceAccessFragment()? Or better even ensureImplicitProjectMemberWithWriteAccessFragment() w/ Role.Stream.Owner specified, then you can also remove the other is logged in, is project owner checks

packages/shared/src/authz/policies/project/tokens/canReadEmbedTokens.spec.ts
```diff
@@ -0,0 +1,63 @@
+import cryptoRandomString from 'crypto-random-string'
+import { canReadEmbedTokensPolicy } from './canReadEmbedTokens.js'
+import { Roles } from '../../../../core/constants.js'
+import { assert, describe, expect, it } from 'vitest'
+import { getProjectFake, getWorkspaceFake } from '../../../../tests/fakes.js'
+import { parseFeatureFlags } from '../../../../environment/index.js'
+import { ProjectNoAccessError } from '../../../domain/authErrors.js'
+
+const buildCanReadEmbedTokens = (
+  overrides?: Partial<Parameters<typeof canReadEmbedTokensPolicy>[0]>
```
fabis94, line 10: `OverridesOf<typeof canReadEmbedTokensPolicy>`

packages/shared/src/authz/policies/project/tokens/canRevokeEmbedToken.ts
```diff
@@ -0,0 +1,101 @@
...
+>
+
+export const canRevokeEmbedTokenPolicy: AuthPolicy<
+  PolicyLoaderKeys,
+  PolicyArgs,
+  PolicyErrors
+> =
+  (loaders) =>
+  async ({ userId, projectId }) => {
+    // Ensure logged in
```
fabis94, line 54: same comments as i left in the other one. these seem duplicate too, so maybe just invoke one from the other?

fabis94 (changes requested):
seems OK for the most part, I did have some questions tho

packages/server/assets/core/typedefs/apitoken.graphql
cdriesler: yes, didn't realize we had a name for it. Fixed all references

packages/server/assets/core/typedefs/apitoken.graphql
cdriesler: I don't think scope checks are necessary given the policy, but I dig the return type. Done

packages/server/modules/core/migrations/20250630073647_add_embed_tokens.ts
cdriesler: +1 for next time when it is a string array, but encoded `resourceIdString` as string here

packages/server/modules/core/graph/resolvers/embedTokens.ts
cdriesler: This one I don't follow, I only see resource access support for workspace and projects. Is this a check that the token allows access to the project, if resource limit exist?

packages/server/assets/core/typedefs/apitoken.graphql
cdriesler: Discussed with Mike and will hand-wave that this is enough (with inclusion of your `LimitedUser` change)

packages/server/modules/core/tests/embedTokens.spec.ts
cdriesler: This is meant to use the token explicitly and this was how I managed to get it to consider scope/resource access limits. Is there a better way to set this up if I want to emulate a client with a token I have?

packages/shared/src/authz/policies/project/tokens/canCreateEmbedToken.spec.ts
cdriesler: imo the relevant workspace/non-workspace cases are handled in the overrides in the tests, can be tweaked if you don't feel so

packages/shared/src/authz/policies/project/tokens/canCreateEmbedToken.ts
cdriesler: much nicer, done

packages/shared/src/authz/policies/project/tokens/canReadEmbedTokens.spec.ts
cdriesler: love it

packages/shared/src/authz/policies/project/tokens/canRevokeEmbedToken.ts
cdriesler: done, unified

packages/shared/src/authz/policies/project/tokens/canReadEmbedTokens.ts
```diff
@@ -0,0 +1,54 @@
...
+    | typeof ProjectNoAccessError
+    | typeof WorkspaceNoAccessError
+    | typeof WorkspaceSsoSessionNoAccessError
+    | typeof WorkspaceNotEnoughPermissionsError
+    | typeof ProjectNotEnoughPermissionsError
+  >
+> =
+  (loaders) =>
+  async ({ userId, projectId }) => {
+    const hasReadAccess = await ensureImplicitProjectMemberWithReadAccessFragment(
```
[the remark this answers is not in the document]
cdriesler: spoke w/ mike, this was a misunderstanding of how listing will work and is not true. Only being who can update tokens can list them

packages/server/assets/core/typedefs/apitoken.graphql
```diff
@@ -55,6 +55,28 @@ input AppTokenCreateInput {
   limitResources: [TokenResourceIdentifierInput!]
 }
 
+"""
+A token used to enable an embedded viewer for a private project
+"""
+type EmbedToken {
+  tokenId: String!
```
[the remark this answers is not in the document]
cdriesler: dig it, done

packages/server/modules/core/tests/embedTokens.spec.ts
fabis94: authUserId does set default scopes/roles, that's true, so if you want to customize those then yea you have to create a manual test context

packages/server/assets/core/typedefs/apitoken.graphql
fabis94: scope checks have a different point than auth policies - auth policies check what the user can do, scopes check what the token can do. you, as a user may be a llowed to manage embed tokens, but you still might not want to allow your token to do so

fabis94 (approved):

PR #5016, opened 2025-07-01
feat(fe2): improved file import error handling

fabis94:
Main changes:
* Show file upload/import errors in an error dialog. Upload cards no longer have an error state.
* If a model's last upload failed, it says so on the model card. Clicking on the label uploads the historical model uploads list dialog.


packages/frontend-2/components/project/PendingFileImportStatus.vue
```diff
@@ -10,8 +10,8 @@
         ].includes(upload.convertedStatus)
       "
     >
-      <span class="text-body-xs mb-1">
-        {{ isSelfImport ? 'Importing' : 'Uploading new version' }}
+      <span class="text-body-xs mb-1 text-center">
+        {{ isSelfImport ? 'Processing upload' : 'Processing uploaded version' }}
```
fabis94, line 14: improved copy to better imply that the user does not have to sit there waiting for the "upload" to finish (cc @iainsproat )

packages/frontend-2/components/singleton/FileUploadErrorDialog.vue
```diff
@@ -0,0 +1,124 @@
+<template>
```
fabis94, line 1: image

Re-usable error reference component also used here

packages/frontend-2/components/singleton/Managers.vue
```diff
@@ -3,6 +3,7 @@
     <ClientOnly>
       <SingletonToastManager />
     </ClientOnly>
+    <SingletonFileUploadErrorDialog />
```
fabis94, line 6: Singleton dialog that shows all failed file uploads

packages/frontend-2/lib/core/composables/error.ts
```diff
@@ -83,3 +86,58 @@ export const useLogToLoggingTransports = () => {
...
+   */
+  date: Date | dayjs.Dayjs
+  /**
+   * Optionally add extra payload to the logger.error() call
+   */
+  extraPayload?: Record<string, unknown>
+}
+
+export const useGenerateErrorReference = () => {
+  const logger = useLogger()
```
fabis94, line 102: Error reference looks like this:

```
Reference: #sQKUmodtvaTXSce0gG59C | #2a774d63-de45-4119-8310-5e5a890485ee | 2025-07-01T11:11:57.700Z | URL: http://localhost:8081/aa | ID: fe-error-w4fqH6EMB3Lhi-jF2fSSw
```

packages/frontend-2/lib/core/helpers/apolloSetup.ts
```diff
@@ -67,37 +68,58 @@ export function buildAbstractCollectionMergeFunction<T extends string>(
...
+    if (has(args, 'input.cursor')) return get(args, 'input.cursor') as string | null
+    if (has(args, 'filter.cursor')) return get(args, 'filter.cursor') as string | null
+    return undefined
+  }
+
+  const merge: FieldMergeFunction<
+    Optional<AbstractCollection<T>>,
+    AbstractCollection<T>
+  > = (existing, incoming, options) => {
+    // If incoming data was requested w/ cursor: null, its the very first page and should replace whatever existed before
```
fabis94, line 86: We used to always append incoming items to existing items for pagination, but there are situations where we actually want the incoming list to fully replace the existing list. This happens when the updated data comes back from a subscription - its note a fetchMore call, but a full replacement for the list.

So I added logic that essentially replaces instead of appending, if the GQL field's args have cursor explicitly set to null, meaning that it's supposed to be the very first page

packages/server/assets/fileuploads/typedefs/fileuploads.graphql
```diff
@@ -95,6 +95,11 @@ type FileUpload {
   Model associated with the file upload, if it exists already
   """
   model: Model
+  modelId: String
+  """
+  Date when upload was last updated
+  """
+  updatedAt: DateTime!
```
fabis94, line 102: confusing that there's both lastConvertedAt and lastUploadedAt - as a consumer I virtually always just want the latest updatedAt

packages/ui-components/src/components/layout/Dialog.vue
```diff
@@ -86,7 +86,12 @@
               >
                 <XMarkIcon class="h-6 w-6" />
               </FormButton>
-              <div ref="slotContainer" :class="slotContainerClasses" @scroll="onScroll">
+              <div
+                ref="slotContainer"
+                v-memo="isClosing ? [memoKey] : [(memoKey = Math.random())]"
```
fabis94, line 91: Made it so that when the dialog is starting to close and the fade-out transition starts playing, the dialog's contents stop being reactive and show the same DOM that was shown when closing was initiated.

This resolves the issue when you have a single value from which you derive both the dialog's state and also whether it's open or not. It used to be that in such cases when you'd close the dialog, the data inside the dialog would disappear too and you could see it in the closing transition.

fabis94:
```
/**
 * This allows you to only mutate modelToEdit, and editDialogOpen will update automatically depending on whether
 * you've chosen a model to edit.
 *
 * But up until now, closing editDialogOpen would also wipe modelToEdit and make the dialog show up as empty for the split second while its closing
 */
const modelToEdit = ref(null)
const editDialogOpen = computed({
  get: () => !!modelToEdit.value,
  set: (newVal) => {
    if (!newVal) {
      modelToEdit.value = null
    }
  }
})

```

packages/frontend-2/components/project/page/models/Card.vue
```diff
@@ -178,10 +210,28 @@ const importArea = ref(
...
 const hovered = ref(false)
 
+const showLastUploadFailed = computed(() => {
+  if (isPendingModelFragment(props.model)) return false
+  const lastUpload = props.model.lastUpload?.items[0]
+  const lastVersion = props.model.lastVersion?.items[0]
+
+  // Only show if last upload failed & there is no last version,
+  // or last version is older than last upload
+  if (lastUpload?.convertedStatus !== FileUploadConvertedStatus.Error) return false
```
fabis94, line 230: If there was a stuck pending upload, we used to show it even if there were newer successfully created versions. This fixes that - if you create a new version after the pending upload, its gonna show that version, not the pending (stuck) upload.

Mikehrn (approved):

PR #5041, opened 2025-07-08
feat(backgroundjobs): add new background jobs module for file imports

gjedlicska:
# Things this PR is meant to tackle
With bull not being adequately cross platform, we're moving to using postgres as a job queuing system. I'm introducing a new module for this reason and if this setup yields good results, we can roll it out to all background jobs.

- [x] adds new background jobs module for postgres based queueing
- [x] uses background jobs for fileimports driven by a featureflag
- [ ] handle job timeouts

# Not in scope for this PR
- [ ] publish / notify handlers for job status
- [ ] obfuscate token in payload


packages/server/modules/backgroundjobs/domain.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from 'zod'
+export const BackgroundJobStatus = {
+  Queued: 'queued',
+  Processing: 'processing',
```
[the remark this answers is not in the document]
gjedlicska: forgot to push that, there is no processing state. Workers must adhere to timeouts and release the lock. If they loose connection or just stop, the lock immediately goes away and the job becomes available for workers again

packages/server/modules/backgroundjobs/domain.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from 'zod'
+export const BackgroundJobStatus = {
+  Queued: 'queued',
+  // Processing: 'processing',
```
fabis94, line 4: i'm confused why is there no processing status? what happens to jobs while they're processing? surely they don't just disappear from the table

packages/server/modules/backgroundjobs/repositories.ts
```diff
@@ -0,0 +1,48 @@
...
+export const storeBackgroundJobFactory =
+  ({
+    db,
+    originServerUrl
+  }: {
+    db: Knex
+    originServerUrl: string
+  }): StoreBackgroundJob =>
+  async ({ job }) => {
+    await tables.backgroundJobs(db).insert({ ...job, originServerUrl })
```
fabis94, line 40: why not return it too w/ .insert(x, '*')?

packages/server/modules/backgroundjobs/tests/unit/services.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+          throw new Error(errorMessage)
+        }
+
+        const scheduleBackgroundJob = scheduleBackgroundJobFactory({
+          storeBackgroundJob: failingStore,
+          jobConfig: mockJobConfig
+        })
+
+        try {
+          await scheduleBackgroundJob({ jobPayload: mockJobPayload })
```
fabis94, line 147: expectToThrow()

fabis94 (approved):
only seems to be the base foundation here and all of that looks good enough, altho im confused about how this generally is gonna work especially w/o a processing state.

maybe im misunderstanding it, but seems like there's no way to get in-progress jobs and there's also the issue of the worker having to reinsert the job on failure (which - if the worker is failing/being closed down or whatever, can also fail)

packages/server/modules/backgroundjobs/domain.ts
fabis94: @iainsproat can you query for all that are locked tho? i still see value in a clearly visible processing status

PR #5058, opened 2025-07-10
Feat: Improved workspace switcher

Mikehrn:


packages/frontend-2/components/dashboard/Sidebar.vue
```diff
@@ -183,12 +216,6 @@ const projectsLink = computed(() => {
     : projectsRoute
 })
 
-const showSidebar = computed(() => {
```
Mikehrn, line 186: Everybody can now see the sidebar, as long as you are logged in, just hiding the "projects" link if you are not part of the active workspace

packages/frontend-2/middleware/004-onboarding.global.ts
```diff
@@ -137,86 +120,4 @@ export default defineNuxtRouteMiddleware(async (to) => {
       return navigateTo(workspaceCreateRoute)
     }
   }
-
-  // 4. Redirect to the correct workspace
```
Mikehrn, line 141: All of these checks became redundant

packages/frontend-2/components/dashboard/Sidebar.vue
```diff
@@ -162,19 +172,45 @@ import {
 } from '~/lib/common/helpers/route'
 import { useRoute } from 'vue-router'
 import { useActiveUser } from '~~/lib/auth/composables/activeUser'
-import { useNavigation } from '~~/lib/navigation/composables/navigation'
 import { useMixpanel } from '~~/lib/core/composables/mp'
+import { useActiveWorkspace } from '~/lib/user/composables/activeWorkspace'
+import { graphql } from '~/lib/common/generated/gql'
+import { useQuery } from '@vue/apollo-composable'
+
+const dashboardSidebarQuery = graphql(`
```
fabis94, line 180: isn't this more or less the same query thats already invoked in `useActiveWorkspace()`? you could just make useActiveWorkspace() return the id and role as well

packages/frontend-2/components/header/WorkspaceSwitcher/header/Header.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <div class="p-2 pb-3 flex flex-col gap-y-4">
-    <div class="flex gap-x-2 items-center">
+    <div class="flex gap-x-3 items-center">
       <MenuItem>
-        <NuxtLink :to="to">
+        <NuxtLink :to="to || ''">
```
fabis94, line 5: why? isnt it better to not have a `href` at all than a `href` that points to an invalid URL

packages/frontend-2/lib/core/configs/apollo.ts
```diff
@@ -105,6 +105,10 @@ function createCache(): InMemoryCache {
           automateFunctions: {
             keyArgs: ['filter', 'limit'],
             merge: buildAbstractCollectionMergeFunction('AutomateFunctionCollection')
+          },
+          workspaceBySlug: {
+            keyArgs: ['slug'],
+            merge: mergeAsObjectsFunction
```
fabis94, line 111: we didn't usually need a custom merge function for non-array/collection fields, is this really needed?

packages/frontend-2/lib/user/composables/activeWorkspace.ts
```diff
@@ -0,0 +1,50 @@
...
+        'activeWorkspace',
+        ({ helpers: { ref } }) => {
+          // Handle case where workspace is set to null (personal projects)
+          return newWorkspace ? ref('LimitedWorkspace', newWorkspace.id) : null
+        }
+      )
+    }
+  })
+
+  const setActiveWorkspace = async (options: {
```
fabis94, line 39: to ensure either id or slug are passed in, you can type this as `options: {id: string } | { slug: string }`

packages/frontend-2/lib/user/graphql/queries.ts
```diff
@@ -21,3 +21,13 @@ export const userEmailsQuery = graphql(`
     }
   }
 `)
+
+export const userActiveWorkspaceQuery = graphql(`
+  query UserActiveWorkspace {
+    activeUser {
```
fabis94, line 27: for both activeUser and activeWorkspace https://www.notion.so/speckle/General-Web-Code-Review-Guide-1f2bf3dfea634460b3778ac3cf4aa1bb?source=copy_link#2851514fbf9a406da31e430a57859565

packages/frontend-2/lib/user/graphql/queries.ts
```diff
@@ -21,3 +21,13 @@ export const userEmailsQuery = graphql(`
     }
   }
 `)
+
+export const userActiveWorkspaceQuery = graphql(`
```
fabis94, line 25: also - what's this for? im asking cause i saw multiple queries already asking for similar data

fabis94 (changes requested):

packages/frontend-2/components/dashboard/Sidebar.vue
Mikehrn: There will be some other information we need aswell in the sidebar, such as plan name and status to conditionally render some items, however that is not in this branch yet. It felt a bit overkill to always fetch that info

packages/frontend-2/lib/user/graphql/queries.ts
Mikehrn: In some places we need just the active workspace slug in order to fetch additional workspace info using a workspaceBySlug query for example, I've renamed it to useActiveWorkspaceSlug to make it a bit clearer that this is meant to just get the slug, not the full activeWorkspace

fabis94 (approved):

fabis94 (dismissed):

andrewwallacespeckle (approved):

PR #5062, opened 2025-07-11
refactor(fe): Remove LegacyViewer

andrewwallacespeckle:
- Front-end now talks exclusively to `CameraController`
   - replaced all `instance.zoom` / `viewer.setCameraView` calls with `useCameraUtilities().zoom`
   - added runtime overload guard in `ui.ts`
   - removed frame-based camera transform updates in favor of stationary events
- Camera bidirectional sync simplified
   - merged separate position/target watchers into a single one to avoid duplicate `setCameraView` calls.
- Selection now managed entirely through `SelectionExtension`
   - removed legacy `filters.selectedObjects` state
   - unified selection tracking under `useSelectionUtilities()`
- Registered missing `ViewModes` extension during viewer setup
- Updated `Row.vue` funnel-button logic to use new zoom helper
- `anchorPoints.ts` – added null-guards for viewer query results
- Removed last references to `PolarView`, `viewer.getCurrentSectionBox`
- Renamed GQL redirect query
  - `LegacyViewerCommitRedirectMetadata` → `ViewerCommitRedirectMetadata`
- Misc clean-ups: stray empty file deleted, linter issues fixed


packages/frontend-2/middleware/002-redirects.global.ts
```diff
@@ -22,8 +22,8 @@ const legacyBranchMetadataQuery = graphql(`
   }
 `)
 
-const legacyViewerCommitMetadataQuery = graphql(`
```
fabis94, line 25: these have nothing to do with the LegacyViewer so they shouldnt be touched, they refer to the FE1 (legacy) GQL API that's used by the viewer

fabis94 (changes requested):
have some performance questions

packages/frontend-2/lib/viewer/composables/setup.ts
```diff
@@ -335,11 +335,21 @@ function createViewerDataBuilder(params: { viewerDebug: boolean }) {
...
+    viewer.createExtension(CameraController)
+    viewer.createExtension(SectionTool)
+    viewer.createExtension(SectionOutlines)
+    viewer.createExtension(FilteringExtension)
+    viewer.createExtension(ExplodeExtension)
+    viewer.createExtension(DiffExtension)
+    viewer.createExtension(MeasurementsExtension)
+    viewer.createExtension(SelectionExtension)
+    viewer.createExtension(ViewModes)
+
```
AlexandruPopovici, line 352: You would also need either:
- A separate [HighlightExtension](https://github.com/specklesystems/speckle-server/blob/cad77b92c9f0ec7ccc7d16140125769a439eb256/packages/viewer/src/modules/LegacyViewer.ts#L63) like the `LegacyViewer` used
- Handle highlighting yourself via [`setMaterial`](https://docs.speckle.systems/developers/viewer/speckle-renderer-api#setmaterial)

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -593,13 +602,15 @@ function useViewerFiltersIntegration() {
 
       if (hidable.length) {
         withWatchersDisabled(() => {
-          instance.hideObjects(hidable, stateKey, true)
+          instance.getExtension(FilteringExtension).hideObjects(hidable, stateKey, true)
```
AlexandruPopovici, line 605: Same thing as with isolating

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -257,59 +272,28 @@ export function useFilterUtilities(
...
-    selectedObjects.value = [...selectedObjects.value, object]
-  }
+  const { instance } = useInjectedViewer()
+  const selExt = instance.getExtension(SelectionExtension)
 
-  const removeFromSelection = (objectOrId: SpeckleObject | string) => {
-    const oid = isString(objectOrId) ? objectOrId : objectOrId.id
-    const idx = selectedObjects.value.findIndex((o) => o.id === oid)
-    if (idx === -1) return
+  const objects = shallowRef<SpeckleObject[]>(
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/5062/commits/8c6f5fcb16ae9278a7daeb2056b68dee3ed36736

packages/frontend-2/lib/viewer/composables/setup.ts
```diff
@@ -923,8 +936,15 @@ function setupInterfaceState(
   const loadProgress = ref(0)
 
   const isolatedObjectIds = ref([] as string[])
+  /**
+   * Currently selected objects in the viewer.
+   * Stored in viewer state so that every composable/component can access the
+   * same reactive reference without spawning duplicate refs or event
+   * listeners.  Populated & kept in sync inside useSelectionUtilities().
+   */
+  const selectedObjectIds = ref([] as string[])
```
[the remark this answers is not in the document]
fabis94: (see original selectedObjectIds implementation)

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -257,59 +271,23 @@ export function useFilterUtilities(
...
-    const idx = selectedObjects.value.findIndex((o) => o.id === oid)
-    if (idx === -1) return
+  const selExt = instance.getExtension(SelectionExtension)
 
-    const newObjects = selectedObjects.value.slice()
-    newObjects.splice(idx, 1)
-    selectedObjects.value = newObjects
-  }
+  const objects = selectedObjects
+  const objectIds = computed(() => selectedObjectIds.value)
```
fabis94, line 282: why not just use the selectedObjectIds ref/computed directly

packages/frontend-2/lib/viewer/composables/setup/selection.ts
```diff
@@ -22,9 +22,30 @@ function useCollectSelection() {
...
+  const state = useInjectedViewerState()
+  const selExt = state.viewer.instance.getExtension(SelectionExtension)
+
+  const update = () => {
+    const objs = selExt.getSelectedObjects() as SpeckleObject[]
+    state.ui.selectedObjects.value = objs
+    state.ui.selectedObjectIds.value = objs.map((o) => o.id as string)
+  }
+
+  update()
```
fabis94, line 35: should probably do this when viewer is ready, not the moment this composable is invoked (may be too early)

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -472,53 +478,48 @@ function useViewerCameraIntegration() {
...
+        },
+        true
+      )
+    },
+    { immediate: true, deep: false }
   )
 
+  // react to projection mode changes
   watch(
-    target,
```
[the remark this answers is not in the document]
andrewwallacespeckle: It's just been merged with the watcher above

packages/frontend-2/middleware/002-redirects.global.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/5062/commits/f2240ebe7cbe59edf82bbd29dd5a8bb09d21f60e

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -551,7 +552,11 @@ function useViewerFiltersIntegration() {
     (newVal, oldVal) => {
       if (arraysEqual(newVal, oldVal || [])) return
 
-      instance.highlightObjects(newVal)
+      if (!newVal.length) {
+        instance.getExtension(SelectionExtension).clearSelection()
+      } else {
+        instance.getExtension(SelectionExtension).selectObjects(newVal)
```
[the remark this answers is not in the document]
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/5062/commits/9ab93983ca3a10c4732e1cbd902e59b8d408ff07

packages/frontend-2/lib/viewer/composables/setup.ts
andrewwallacespeckle: I added HighlightExtension:
https://github.com/specklesystems/speckle-server/pull/5062/commits/9ab93983ca3a10c4732e1cbd902e59b8d408ff07

packages/frontend-2/lib/viewer/composables/setup/selection.ts
andrewwallacespeckle: https://github.com/specklesystems/speckle-server/pull/5062/commits/5e9092be89fbd2e4d07451a314752eb22d1ca6c9

packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -639,27 +651,6 @@ function useViewerFiltersIntegration() {
     },
     { initialOnly: true }
   )
-
-  watch(
```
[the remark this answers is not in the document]
andrewwallacespeckle: This watcher synced viewer selection with filters.selectedObjects.
selectedObjects was moved from the filters object to top-level ui state, and selection sync is now handled in useSelectionStateSync() in selection.ts.

packages/frontend-2/lib/viewer/composables/setup/selection.ts
```diff
@@ -31,6 +31,15 @@ function useSelectionStateSync() {
     state.ui.selectedObjects.value = objs
   }
 
+  watch(state.ui.selectedObjects, (newVal) => {
```
fabis94, line 34: make sure to test this doesnt cause an infinite loop: selectedObjects updates viewer updates selectedObjects updates viewer etc. etc.

in some cases for some sync we used to disable watchers while making the mutation, to avoid this.

packages/frontend-2/lib/viewer/composables/setup/selection.ts
```diff
@@ -22,9 +22,52 @@ function useCollectSelection() {
...
+function useSelectionStateSync() {
+  const state = useInjectedViewerState()
+  const selExt = state.viewer.instance.getExtension(SelectionExtension)
+
+  const update = () => {
+    const objs = selExt.getSelectedObjects() as SpeckleObject[]
+    state.ui.selectedObjects.value = objs
+  }
+
+  watch(state.ui.selectedObjects, (newVal) => {
```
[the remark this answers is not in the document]
AlexandruPopovici: @fabis94 
`SelectionExtrension` selects objects on it's own. Then the frontend listens on `ViewerEvent.ObjectClicked` which will get `state.ui.selectedObjects.value` written to. If that will trigger the handler inside the watch, then the frontend calls `      selExt.selectObjects(newIds)` which selects again. If that watch handler isn't triggered then it's fine.

The easiest thing to do is to run and debug it, and it be clear then how things unfold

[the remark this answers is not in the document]
fabis94: @AlexandruPopovici ok if i understand you correctly you're echoing this concern i had: https://github.com/specklesystems/speckle-server/pull/5062#discussion_r2224770347

we still need to be able to sync changes from selectedObjects.value -> viewer, cause the frontend selects objects by writing to selectedObjects.value, not by interacting w/ the viewer instance directly. so if what you're saying does in fact happen, we should instead disable watchers (we've done this in other sync mechanism) while the update is made or do something else to avoid this infinite loop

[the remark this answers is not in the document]
andrewwallacespeckle: I'm on another task, but i'll be back on this shortly and will confirm.

andrewwallacespeckle:
- Added watcher loop prevention using the same pattern as filters (`preventSelectionWatchers` flag)
- Fixed ID mismatch - scene explorer uses full URLs but viewer expects clean IDs, so I added normalisation using the `segments[4]` pattern used elsewhere
- Fixed broken scene explorer tree clicking
- Cleaner utilities update state, watcher handles viewer sync
- Everything works with proper bidirectional sync now. The ID normalisation works well, but wondering if there's a better upstream way to avoid full URLs in scene explorer?

PR #5065, opened 2025-07-14
chore(server): get rid of duplicate gql TS types

fabis94:
less TS types for the TS language server to keep track of and hopefully somewhat improved IDE performance


cdriesler (approved):
hot

danielgak (approved):

PR #5079, opened 2025-07-15
feat(file_upload): update observability metrics

danielgak:
# Changes
- updated observability metrics to also track postgres job counts from background_jobs
- added second metrics to file uplaods
- fixed counts on next gen metrics


packages/server/modules/fileuploads/observability/metrics.ts
```diff
@@ -46,7 +62,12 @@ export const initializeMetrics = (params: {
       requestQueues.forEach(async (requestQueue) => {
         this.set(
           { parser: requestQueue.label },
-          await requestQueue.queue.getWaitingCount()
+          'queue' in requestQueue
+            ? await requestQueue.queue.getWaitingCount()
+            : await requestQueue.getBackgroundJobCount({
+                status: 'queued',
+                jobType: 'fileImport'
+              })
```
danielgak, line 70: @iainsproat I'm doing the same query for both gauges and i'm not sure there is any difference between:
- speckle_server_file_import_jobs_request_queue_pending gauge
- speckle_server_file_import_jobs_request_queue_waiting gauge

in the postgres queue context

PR #5095, opened 2025-07-17
DONT MERGE: Saved Views proposed GQL spec

fabis94:


packages/server/assets/viewer/typedefs/savedViews.graphql
```diff
@@ -0,0 +1,203 @@
...
+  groupId: ID!
+}
+
+"""
+One or both of the position values must be specified
+"""
+input ChangeViewOrderInput {
+  id: ID!
+  projectId: ID!
+  previousViewPosition: Float
```
fabis94, line 188: gonna try something like this 

Image

packages/server/assets/viewer/typedefs/savedViews.graphql
```diff
@@ -0,0 +1,203 @@
...
+  id: ID!
+  name: String!
+  description: String
+  author: LimitedUser!
+  createdAt: DateTime!
+  updatedAt: DateTime!
+  """
+  Original resource ID string that this view is associated with.
+  """
+  resourceIdString: String!
```
fabis94, line 16: Maybe string array instead? Or both?

PR #5096, opened 2025-07-17
Cache By Reference count

adamhathcock:
adds in reference finding and tracking
scans for `referencedId`
if referencedId isn't cache then adds to the request queue
for eviction: if reference count is greater than zero, it's not evicted.


adamhathcock:
https://linear.app/speckle/issue/WEB-3844/cache-by-reference

AlexandruPopovici (approved):

PR #5112, opened 2025-07-18
fix: getting all user projects on deletion and user filter

danielgak:
## Changes
- Created new repo method for iterating over projects via userId or workspaceId _(filtering them)_
- This fixes iterating over all projects on user delete and hitting the timeout

### DIFF
--- a/packages/server/modules/core/repositories/streams.ts
+++ b/packages/server/modules/core/repositories/streams.ts
@@ -108,7 +108,8 @@ import {
   GetStreamsCollaborators,
   GetStreamsCollaboratorCounts,
   GetImplicitUserProjectsCountFactory,
-  GrantProjectPermissions
+  GrantProjectPermissions,
+  GetExplicitProjects
 } from '@/modules/core/domain/streams/operations'
 import { generateProjectName } from '@/modules/core/domain/projects/logic'
 import { WorkspaceAcl } from '@/modules/workspacesCore/helpers/db'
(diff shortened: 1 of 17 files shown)


packages/server/modules/core/repositories/streams.ts
```diff
@@ -1489,3 +1490,46 @@ export const getImplicitUserProjectsCountFactory =
     const [{ count }] = await q
     return parseInt(count)
   }
+
+/**
+ * Batch the explicit projects givent by the workspace, the user or both
+ */
+export const getExplicitProjects =
```
danielgak, line 1497: Not super sure about the naiming, got inspired by the repo methods above.

tried diferent things but legacy repo call does not work as it left joins users, and does not filter by acl and other user project methods, make the userId mandatory without allowing for querying via userId or workspaceId as desired.

packages/server/modules/workspaces/events/eventListener.ts
```diff
@@ -986,7 +989,9 @@ export const initializeEventListenersFactory =
         await withTransaction(
           async ({ db: trx }) => {
             const onWorkspaceRoleDeleted = onWorkspaceRoleDeletedFactory({
```
danielgak, line 991: @fabis94 please have a look here. 

My guess it's that the change wont affect this functionality but I'm maybe missing something.

Now that we filter by role this functions (`onWorkspaceRoleDeleted`, `onWorkspaceRoleUpdated` and `onWorkspaceSeatUpdated`) will only load the project if a user has a project acl on it. Prevously we were loading all projects with or without the acl.

packages/server/modules/core/repositories/streams.ts
danielgak: Does not really fit as we might not provide the userId, but only filter by workspaceId

gjedlicska: Depends on what the valid permutations are for the workspaceId, userId billable options. Can both be null? If yes it needs a generic name unfortunately

fabis94 (approved):

PR #5124, opened 2025-07-22
feat(fe2): conditional cross origin isolation on viewer page

fabis94:
Set `sharedArrayBufferHeaders=1` in querystring on viewer page, to enable

### DIFF
--- a/packages/server/modules/shared/middleware/security.ts
+++ b/packages/server/modules/shared/middleware/security.ts
@@ -6,3 +6,6 @@ export const crossOriginResourcePolicyMiddleware =
     res.setHeader('Cross-Origin-Resource-Policy', value)
     next()
   }
+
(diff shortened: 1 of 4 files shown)


packages/server/modules/shared/middleware/security.ts
```diff
@@ -0,0 +1,8 @@
+import type express from 'express'
+
+export const crossOriginResourcePolicyMiddleware =
+  (value: 'same-origin' | 'cross-origin' | 'same-site'): express.RequestHandler =>
+  (req, res, next) => {
+    res.setHeader('Cross-Origin-Resource-Policy', value)
```
fabis94, line 6: any cross-origin resources (previews, attachments etc.) coming from server must have this header set or the FE will be blocked from receiving them

Mikehrn (approved):

PR #5141, opened 2025-07-24
feat(fe): Update viewer UI/UX

Mikehrn:
Viewer UX refresh


packages/frontend-2/components/viewer/models/Card.vue
```diff
@@ -410,4 +414,69 @@ const onActionChosen = (params: { item: LayoutMenuItem }) => {
...
+  (selectedObjects) => {
+    const selectedIds = selectedObjects.map((obj: { id: string }) => obj.id)
+
+    if (selectedObjects.length > 0) {
+      expandToShowSelectedObjects(selectedIds)
+    } else {
+      forceExpandedNodeIds.value.clear()
+    }
+  },
+  { deep: true }
```
fabis94, line 478: this is a bit concerning, i assume that the objects can be pretty big/deep at times. is a shallow watcher not enough here? selectedObjects should update every time something gets added or removed anyway

packages/frontend-2/components/viewer/models/TreeItem.vue
```diff
@@ -444,4 +394,122 @@ const isolateOrUnisolateObject = () => {
...
+
+  disableScrollOnNextSelection.value = false
+})
+
+watch(
+  () => props.forceExpandedNodeIds,
+  () => {
+    checkAndExpand()
+  },
+  { deep: true }
```
fabis94, line 490: why use a deep watcher if its a single dimensional array/set? `deep` should be used sparingly because of the performance implications

AlexandruPopovici (dismissed):
Looks good to me!

AlexandruPopovici (dismissed):

benjaminvo (approved):

PR #5155, opened 2025-07-28
fix: publish server as latest too on release

danielgak:
The SDK integration tests rely on the `speckle-server:latest` image from dockerhub that we stop pushing affect moving the ci. This mr reenables this image push. 

It only runs on main and it also pushes the expected taged image too.

Checked on testing4 branch


gjedlicska (approved):
thanks, this is indeed important for catching early breaking changes with SDK vs server

PR #5157, opened 2025-07-28
Snap section box rotation to nearest 15 degrees

benjaminvo:
This adds the ability to snap rotation to a 15-degree grid when holding the shift key during section box rotation.

Cursor and I made it so that the snapping always aligns to the standard grid (0°, 15°, 30°, 45°, ...) rather than snapping relative to the initial rotation position.

@AlexandruPopovici I'm unsure if you actually want this to be part of the sectiontool extension or if this snapping should instead be implemented in the frontend. Let me know...

### DIFF
--- a/packages/viewer/src/modules/extensions/sections/SectionTool.ts
+++ b/packages/viewer/src/modules/extensions/sections/SectionTool.ts
@@ -136,6 +136,9 @@ export class SectionTool extends Extension {
     return [CameraController]
   }
 
+  /** Configurable rotation snap angle in radians. Set to null to disable snapping */
+  public rotationSnapAngle: number | null = Math.PI / 12 // 15 degrees by default
+
   /** This is our data model. All we need is an OBB */
   protected obb: OBB = new OBB()
 
@@ -961,7 +964,7 @@ export class SectionTool extends Extension {
   }
 
   /**
-   * Snaps a quaternion to the nearest 15-degree grid.
+   * Snaps a quaternion to the nearest grid based on rotationSnapAngle.
    * This is useful for rotation snapping.
    * @param q The quaternion to snap.
    * @returns The snapped quaternion.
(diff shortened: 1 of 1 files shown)


packages/viewer/src/modules/extensions/sections/SectionTool.ts
```diff
@@ -920,4 +959,37 @@ export class SectionTool extends Extension {
...
+   * This is useful for rotation snapping.
+   * @param q The quaternion to snap.
+   * @returns The snapped quaternion.
+   */
+  protected snapQuaternionToGrid(q: Quaternion): Quaternion {
+    /** Convert quaternion to Euler angles using pooled object */
+    this.tempEuler.setFromQuaternion(q)
+
+    /** Snap each axis to 15-degree increments (π/12 radians) */
+    const snapAngle = Math.PI / 12 // 15 degrees
```
AlexandruPopovici, line 974: I would turn this into an extension option and make it configurable so that we can have any snap value including none

AlexandruPopovici (changes requested):
This can and should be handled by the extension rather than be application specific. Apart from the minor comments, everything else looks good. As long as it's working properly, we good

AlexandruPopovici:
There is another thing to consider though. The PR only treats the angle snap when the section box is set via mouse input. But section boxes can and will be set programmatically. For a complete solution you'd want to handle that as well.

benjaminvo:
@AlexandruPopovici Thanks a lot, I'll have a look at your comments!

> But section boxes can and will be set programmatically. For a complete solution you'd want to handle that as well.

I wonder, though, if it's relevant to also support incremented rotations for section boxes set programmatically?

AlexandruPopovici:
> I wonder, though, if it's relevant to also support incremented rotations for section boxes set programmatically?

Might not be an urgent need, but in order to truly have the section box snapping to fixed angle steps, you would. If you don't want to handle this now, you can add a comment specifying this caveat

AlexandruPopovici (approved):
Looks good!

PR #5159, opened 2025-07-29
[WIP] feat: proposal structre for 2pc operations in multiregion

danielgak:


danielgak:
@cdriesler theck the test flags here, as this test run in normal mode with the multiregion flag on. If you try the `test:multiregion` that tries to do the whole avien set up

PR #5163, opened 2025-07-30
WIP: Saved Views Foundation

fabis94:
## Checklist:


packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -0,0 +1,130 @@
...
+</template>
+<script setup lang="ts">
+import { omit } from 'lodash-es'
+import { usePaginatedQuery } from '~/lib/common/composables/graphql'
+import { graphql } from '~/lib/common/generated/gql'
+import type { ViewerSavedViewsPanelViewsGroup_SavedViewGroupFragment } from '~/lib/common/generated/gql/graphql'
+import { useInjectedViewerState } from '~/lib/viewer/composables/setup'
+
+graphql(`
+  fragment ViewerSavedViewsPanelViewsGroup_SavedViewGroup on SavedViewGroup {
```
fabis94, line 36: one fragment we get from ancestor query, the other we use in the paginated view query that runs from this query

packages/frontend-2/components/viewer/selection/Sidebar.vue
```diff
@@ -205,36 +204,30 @@ onKeyStroke('Escape', () => {
 })
 
 watch(
-  () => objects.value.length,
-  (newLength) => {
-    // Dont open sidebar if a comment is open
-    if (newLength !== 0 && !focusedThreadId.value) {
-      sidebarOpen.value = true
-    } else if (newLength === 0) {
+  [
```
fabis94, line 207: watchers were conflicting with each other and race conditions were happening, so i combined them all into 1 watcher

packages/frontend-2/middleware/requireValidProject.ts
```diff
@@ -28,14 +28,18 @@ export default defineNuxtRouteMiddleware(async (to) => {
...
+
   const { data, errors } = await client
     .query({
       query: projectAccessCheckQuery,
       variables: { id: projectId },
       context: {
         skipLoggingErrors: true
       },
-      fetchPolicy: 'network-only'
+      fetchPolicy: isSamePathAndQuery ? 'cache-first' : 'network-only'
```
fabis94, line 42: cc @Mikehrn @andrewwallacespeckle 

turns out every change to the querystring _or_ the anchor/hash state (e.g. #threadId in the viewer page) causes all of the middlewares to refire, cause technically its a new navigation.

this causes a lot of the queries to re-fire, especially if they have the `network-only` fetchPolicy (which would otherwise make sense). so we have to explicitly check in middlewares if the navigation isn't actually just a minor querystring/anchor string change that shouldn't actually cause full access rechecking w/ refreshed queries

packages/server/.vscode/settings.json
```diff
@@ -1,6 +1,11 @@
 {
-  "mochaExplorer.env": { "NODE_ENV": "test" },
-  "mochaExplorer.nodeArgv": ["--import", "tsx"],
+  "mochaExplorer.env": { "NODE_ENV": "test", "TSX": "true" },
```
fabis94, line 2: native TS support in IDE `server` test runner

packages/server/assets/core/typedefs/common.graphql
```diff
@@ -45,4 +45,8 @@ type PermissionCheckResult {
   code: String!
   message: String!
   payload: JSONObject
+  """
+  Same as message, or undefined if check is authorized
+  """
+  errorMessage: String
```
fabis94, line 51: message, but only if not successful. useful for FE purposes (tooltips)

packages/server/modules/viewer/domain/operations/resources.d.ts
```diff
@@ -0,0 +1,13 @@
+import type { ViewerUpdateTrackingTarget } from '@/modules/core/graph/generated/graphql'
```
fabis94, line 1: .d.ts files (type only TS files) get ignored by coverage, otherwise u get a large untested & untestable file

packages/frontend-2/lib/viewer/composables/activity.ts
```diff
@@ -78,6 +88,57 @@ graphql(`
   }
 `)
 
+const useViewerRealtimeActivityState = () =>
+  useState('viewer_realtime_activity_state', () => ({
+    activity: undefined as Optional<ViewerActivityMetadata>,
+    status: ViewerUserActivityStatus.Viewing as ViewerUserActivityStatus
+  }))
+
+export const useViewerRealtimeActivityTracker = () => {
```
fabis94, line 97: serializing the viewer state and comparing if its the same as a previous one are relatively expensive operations, ones we already try to do on every change to the viewer state for activity tracking.

so i made sure that collected activity is stored in a global singleton state, one accessible through this new composable anywhere in the viewer, so that you can easily get the state and its unique ID (changes w/ every change to the state) anywhere without having to re-serialize

this will be useful whenever you want to track if there's a change from some previously established state - e.g. you want to see if the user "exited" the saved view by making any changes to the viewport, and you can do it by comparing the state ids

cdriesler (approved):
LGTM

One note on style. This is an old opinion of mine, so the thinking may have shifted, but I consider `.d.ts` to be almost always meant for build artefacts. Sometimes, as a library, you have to manually write them to augment a type of your own dependency, but they're still shipped to your consumers as "just part of" the pile of `.d.ts` they're already getting.

If it's only being done here to omit the untestable types from codecov, I'd reach for configuring codecov to like ignore `/domain` or something and move `logic.ts` stuff to helpers. But I appreciate that this works and is pretty harmless.

One benefit too is that `.d.ts` forces everything to be ambient, so no sneaking logic into these like we do sometimes.

(tldr I don't want to hold the PR on this, but we should revisit other options if/when `.d.ts` gets awkward)

packages/shared/src/environment/featureFlags.ts
```diff
@@ -0,0 +1,52 @@
...
+export type FeatureFlags = {
+  [K in (typeof featureFlags)[number]]: boolean
+}
+
+/**
+ * Load feature flags in-memory, so that @speckle/shared internals can use them too.
+ * These are supposed to be initialized on app startup/bootstrapping
+ */
+
+export const loadFeatureFlags = (input?: Partial<FeatureFlags>): FeatureFlags => {
```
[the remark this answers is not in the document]
fabis94: nvm, undo. the problem in speckle/shared is that we need to be able to change active feature flags in realtime while running tests. a global feature flag state like this muddies the water.

even tho its also not ideal, the better option is to feed in feature flags when loading workspace plan features

cdriesler (approved):

PR #5227, opened 2025-08-12
refactor(fe): Models panel with versions and diff

andrewwallacespeckle:
Tidying up some tech debt here too. 

Original task was making a click of models panel button go back to models, when in versions. 

See ticket for more context.

### DIFF
--- a/packages/frontend-2/components/viewer/controls/Left.vue
+++ b/packages/frontend-2/components/viewer/controls/Left.vue
@@ -133,7 +133,7 @@
       :style="`width: ${widthClass};`"
     >
       <KeepAlive v-show="activePanel === 'models'">
-        <ViewerModelsPanel />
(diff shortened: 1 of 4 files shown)


packages/frontend-2/components/viewer/controls/Left.vue
```diff
@@ -133,7 +133,7 @@
       :style="`width: ${widthClass};`"
     >
       <KeepAlive v-show="activePanel === 'models'">
-        <ViewerModelsPanel />
+        <ViewerModelsPanel v-model:sub-view="modelsSubView" />
```
fabis94, line 136: not necessarily within the scope of this task, but maybe we can also clean up some of those KeepAlives/v-shows we see in the viewer? the less inactive components we keep alive, the more performant the experience is

PR #5238, opened 2025-08-13
feat(acc): revamp sync item add flow

oguzhankoral:
ACC has a nested structure to retrieve files, which kinda stands for our structure

| Speckle | ACC |
|--------|--------|
| Workspace | Hub |
| Project | Project |
| Model | Folder (📍 we were skipping this) |
| Version | File (📍 and were flattening all files from all hubs and projects) |

📍: mentioned previous implementation was causing a bottleneck on the ACC side since we fetch details per file to retrieve its metadata. 

Now we fetch data on demand:
1. Build the folder tree for navigation panel
2. Once user clicked a folder - we fetch the files and their relevant metadata

Other improvements
- `Syncs.vue` file was dumping ground before - now we have all subcomponents and composables. I am more confident now to handover for design + FE
- Model selection was terrible dropdown even without search or pagination. Now we have proper model selection and search

https://github.com/user-attachments/assets/624d0432-27e7-4866-aa4b-e18f9bc090be

### DIFF
--- a/packages/server/modules/acc/domain/constants.ts
+++ b/packages/server/modules/acc/domain/constants.ts
@@ -3,8 +3,8 @@ import { StringEnum } from '@speckle/shared'
 
 export const ImporterAutomateFunctions = {
   svf2: {
-    functionId: '2909d29a9d',
-    functionReleaseId: 'ede3667b40'
+    functionId: '4665e0b3ba',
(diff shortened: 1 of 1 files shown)


cdriesler (dismissed):
LGTM

PR #5241, opened 2025-08-14
feat(tokens): enable resource limits for PAT creation

cdriesler:
## Description & motivation

- PowerBI connector requires ability to:
  - Create tokens for the given user
  - Scope them to just a given project
- Ability to do this is largely written, just not exposed by the public API

## Changes:

- Make it available via gql
- Gives PowerBI default app `tokens:write` scope

### DIFF
--- a/packages/frontend-2/lib/common/generated/gql/graphql.ts
+++ b/packages/frontend-2/lib/common/generated/gql/graphql.ts
@@ -250,6 +250,8 @@ export type ApiToken = {
 
 export type ApiTokenCreateInput = {
   lifespan?: InputMaybe<Scalars['BigInt']['input']>;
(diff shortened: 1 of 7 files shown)


cdriesler:
I think patch coverage is 100% because PAT and app tokens both go to the same (already tested) paths. I more or less specified an optional argument that was missing. (It does work out of the box because of this) tldr they both make (1) the same token and (2) a light metadata record in their own tables (which we use to query/show in developer settings). 

We check scope/resource limits on those tokens, and do nothing with the metadata record during auth. By the time we've hit the resolver we have no (automatic) way to know if it's an app or PAT token (but can go find that information if we really wanted to)

Tests might be necessary if we ever surface this ability/information on the FE? Merging as-is but will revisit tests if someone raises a flag

PR #5246, opened 2025-08-14
Feat: Cursor code rules

Mikehrn:
While Cursor is decent at grasping code style and guidelines it still quite often messes things up, especially around ordering in files, naming, documenting etc. I've created this set of rules that aims to combine all our code styles and conventions (both written and unwritten) and help point Cursor to the correct config files (for example for tailwind classes or linting rules).

It looks like a lot of rules, but when trying it out it doesnt really impact performance, as not all rules are always applied. Curious to get your opinion on this ruleset and if it's useful to commit.


.cursor/mcp.json
```diff
@@ -0,0 +1,18 @@
+{
+  "mcpServers": {
```
Mikehrn, line 2: @andrewwallacespeckle @benjaminvo This MCP gives cursor access to Playwright, it can actually interact with the browser in various ways. This is very useful to give it more context about what it's working on, and it can also debug easier. For example, you dont have add 10 random console logs to copy paste when debugging as it can get that context from the brower

Mikehrn: Screenshot 2025-08-15 at 09 54 32

andrewwallacespeckle (approved):
I say we get this in as soon as we can, and we can tweak it over time. Nice work!

PR #5249, opened 2025-08-15
feat(fe): improve filter by property

andrewwallacespeckle:
- Consolidated filtering logic from `Panel` and `KeyValuePair` into shared `useFilterUtilities` composable
- Create reusable filter utilities (`isKvpFilterable`, `getFilterDisabledReason`, `applyKvpFilter`) to replace logic in `KeyValuePair`
- Fix limited filtering from selection info - previously only worked for properties with exact backend key matches, leaving many properties unfilterable
- Implement property matching that handles complex nested properties by matching display names to backend filter keys

### DIFF
--- a/packages/frontend-2/components/viewer/selection/KeyValuePair.vue
+++ b/packages/frontend-2/components/viewer/selection/KeyValuePair.vue
@@ -66,15 +66,19 @@ import { LayoutMenu, type LayoutMenuItem } from '@speckle/ui-components'
 import { Ellipsis } from 'lucide-vue-next'
 import { useFilterUtilities } from '~~/lib/viewer/composables/ui'
 import { useInjectedViewer } from '~~/lib/viewer/composables/setup'
-import type { PropertyInfo } from '@speckle/viewer'
 
 const props = defineProps<{
-  kvp: Record<string, unknown> & { key: string; value: unknown; units?: string }
+  kvp: Record<string, unknown> & {
+    key: string
+    value: unknown
+    units?: string
+    backendPath?: string
+  }
 }>()
 
 const showActionsMenu = ref(false)
 
-const { setPropertyFilter, applyPropertyFilter, isPropertyFilterable } =
+const { isKvpFilterable, getFilterDisabledReason, applyKvpFilter } =
   useFilterUtilities()
 const {
   metadata: { availableFilters }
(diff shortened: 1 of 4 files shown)


packages/frontend-2/components/viewer/selection/KeyValuePair.vue
```diff
@@ -87,17 +91,24 @@ const isCopyable = (kvp: Record<string, unknown>) => {
 }
 
 const isFilterable = (kvp: Record<string, unknown>) => {
-  const key = kvp.key as string
-  return isPropertyFilterable(key, availableFilters.value)
+  return isKvpFilterable(
+    kvp as Record<string, unknown> & { key: string; backendPath?: string },
```
fabis94, line 95: whats the reason for the type cast? same thing in props - can we really not be sure what its gonna be? if that is the case, we should use type guards cause apparently it can be many things. if we are sure its gonna be that type, why is it typed wrong in the first place? 🤔

fabis94 (commented):
a lot of force type casting. maybe its justified, but id like to hear more

packages/frontend-2/components/viewer/selection/KeyValuePair.vue
andrewwallacespeckle: Sorry, I should have explained. I am not 100% on what is the most useful implementation here, and was talking about it with Jonathon and Bilal [here.](https://discord.com/channels/726756379083145269/1405894754964738069/1405900969367179284)

I said I would make this change to get feedback, and was lazy and didn't type it properly incase it wasn't what they wanted, but I've done it properly now.

packages/frontend-2/components/viewer/models/versions/Card.vue
```diff
@@ -197,12 +197,12 @@ async function handleViewChanges(version: ViewerModelVersionCardItemFragment) {
 
 const highlightObject = () => {
   const refObject = props.model.loadedVersion.items[0]?.referencedObject
-  if (refObject) highlightObjects([refObject])
+  if (refObject && typeof refObject === 'string') highlightObjects([refObject])
```
fabis94, line 200: this is fine, but in the future u may find lodash's isString to be less verbose

fabis94 (approved):

PR #5251, opened 2025-08-15
fix(frontend-2): make workspace member list text colors consistent

Mikehrn:
**Fix workspace member list text color inconsistencies**
Make all workspace member list components use consistent dark text colors (text-foreground) instead of mixing light and dark colors across different elements.


Mikehrn:
Generated this PR with one command, that reads the Linear ticket, makes the changes and opens the PR using the Linear and Github MCPs. Seems to work well for simple changes like this (albeit it's very lazy), curious to see if it can get decent results with more complicated tickets, probably needs a very clear description and some checkpoints set up

benjaminvo (approved):

PR #5252, opened 2025-08-15
fix(frontend-2): update workspace switcher cache when joining from discoverable workspaces

Mikehrn:
**Fix workspace switcher not updating when joining from discoverable workspaces modal**
Updates Apollo cache to include newly joined workspace in workspace list, ensuring immediate UI update without requiring hard refresh.


benjaminvo:
So what's the exact workflow for these PRs?

PR #5256, opened 2025-08-18
fix(fe): preload navigationWorkspaceSwitcherQuery

andrewwallacespeckle:
image

Preload navigationWorkspaceSwitcherQuery to avoid empty workspace switcher on refresh.


fabis94:
@andrewwallacespeckle don't we need FF checks behind the query? or is it fine to always fire

PR #5259, opened 2025-08-19
Always support drag and drop in the Models panel 

benjaminvo:
At the moment you can only drag and drop a file into a project if it's empty. This PR wraps a file upload drop zone around the model list (both card view and list view) so that you can also drag and drop files into the Models tab when it already has models. You can only drop one file at a time.  

CleanShot 2025-08-18 at 17 40 32@2x

### DIFF
--- a/packages/frontend-2/components/project/page/models/Tab.vue
+++ b/packages/frontend-2/components/project/page/models/Tab.vue
@@ -36,15 +36,9 @@
 
         <div
           v-if="isDraggingFiles && canCreateModel"
-          class="absolute inset-0 bg-primary/10 border border-dashed border-primary rounded-lg flex items-center justify-center z-10"
+          class="absolute inset-0 bg-primary/10 border border-dashed border-primary rounded-lg flex items-center justify-center z-10 bg-white/50 dark:bg-black/50 text-center text-heading text-primary"
         >
-          <div class="text-center p-8">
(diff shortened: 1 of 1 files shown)


packages/frontend-2/components/project/page/models/Tab.vue
```diff
@@ -10,16 +10,63 @@
...
+          :source-apps="selectedApps"
+          :contributors="selectedMembers"
+          :project="project"
+          :project-id="projectId"
+          @clear-search="clearSearch"
+        />
+
+        <div
+          v-if="isDraggingFiles && canCreateModel"
+          class="absolute inset-0 bg-primary/10 border border-dashed border-primary rounded-lg flex items-center justify-center z-10"
```
andrewwallacespeckle, line 39: Image

The background opacity isn't working here. These opacity modifiers never work for me with our custom colours. 

You might be able to use `bg-opacity-10`, but that may not work either. 

Worst case scenario, add `relative` to the parent, and absolutely position a div with `absolute inset-0 bg-primary opacity-10`. 

Then add `relative z-10` to the items that should sit in front. 

Let me know if you're busy in London and want me to do this.

benjaminvo: CleanShot 2025-08-19 at 12 59 52@2x

OK I changed the styling a bit. Removed some text and added a transparent background inside it.

PR #5275, opened 2025-08-20
fix(fe): prevent native context menu on Windows

andrewwallacespeckle:
Still getting reports of the default right click menu appearing in the viewer, when it should be prevented to allow our context menu. 

I don't have Windows, so can't test, but I'm assuming the problem is that `preventDefault()` is called too late in  the `ObjectClicked` handler, and the default menu has already been shown. 

Used `useEventListener` to catch this earlier.


fabis94 (approved):
tested, seems to work

PR #5276, opened 2025-08-20
fix(fe2): full cache reset on log out

fabis94:
we used to just evict activeUser, but there's a bunch of user-specific keys like Workspace.seatType that got retained and caused weird cache pollution where if you logged out and logged in to a different account, the new account would have some values cached from the previous one

the correct thing to do is to evict the entire cache on log out

### DIFF
--- a/packages/frontend-2/pages/settings/workspaces/[slug]/general.vue
+++ b/packages/frontend-2/pages/settings/workspaces/[slug]/general.vue
@@ -157,7 +157,7 @@
           </div>
         </div>
       </template>
-      <template v-if="(isServerAdmin || isAdmin) && workspaceId">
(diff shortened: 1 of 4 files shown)


packages/frontend-2/pages/settings/workspaces/[slug]/general.vue
```diff
@@ -157,7 +157,7 @@
           </div>
         </div>
       </template>
-      <template v-if="(isServerAdmin || isAdmin) && workspaceId">
+      <template v-if="workspaceResult?.workspaceBySlug?.id">
```
Comment (fabis94) [inline packages/frontend-2/pages/settings/workspaces/[slug]/general.vue:160]: doesnt make sense that this would be admin only

PR #5289, opened 2025-08-21
feat(sso): allow sso session duration overrides

cdriesler:
For Pelicad, see [forum thread](https://speckle.community/t/personal-access-token-stops-working-when-workspace-sso-session-expires/18858).

SSO sessions currently default to a (dubious but appropriate) 7 days because user visit the frontend enough to keep this up to date and easy to refresh. Pelicad drive some processes with "headless" users that won't ever visit the frontend and the 7 day limit is too punitive.

I am going to update this value for them manually, we can revisit exposing this publicly as a setting if it ever comes up again.

Tested on `testing2`


andrewwallacespeckle (approved):
Here's an approval if you want to get this out

gjedlicska (dismissed):

PR #5292, opened 2025-08-21
chore(fe2): even more performant and even less hydration mismatches

fabis94:
## Checklist:

- [ ] My pull request follows the guidelines in the [Contributing guide](https://github.com/specklesystems/speckle-server/blob/main/CONTRIBUTING.md)?
- [ ] My pull request does not duplicate any other open [Pull Requests](../../pulls) for the same update/change?
- [ ] My commits are related to the pull request and do not amend unrelated code or documentation.
- [ ] My code follows a similar style to existing code.
- [ ] I have added appropriate tests.
- [ ] I have updated or added relevant documentation.

PR #5295, opened 2025-08-22
Revert "chore(fe2): even more performant and even less hydration mismatches"

fabis94:
Reverts specklesystems/speckle-server#5292

in case everything implodes


fabis94:
@iainsproat nope

PR #5296, opened 2025-08-22
fix(fe2): use users time zone in SSR

fabis94:
image

same date is rendered in 1 time zone in SSR (server's time zone) and in another one in CSR, this causes a hydration mismatch

### DIFF
--- a/packages/frontend-2/components/preview/Image.vue
+++ b/packages/frontend-2/components/preview/Image.vue
@@ -102,6 +102,7 @@ const finalPreviewTransitioner = ref(
   null as Nullable<{ triggerTransition: () => Promise<void> }>
 )
 
+const { $isAppHydrated } = useNuxtApp()
 const isInViewport = useElementVisibility(parent)
 const basePreviewUrl = computed(() => props.previewUrl)
 const {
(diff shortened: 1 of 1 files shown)


packages/frontend-2/lib/projects/composables/previewImage.ts
```diff
@@ -287,6 +287,10 @@ export function usePreviewImageBlob(
      * Run this at the bottom of the component to fully initialize it
      */
     init: async () => {
+      if (!eagerLoad && import.meta.server) {
```
fabis94, line 290: unrelated - tryna deal w/ more hydration mismatches

packages/frontend-2/components/preview/Image.vue
```diff
@@ -102,6 +102,7 @@ const finalPreviewTransitioner = ref(
   null as Nullable<{ triggerTransition: () => Promise<void> }>
 )
 
+const { $isAppHydrated } = useNuxtApp()
```
fabis94, line 105: super annoying! there are preview image related mismatches only happening in prod, and while i've dealt w/ most of them, some still remain

PR #5301, opened 2025-08-22
Daniel/refactor replicate query to multiregion operation

danielgak:
## Checklist:


packages/server/modules/auth/index.ts
```diff
@@ -91,36 +80,60 @@ const finalizeInvitedServerRegistration = finalizeInvitedServerRegistrationFacto
 })
 const resolveAuthRedirectPath = resolveAuthRedirectPathFactory()
 
-const buildCreateUser = async () => {
-  const regionClients = await getRegisteredRegionClients()
-  const regionDbs = Object.values(regionClients)
+const createUser: CreateValidatedUser = async (...input) =>
```
danielgak, line 83: @cdriesler the `setupStrategies` flow seemed a bit too long to wrap in this multiregional transaction, so I ended up just wrapping the user creation cases. I'm not too familiar with this this flow do you think it's okey?

packages/server/modules/activitystream/tests/integration/activity.graph.spec.ts
```diff
@@ -80,7 +79,7 @@ const requestNewEmailVerification = requestNewEmailVerificationFactory({
 const createUser = createUserFactory({
   getServerInfo: getServerInfoFactory({ db }),
   findEmail,
-  storeUser: replicateQuery([db], storeUserFactory),
+  storeUser: storeUserFactory({ db }),
```
danielgak, line 82: As we do not have a `replicateQuery` anymore, and we do not use a transaction, we can use the operator as normal for those cases were we do not need replication _(specs basically)_

packages/server/modules/core/graph/resolvers/users.ts
danielgak, line 1: Create, update and delete user

packages/server/modules/shared/command.ts
```diff
@@ -120,3 +128,117 @@ export const asOperation = async <T>(
     }
   )
 }
+
+export const asMultiregionalOperation = async <T, K extends [Knex, ...Knex[]]>(
+  operation: (args: {
+    txs: Knex.Transaction[]
```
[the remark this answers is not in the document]
danielgak: @gjedlicska sure.
`asMultiregionOperation` does not expect as starting transaction, but rather expect the operation function (the function that we pass to be atomic) to consume a started transaction .

PR #5303, opened 2025-08-25
feat(gatekeeper): add per workspace feature flags

gjedlicska:
## Checklist:


packages/server/modules/gatekeeper/services/workspacePlans.ts
```diff
@@ -59,7 +63,14 @@ export const updateWorkspacePlanFactory =
...
           case 'canceled':
           case 'paymentFailed':
-            workspacePlan = { workspaceId, status, name, createdAt, updatedAt }
+            workspacePlan = {
+              workspaceId,
+              status,
+              name,
+              createdAt,
+              updatedAt,
+              featureFlags: WorkspaceFeatureFlags.none
```
danielgak, line 72: Are we then erasing all FF in case paymentFailed stripe callback then? Seems a bit unexpected.

packages/server/modules/gatekeeper/services/workspacePlans.ts
```diff
@@ -77,7 +88,14 @@ export const updateWorkspacePlanFactory =
...
             if (workspaceSubscription) throw new InvalidWorkspacePlanStatus()
 
-            workspacePlan = { workspaceId, status, name, createdAt, updatedAt }
+            workspacePlan = {
+              workspaceId,
+              status,
+              name,
+              createdAt,
+              updatedAt,
+              featureFlags: WorkspaceFeatureFlags.none
```
danielgak, line 97: same here 🤔

packages/server/modules/gatekeeper/services/checkout.ts
```diff
@@ -79,7 +79,8 @@ export const completeCheckoutSessionFactory =
       updatedAt: new Date(),
       workspaceId: checkoutSession.workspaceId,
       name: checkoutSession.workspacePlan,
-      status: 'valid'
+      status: 'valid',
+      featureFlags: WorkspaceFeatureFlags.none
```
danielgak, line 83: why not use `previousWorkspacePlan.featureFlags` here?

Looking at other changes, seems that the behaviour is that on plan change we loose the enabled FF.

danielgak (approved):
Good one. 

A bit weird to me that a `paymentFail`, or other flows disable the flags as it's something that I personally would end up forgetting. 

I also checked how the queries to interact with this field would look like and postgres has all this operators in place to perform bulk udates etc.  

So to me, all good 👍

danielgak (approved):

PR #5323, opened 2025-08-27
refactor(multiregion): split delete stream method

danielgak:
I can't have the factory method `deleteStream` deleting also from the commit table, so I've moved this logic out from the repo layer call outside as a service method, so every factory that uses this `deleteStream` needs to define the target db (so I can properly wrap the multiregion operator) and make deletes for `projectDb/main` or just main as needed.


danielgak:
(target branch will get updated to main)

PR #5351, opened 2025-09-01
feat(fe2): save/reload measurements integration

fabis94:
## Checklist:

### DIFF
--- a/packages/frontend-2/lib/viewer/composables/setup/measurements.ts
+++ b/packages/frontend-2/lib/viewer/composables/setup/measurements.ts
@@ -1,0 +1,118 @@
+import {
+  MeasurementType,
+  type MeasurementData,
+  type MeasurementOptions
+} from '@speckle/shared/viewer/state'
+import type { Measurement } from '@speckle/viewer'
+import {
+  MeasurementEvent,
+  MeasurementsExtension,
+  MeasurementState
+} from '@speckle/viewer'
+import { onKeyStroke, watchTriggerable } from '@vueuse/core'
+import { useInjectedViewerState } from '~/lib/viewer/composables/setup'
+import { useMeasurementUtilities } from '~/lib/viewer/composables/ui'
+import { useOnViewerLoadComplete } from '~/lib/viewer/composables/viewer'
+
+export const useMeasurementsSetup = () => {
+  return {
+    enabled: ref(false),
+    options: ref<MeasurementOptions>({
+      visible: true,
+      type: MeasurementType.POINTTOPOINT,
+      units: 'm',
+      vertexSnap: true,
+      precision: 2
+    }),
+    measurements: ref([] as MeasurementData[])
+  }
+}
+
+export const useMeasurementsPostSetup = () => {
+  const {
+    viewer: {
+      instance,
+      init: { promise: viewerInitialized }
+    },
+    ui: { measurement }
+  } = useInjectedViewerState()
+  const { reset, removeActiveMeasurement } = useMeasurementUtilities()
+
+  const measurementsInstance = () => instance.getExtension(MeasurementsExtension)
+
+  // state -> viewer
+  const { trigger: triggerEnabledWatch } = watchTriggerable(
+    measurement.enabled,
+    (newVal, oldVal) => {
+      if (newVal !== oldVal) {
+        instance.enableMeasurements(newVal)
+      }
+    },
+    { immediate: true }
+  )
+
+  const { trigger: triggerOptionsWatch } = watchTriggerable(
+    measurement.options,
+    (newVal) => {
+      if (newVal) {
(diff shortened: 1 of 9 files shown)


packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -643,54 +646,31 @@ export function useMeasurementUtilities() {
     state.ui.measurement.options.value = options
   }
 
-  const removeMeasurement = () => {
```
fabis94, line 646: same problem viewModes had - state creation, the actual integration and helpers/utilities are all mixed together and needed refactoring so that each of those happens in its own dedicated computed

packages/frontend-2/lib/viewer/composables/setup/measurements.ts
```diff
@@ -0,0 +1,117 @@
...
+  MeasurementEvent,
+  MeasurementsExtension,
+  MeasurementState
+} from '@speckle/viewer'
+import { onKeyStroke, watchTriggerable } from '@vueuse/core'
+import { useInjectedViewerState } from '~/lib/viewer/composables/setup'
+import { useMeasurementUtilities } from '~/lib/viewer/composables/ui'
+import { useOnViewerLoadComplete } from '~/lib/viewer/composables/viewer'
+
+export const useMeasurementsSetup = () => {
```
fabis94, line 17: 1. state create

packages/frontend-2/lib/viewer/composables/setup/measurements.ts
```diff
@@ -0,0 +1,117 @@
...
+  return {
+    enabled: ref(false),
+    options: ref<MeasurementOptions>({
+      ...defaultMeasurementOptions
+    }),
+    measurements: ref([] as MeasurementData[])
+  }
+}
+
+export const useMeasurementsPostSetup = () => {
```
fabis94, line 27: 2. viewer integration

packages/frontend-2/lib/viewer/composables/ui.ts
```diff
@@ -631,9 +633,10 @@ export function useThreadUtilities() {
 export function useMeasurementUtilities() {
```
fabis94, line 633: 3. optional helpers

packages/viewer/src/modules/extensions/measurements/MeasurementsExtension.ts
```diff
@@ -423,6 +437,36 @@ export class MeasurementsExtension extends Extension {
     }
   }
 
+  /**
+   * Idempotent way of setting measurements
+   */
+  public setMeasurements(measurements: MeasurementData[]) {
```
fabis94, line 443: cc @AlexandruPopovici these changes seem to work

ultimately we need an idempotent setMeasurements() function that will take care of removing unnecessary ones and adding necessary ones, and do nothing if nothing needs to be done

packages/viewer/src/modules/extensions/measurements/MeasurementsExtension.ts
```diff
@@ -408,13 +409,26 @@ export class MeasurementsExtension extends Extension {
...
   }
 
-  public removeMeasurement() {
-    if (this._selectedMeasurement) {
-      this._measurements.splice(
-        this._measurements.indexOf(this._selectedMeasurement),
-        1
-      )
-      this.renderer.scene.remove(this._selectedMeasurement)
+  private measurementDataId(measurementData: MeasurementData) {
```
fabis94, line 412: JSON.stringify to get id is not ideal, but seems to be the best option

AlexandruPopovici: I don't think it is practical. Why don't we just generate an uuid when creating them, and add an `id` filed in the data model?

fabis94: the benefit of JSON.stringify is that if two measurements are structurally identical, they are treated as identical. I was concerned about the same measurement somehow arriving w/ multiple uuids because they could get re-generated, but I think you're right and it should be possible to avoid this

packages/viewer/src/modules/extensions/measurements/Measurement.ts
```diff
@@ -28,11 +29,17 @@ export abstract class Measurement extends Object3D {
...
   public precision = 2
+  public measurementId: string
 
   protected _state: MeasurementState = MeasurementState.HIDDEN
   protected renderingCamera: Camera | null
   protected renderingSize: Vector2 = new Vector2()
 
+  constructor() {
+    super()
+    this.measurementId = MathUtils.generateUUID()
```
fabis94, line 40: @AlexandruPopovici new ID mechanism, no JSON.stringify needed anymore

PR #5353, opened 2025-09-01
Support drag and drop for moving views into groups

benjaminvo:
1. Adds support for drag and drop for moving views between groups.
2. Reintroduces the "Ungrouped" group. It's actually nice – anyway – that you can hide these views + it's needed for ungrouping via drag and drop.
 
CleanShot 2025-09-01 at 15 26 44@2x


packages/frontend-2/components/viewer/saved-views/panel/view/MoveDialog.vue
```diff
@@ -84,8 +84,12 @@ const buttons = computed((): LayoutDialogButton[] => [
 
 const onSubmit = handleSubmit(async (values) => {
   if (!props.view) return
-  const groupId = values.group.id !== props.view.group.id ? values.group.id : null
-  if (!groupId) return
+  const selectedGroupId = values.group.isUngroupedViewsGroup ? null : values.group.id
```
fabis94, line 87: ditto

packages/frontend-2/components/viewer/saved-views/panel/view/MoveDialog.vue
```diff
@@ -97,7 +101,7 @@ const onSubmit = handleSubmit(async (values) => {
   })
 
   if (res?.id) {
-    emit('success', groupId)
+    emit('success', groupId || 'ungrouped')
```
fabis94, line 104: it should be `string | null` w/ null signifying ungrouped

packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -169,6 +175,104 @@ const menuItems = computed((): LayoutMenuItem<MenuItems>[][] => [
   ]
 ])
 
+const onDragOver = (_event: DragEvent) => {
```
fabis94, line 178: the convention is that `_var` is only for unused variables, you're using `_event` everywhere

packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -50,13 +55,6 @@
       </div>
     </template>
   </LayoutDisclosure>
-  <ViewerSavedViewsPanelViewsGroupInner
-    v-else
```
fabis94, line 54: IMO u should've just wrapped the entire component w/ a new div that u add all of the dragging handlers on. 
if there's no longer this v-else, how are we ensuring auto-open for the ungrouped group?

packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -184,7 +288,7 @@ const onActionChosen = async (item: LayoutMenuItem<MenuItems>) => {
 
 const onAddGroupView = async () => {
   await createView({
-    groupId: props.group.id
+    groupId: isUngroupedGroup.value ? null : props.group.id
```
fabis94, line 291: view.groupId should work

packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -140,12 +141,17 @@ const { triggerNotification } = useGlobalToast()
...
 const updateGroup = useUpdateSavedViewGroup()
+const updateView = useUpdateSavedView()
 const renameMode = defineModel<boolean>('renameMode')
 
 const open = defineModel<boolean>('open')
 const showMenu = ref(false)
 const menuId = useId()
 
+// Drag and drop state
+const isDragOver = ref(false)
```
fabis94, line 152: [would be nicer](https://vuejs.org/guide/reusability/composables#extracting-composables-for-code-organization) to keep all of this in its own composable under `~/lib` so that dragging internals aren't mixed together with all of the other logic

packages/frontend-2/components/viewer/saved-views/panel/views/Group.vue
```diff
@@ -169,6 +175,104 @@ const menuItems = computed((): LayoutMenuItem<MenuItems>[][] => [
...
+    // Move the view to this group using the enhanced drag data
+    const success = await moveViewToGroup(data)
+
+    if (success) {
+      // Auto-open the group if it was closed
+      if (!open.value) {
+        open.value = true
+      }
+
+      triggerNotification({
```
fabis94, line 221: we probably want these to always fire whenever a view is moved any way (through dragging or otherwise), so all of this should happen inside of `moveViewToGroup`

packages/frontend-2/components/viewer/saved-views/panel/View.vue
```diff
@@ -252,17 +261,54 @@ const menuItems = computed((): LayoutMenuItem<MenuItems>[][] => [
...
 
+const onDragStart = (event: DragEvent) => {
+  if (!canUpdate.value?.authorized || isLoading.value) {
+    event.preventDefault()
+    return
+  }
+
+  isDragging.value = true
+
+  const dragData = {
```
fabis94, line 290: this should be TS typed so u can ensure correct props are set here and then type the dragData on the other end when receiving it, to have a type for the result

fabis94 (approved):

PR #5370, opened 2025-09-03
feat(fe2): saved views panel keepalive

fabis94:
## Checklist:


packages/frontend-2/components/viewer/controls/Left.vue
```diff
@@ -154,10 +154,12 @@
         :summary="summary"
       />
       <ViewerDataviewerPanel v-if="activePanel === 'devMode'" />
-      <ViewerSavedViewsPanel
-        v-if="isSavedViewsEnabled && activePanel === 'savedViews'"
-        @close="activePanel = 'none'"
-      />
+      <KeepAlive>
```
fabis94, line 157: @Mikehrn @andrewwallacespeckle 

Image

KeepAlive is used incorrectly here. The whole point of KeepAlive is that it will cache the component as its unmounted from the DOM and reactivate it to the same state upon remount. But the components there will never get unmounted, because they don't have a v-if but a v-show. So KeepAlive is completely redundant.

Instead of removing it, however, you should fix it so that 1) neither KeepAlive nor its child has a v-show on it 2) the KeepAlive child has a v-if instead. The only problem you'll likely notice is that the scroll position wont be restored, but tis can be resolved w/ the new composable I created: `useKeepAliveScrollState()`

I already tried doing this for models, but it has some kind of complex virtual list implementation there that I couldn't quickly figure out. Even if we can't figure it out, we should at least remove the KeepAlive if its not doing anything

packages/frontend-2/components/viewer/controls/Left.vue
```diff
@@ -154,10 +154,12 @@
...
+        <ViewerSavedViewsPanel
```
fabis94, line 158: @andrewwallacespeckle 

Image

this should be throttled. scroll events like resize events are high throughput and very spammy, we can't be reacting to every event. you can use `useThrottleFn` for this

andrewwallacespeckle: Done in this branch :)

packages/frontend-2/components/viewer/controls/Left.vue
andrewwallacespeckle: I removed the KeepAlive for now, it's not doing anything as you say, and it's already on v-show, so there's no harm in this right now.

andrewwallacespeckle (approved):

PR #5380, opened 2025-09-04
fix(fe2): selection being made even when selection is disabled

fabis94:


packages/frontend-2/lib/viewer/composables/setup/selection.ts
```diff
@@ -45,6 +46,11 @@ function useSelectOrZoomOnSelection() {
         if (args.hits.length === 0) return trackAndClearSelection()
         if (!args.multiple) clearSelection() // note we're not tracking selectino clearing here
 
+        // Skip if selection disabled
+        if (!state.viewer.instance.getExtension(SelectionExtension).enabled) {
```
fabis94, line 50: @AlexandruPopovici is this a viewer bug? selection events still keep firing even tho the extension is disabled

PR #5401, opened 2025-09-09
fix(server/fileuploads): garbage collector should collect queued jobs which have reached max attempts

iainsproat:
## Description & motivation

Queued jobs which had reached max attempts were neither being picked up by a worker nor garbage collected. In this case they should be garbage collected, as they have already had sufficient attempts.

## Checklist:

### DIFF
--- a/packages/server/modules/backgroundjobs/repositories/backgroundjobs.ts
+++ b/packages/server/modules/backgroundjobs/repositories/backgroundjobs.ts
@@ -76,25 +76,26 @@ export const failBackgroundJobsWhichExceedMaximumAttemptsOrNoRemainingComputeBud
               db.raw('"maxAttempt"') // camelCase requires the column name to be wrapped in double quotes
             )
           })
-          this.where(function () {
-            this.where(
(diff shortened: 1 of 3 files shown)


packages/server/modules/backgroundjobs/repositories/backgroundjobs.ts
```diff
@@ -65,20 +65,37 @@ export const failBackgroundJobsWhichExceedMaximumAttemptsOrNoRemainingComputeBud
...
-            0
-          )
+          this.where(function () {
+            this.where(
+              BackgroundJobs.withoutTablePrefix.col.status,
+              BackgroundJobStatus.Processing
+            ).andWhere(
+              BackgroundJobs.withoutTablePrefix.col.attempt,
+              '>', // greater than because processing jobs may currently equal maxAttempt and still be running
+              db.raw('"maxAttempt"') // camelCase requires the column name to be wrapped in double quotes
```
gjedlicska, line 76: i think if you use the with table prefix way like `BackgroundJbs.col.attempt` it  should work

gjedlicska (changes requested):

PR #5424, opened 2025-09-11
feat(fe): add require-valid-model to viewer

andrewwallacespeckle:
It was reported that pages never 404d on the viewer pages. I checked and we were missing the `require-valid-model` middleware.


fabis94 (changes requested):
Image

PR #5431, opened 2025-09-12
Fix: Small saved views fixes

benjaminvo:
1. Make sure the saved view thumbnail doesn't overflow in the model card component.
2. Lowercase the name in a tooltip
3. Change home view icon


Mikehrn (approved):
🤩

PR #5438, opened 2025-09-15
Enable skp imports experimentally

JR-Morgan:


JR-Morgan:
currently live on testing6

PR #5450, opened 2025-09-15
feat(fe): enable filtering for block instances

andrewwallacespeckle:
Fixing bug reported on forum:
https://speckle.community/t/no-properties-available-for-object-filtering-on-instance-object/20252

Enable filtering for block instances by allowing `InstanceProxy` objects to be processed and adding support for `definitionId` and `definitionName` properties.

### DIFF
--- a/packages/frontend-2/lib/viewer/composables/filtering/dataStore.ts
+++ b/packages/frontend-2/lib/viewer/composables/filtering/dataStore.ts
@@ -74,12 +74,33 @@ export function useCreateViewerFilteringDataStore() {
       // Map from objectId to its property values for efficient filtering
       const objectProperties: Record<string, Record<string, unknown>> = {}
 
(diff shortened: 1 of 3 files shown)


andrewwallacespeckle:
@fabis94 Not sure if you know what block instances are (i didn't), but it's pretty much a SketchUp/Rhino thing where you can place the same component multiple times (like chairs, windows, etc.) and they all reference a single definition. The user wanted to filter by `definitionId` to find all instances of a specific block type, which makes sense and seems useful.

The issue was that our filtering system was excluding `InstanceProxy` objects (the actual instances) because they had "Proxy" in their name, even though they're atomic objects that users want to filter.

I fixed the guard condition to only exclude `InstanceDefinitionProxy` (the templates) while allowing `InstanceProxy` (the instances) to be processed.

fabis94 (approved):
no idea

PR #5474, opened 2025-09-17
fix(fe): restore copy version link behaviour

andrewwallacespeckle:
Fix bug where the version id is not added to the link in "copy version link"


andrewwallacespeckle:
@fabis94 I see you updated the `copyModelLink` function in this PR:
https://github.com/specklesystems/speckle-server/pull/5250

It seems that it no longer adds the version id to the url when passed. I'm lacking a bit of context on what your change was can you check out this fix and let me know if it conflicts with what you were trying to achieve? Or if there is a better fix?

Thanks

fabis94 (approved):

PR #5530, opened 2025-09-23
feat(fe): filters panel popular filters buttons

andrewwallacespeckle:
image

I've updated the FILTERS_POPULAR_PROPERTIES list based on mixpanel. We decide which properties to show based on the ordering in this list. 

In the property selection panel, we currently have no limit on popular properties. 

In EmptyState, we limit it to 6 popular properties, and it will never wrap into a third line.


benjaminvo:
@andrewwallacespeckle @michalspeckle Another approach could be launch with property search panel open when you enter the Filters panel. Could be handy – or a little confusing. I don't know. But it already has the "popular filters" listed. 

Screenshot 2025-10-08 at 15 26 51

PR #5561, opened 2025-09-25
feat(ifc importer): respect timeouts

gjedlicska:
makes the ifc parsing respect timeouts by using a subprocess, the same way as were doing for rhino


packages/ifc-import-service/src/ifc_importer/client.py
```diff
@@ -0,0 +1,26 @@
+from specklepy.core.api.client import (  # pyright: ignore[reportMissingTypeStubs]
+    SpeckleClient,
+)
+
+from src.ifc_importer.domain import FileimportPayload
```
JR-Morgan, line 5: This import is giving me some trouble running the ifc-import-service docker image locally.

```
Traceback (most recent call last):

  File "/app/main.py", line 10, in <module>

    from ifc_importer.job_manager import job_manager

  File "/app/lib/python3.13/site-packages/ifc_importer/job_manager.py", line 16, in <module>

    from ifc_importer.client import setup_client

  File "/app/lib/python3.13/site-packages/ifc_importer/client.py", line 5, in <module>

    from src.ifc_importer.domain import FileimportPayload

ModuleNotFoundError: No module named 'src'
```

JR-Morgan (changes requested):

JR-Morgan (changes requested):
I suspect the `TimeoutError` exception may not have a message, thus it renders in the FE as 
image

In C# we use the message $"Job was cancelled due to reaching the {jobTimeout} second timeout"

PR #5634, opened 2025-10-03
fix(ifc-import-service): bump speckle py

JR-Morgan:
Bump specklepy to deploy https://github.com/specklesystems/specklepy/pull/458


JR-Morgan:
tested locally

PR #5637, opened 2025-10-03
feat: skip builds on pr

danielgak:
Skiping builds on pr detected files.

Moved change detection to root pr workflow and pass every skip as an argument.
Added the flowing new skips:
- webhook-service
- fileimport-service
- ifc-import-service
- monitor-deployment

image

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -71,6 +71,8 @@ jobs:
             monitor-deployment:
               - 'packages/monitor-deployment/**'
               - 'packages/shared/**'
+            utils:
+              - 'utils/**'
 
   tests:
     needs: [get-version, changes]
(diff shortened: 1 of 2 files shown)


danielgak:
I left the following jobs to always run:
- the precommit check
- lint & prettier 
- test deploy util 
- docker compose ingress

@iainsproat any idea when we can skip some of those in an simplish way? 🤔

danielgak:
Lint and prettier still takes significantly more in this cases (~5 mins)

.github/workflows/pull-request.yml
```diff
@@ -11,14 +11,83 @@ jobs:
...
+              - 'packages/shared/**'
+            fileimport-service:
+              - 'packages/fileimport-service/**'
+              - 'packages/shared/**'
+            ifc-import-service:
+              - 'packages/ifc-import-service/**'
+              - 'packages/shared/**'
+            monitor-deployment:
+              - 'packages/monitor-deployment/**'
+              - 'packages/shared/**'
```
danielgak, line 73: @iainsproat @fabis94 any dependency that I might missed here?

PR #5639, opened 2025-10-03
fix(fe): Ensure application id values are classified as string filters

andrewwallacespeckle:
Fixes issue where `applicationId` properties containing both numeric values (`"468772"`) and UUIDs (`"8c8a46db-b16f-40ba-a2a7-a16c4db4ef3d-0062b95f"`) were incorrectly classified as numeric filters due to batched processing.

- Added `isValueNumeric` utility that excludes values containing letters/dashes from numeric classification
- Implemented type conflict resolution to update Numeric → String when non-numeric evidence is found in later batches
- Ensures consistent filter type determination across datastore population and UI display
- Also made it so boolean is case insensitive


Mikehrn (approved):

fabis94 (changes requested):
t

fabis94 (approved):
t

PR #5647, opened 2025-10-03
fix(server/updateServerInfo): sanitize user input

iainsproat:
## Description & motivation

We should sanitize user input, especially user input that is displayed to other users.
- server info is only editable by server admins, so this has limited scope.

## Checklist:


packages/server/modules/core/utils/input.ts
```diff
@@ -0,0 +1,9 @@
+import sanitizeHtml from 'sanitize-html'
+
+export const sanitizeUserInput = <T extends Record<string, unknown>>(input: T) =>
+  Object.fromEntries(
+    Object.entries(input).map(([key, value]) => [
+      key,
+      typeof value === 'string' ? sanitizeHtml(value) : value
```
fabis94, line 7: are we really OK w/ default settings here? like, wont this strip out anchors? we still want urls to work. would like to see this in tests too

additionally, running this on an entire object seems overkill cause ultimately you need at most 1 field to be sanitized.

PR #5668, opened 2025-10-06
fix(fe2): flashing loader on federated presentation load

fabis94:
## Checklist:

### DIFF
--- a/packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
+++ b/packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
@@ -114,7 +114,7 @@ function useViewerObjectAutoLoading() {
       },
       response: { resourceItems, savedView }
     },
-    ui: { loadProgress, loading, spotlightUserSessionId },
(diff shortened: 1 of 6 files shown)


packages/frontend-2/lib/viewer/composables/setup/postSetup.ts
```diff
@@ -198,6 +198,7 @@ function useViewerObjectAutoLoading() {
 
         if (res.length) {
           hasDoneInitialLoad.value = true
+          hasLoadedQueuedUpModels.value = true
```
fabis94, line 201: this should work, but isnt 100% foolproof because watcher runs don't `await` the async logic inside and so a subsequent watcher run (e.g. triggered by user quickly clicking on different federated views?) could overlap and change the bool value to the wrong state. it could be an edge case tho!

for a better and more thorough fix, we should involve Alex and see if we can do more on the Viewer side to 1) either load all possible models early on and hide them and/or 2) report events back for when a batch of objects have been loaded

Mikehrn (approved):
🙌

PR #5669, opened 2025-10-06
Fix: Decrease permission required for limitedWorkspace on projects

Mikehrn:
Not sure if this is the right fix, please point me in the right direction 😄  Basically the limitedWorkspace should always be returned if the project is public


gjedlicska (approved):

gjedlicska:
Tested and the change makes sense.

PR #5715, opened 2025-10-08
feat(viewer): expose object loader attribute masking to the viewer

gjedlicska:
just some legwork to expose the object loader attribute masking argument


AlexandruPopovici (approved):
LGTM

PR #5717, opened 2025-10-08
fix(gatekeeper): allow checking all feature access on a per project

gjedlicska:
fixes the discrepancy between workspace and project level resolver


Mikehrn (approved):
Works!

PR #5728, opened 2025-10-31
fix null object check in queryHits

bachtiarnaomi:
### Problem
In Speckle Renderer, queryHits sometimes pushes null objects to rvs when rvMaterial[0] is null due to operator precedence in the ternary expression.

### Solution
Added explicit parentheses around the ternary to ensure rvMaterial[0] is checked before evaluating the filter. This prevents null objects from being added.


AlexandruPopovici (approved):
Good catch Naomi!

PR #5751, opened 2026-03-18
feat(auth): add connectors v3 as an app

JR-Morgan:
see https://github.com/specklesystems/speckle-server-internal/pull/1141/changes

### DIFF
--- a/packages/server/modules/auth/defaultApps.ts
+++ b/packages/server/modules/auth/defaultApps.ts
@@ -87,7 +87,7 @@ const SpeckleConnectorsDUI = {
   ]
 }
 
-const ConnectorsV3 = {
+const SpeckleConnectorsV3 = {
   id: DefaultAppIds.ConnectorsV3,
   secret: DefaultAppIds.ConnectorsV3,
   name: 'Desktop Connectors',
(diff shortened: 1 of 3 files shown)


iainsproat (dismissed):

JR-Morgan:
tested locally via docker...
works as expected

.github/workflows/get-version.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       - run: git fetch origin 'refs/tags/*:refs/tags/*'
       - run: chmod +x ./get_version.sh ./common.sh
         working-directory: ./.github/workflows/scripts
+      # sleep for 2 mins for debugging purposes, to check if the
+      - run: sleep 120
```
iainsproat: Do we need this change?

iainsproat (changes requested):

.github/workflows/get-version.yml
JR-Morgan: Currently debgging why the CI is failing. 
Danny and I are are banging our head against the wall

danielgak (approved):

